PR #286, opened 2014-10-10
Improved thumbnail scrolling for large numbers of thumbnails

jaronkk:
1) Fixed bug in throttle function in utils.js
2) Throttle the thumbnail scroll event so it doesn't run more than once every 100 ms
3) Updated the isOnScreen function so it can use an existing viewport
4) Use a single viewport in the loadImages function
5) Break out of the loadImages each loop when the loop reaches the right side of the viewport
6) Removed fade in animation for the thumbnail view, since it caused unnecessary flickering, and delayed the thumbnail loading by a full second.
7) Call currentImageChanged() when the thumbnail view is loaded so it will show the currently active thumbnail.


aeschylus:
Closing this since it has been broken into smaller issues that need to be removed now that the first release is out. Thumbnails are going to change significantly with the Ian Gilman integration work.

PR #433, opened 2015-02-24
Fixes stacks IIIF URL and version errors

jchristo4:


aeschylus:
This was handled by Rashmi I believe in a slightly different way. We will return to this after release, but I'm closing for now.

Issue #451, opened 2015-03-12
User-configurable buttons for mainMenu.


pobocks:
It would be useful to have a user-configurable way to add functionality to the main menu bar.

I have a pull request that implements this.  It supports adding static links, arbitrary callback functions, and sublists containing either of the above.

Right now, sublists don't have any CSS/JS behavior provided, so they effectively don't work without user-provided CSS/JS.  Callbacks are implemented via attaching a handler to document, namespaced by the button's ID.  Exceptions are squashed (on a "per button" level), and turned into empty jQuery objects, which translate to '' when output via template helper.

PR will be forthcoming.

PR #452, opened 2015-03-12
First try @ configurable menu extension - refs #451

pobocks:
Herein is implemented an attempt at allowing users to add
menu items to the main menu in Mirador via a configuration
setting.  Plaintext links to href or that trigger callbacks
are more or less fully handled.  Sublists and explicit handling
of links with icons are the next step here.

index.html contains a basic example of each type of link, and
there is inline documentation co-located with the functions that
implement it.

Refs #451

### DIFF
--- a/css/mirador.css
+++ b/css/mirador.css
@@ -63,6 +63,10 @@ a {
   text-align: right;
 }
 
+.mirador-main-menu.user-buttons {
+  padding-right: 0;
+}
+
 .mirador-main-menu li {
   cursor: pointer;
   float: left;
--- a/css/mirador.css
+++ b/css/mirador.css
@@ -66,6 +66,11 @@ a {
 .mirador-main-menu.user-buttons {
   padding-right: 0;
 }
+.mirador-main-menu.user-logo {
+  padding-right: 0;
+  padding-left: 0;
+  float: left;
+}
 
 .mirador-main-menu li {
   cursor: pointer;
--- a/index.html
+++ b/index.html
@@ -19,6 +19,28 @@
       Mirador({
         "id": "viewer",
         "layout": "1x1",
+        "mainMenuSettings" :
+          {"show": true,
+           "buttons" : {"bookmark" : true, "layout" : true},
+           "userButtons": [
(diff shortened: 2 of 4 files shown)


aeschylus:
This is great behaviour to have, and I will probably look at this for making the canvas-level controls more extensible. I think the merge will have to wait until after release, so that our tests are meaningful (currently there is a problem with the build).

pobocks:
Ok.  Let me know if there's anything you can think of that you want here, or if you want me to generalize this out further for use with other lists of buttons.

Side note: Is there a decided-upon standard method for doing drop-downs?  I'd like to provide basic CSS/JS support for sublists, but in the absence of a broader "whole application" decision, I'm uncomfortable doing so ad-hoc.

pobocks:
Also, this is essentially broken right now because replaceState can only take a serializable object, which means no function properties.  Gonna have a thunk on this.

pobocks:
So, as long as Mirador is putting its entire configuration into localStorage and replaceState, callbacks can't go in config.  I got about halfway through setting up a registry to mask them out before storing and back in after, but I'm unsure as to what the right behavior is here on my part. 

As far as I can see, to be interpolated into the mainMenu, things more or less have to live in the config passed to Mirador.  Callbacks can live outside of Mirador and just hang off of document, no problem - but am I correct in thinking that, without special handling, anything that's in config will get overwritten by localStorage version if one exists?

pobocks:
I've reworked this so that "callback" is just a boolean that enables error checking (essentially, it triggers a check to make sure there's an id attribute to hang callbacks off of.)  An example callback is included at the bottom of index.html.  I'll keep this rebased on top of origin/master, you can merge when you feel like it.

Alternately, if you want me to make changes first, just let me know.

pobocks:
@aeschylus This should be ready to go if you think it's acceptable - if not, let me know.

aeschylus:
Finally got a chance to review this last night. Thanks for implementing this @pobocks (and for the tests)!

pobocks:
You're very welcome! Thanks for merging this!

Issue #481, opened 2015-03-20
"dead click" spots in ManifestListItems.


pobocks:
Spaces between thumbnails, end of ManifestListItem, etc, are not clickable, but show clickable cursor and are logically part of the item that could be used for selection. 

I think making the entire item clickable, and stopping propagation in any handlers for subElements that need special handling (right now just thumbnails) seems like reasonable behavior.

aeschylus:
Agreed.

aeschylus:
Nice. This is what I would have done for the problem. I'll merge this as soon as I test it out.

PR #482, opened 2015-03-20
Fix for issue #481 - dead click spots in ManifestListItems

pobocks:
Proposed fix: basically just attaches the event handler one level higher, and does stopPropagation in the handlers for the thumbnails.


pobocks:
Refs #481

Issue #517, opened 2015-04-13
Image URLs are returned as .jpg regardless of mime type in manifest 


pobocks:
According to http://iiif.io/api/image/2.0/#image-request-uri-syntax, format specifiers other than jpg should be handled.  However, in the URL generation code ([hereabouts](https://github.com/IIIF/mirador/blob/master/js/src/utils/iiif.js#L28)), the format specifier is being set to .jpg regardless of anything in the manifest.

My theory on the right thing to do: use a hash of mime types that work with OpenSeadragon I'm actively working on this - should have a pull request soon.

azaroth42:
:-1:
I disagree that this is an issue.  IIIF servers MUST support jpg as a format at all levels of compliance:
See http://iiif.io/api/image/2.0/compliance.html#format

And thus if there is a IIIF service associated with an image resource, then using the jpg format should always work.

Recommend: wontfix

aeschylus:
I agree with Rob that this would add complexity without much benefit. Jpeg can be dependably expected from every server.

pobocks:
We might be unique, in having significant image resources that don't have
jpg deliverables, but it seems like the amount of work is small enough
(it's a few lines in iiif.js and two lines in utils.js) that it's worth
doing for the benefit of people with large sets of legacy image assets (or,
as in our case, services that do cheaper tif->gif conversion) or preferred
formats that aren't jpg.

After all, even if .jpg is guaranteed by the standard, is silently failing
to use alternate formats where available and requested really the right
thing to do?  And at some point it has to start caring for level 3
compliance, unless I'm misunderstanding something?

PR added to illustrate, not in expectation of being merged.

azaroth42:
Still :-1: to guessing a format, based on a potentially unrelated URI's
format (static thumbnail in a cache vs a IIIF service driven from a
different master), rather than selecting from the list of formats available
from the service.

In this situation, you would need to either:
 a. Eat the cost of requesting info.json for every thumbnail to find out
which formats are supported
 b. Include that information in the manifest for every canvas

Basically, if the implementation doesn't support "jpg" in the format slot,
it's not conformant with the specification.  Every level of conformance
requires a jpg formatted version to be available. OpenSeadragon also simply
requests JPGs on the same basis, as per:

https://github.com/openseadragon/openseadragon/blob/master/src/iiiftilesource.js#L240

So you're not gaining anything anyway as you won't be able to see the image
in the main view either, regardless of getting a thumbnail.

[...]

PR #518, opened 2015-04-13
Refs #517 Respect mime types for thumbs.

pobocks:
Strawman causing thumbnail uri creation to respect mime types with default to jpeg.

### DIFF
--- a/js/src/utils/iiif.js
+++ b/js/src/utils/iiif.js
@@ -25,12 +25,23 @@
       }
     },
 
-    makeUriWithWidth: function(uri, width, version) {
+    makeUriWithWidth: function(uri, width, version, mimetype) {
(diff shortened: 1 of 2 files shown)


aeschylus:
This is a reasonable bit of logic for a wrapper for the IIIF Image API in general, in that the API ought to return the format specified, assuming it supports it (which ought to have been checked in the info.json).

However, in mirador I don't see this being used. I think the community of people who will deploy Mirador and want all the thumbnails to be PNGs for example will be very small, and I can't see a reason for wanting this unless their image server is not compatible with the IIIF APIs.

aeschylus:
Closing for now, but the commit will still be useful for further discussion and is available in the closed PRs section.

Issue #287, opened 2014-10-11
Mouseover in ManifestsPanel should show full label


azaroth42:
Otherwise when the first part is the same, the title isn't useful and there's no way to get to it.

![screen shot 2014-10-10 at 6 27 08 pm](https://cloud.githubusercontent.com/assets/871868/4600973/d336799c-50e5-11e4-9422-33a603f6e096.png)

regisrob:
+1. Would be nice for first m2 release (at least by providing a tooltip when there is not enough space to display the manifest label AND the preview)
Moreover, manifest labels should use as much space as possible (see screenshot below):
![capture d ecran 2015-03-09 a 17 53 18](https://cloud.githubusercontent.com/assets/402117/6559811/c0a905b8-c685-11e4-9418-0454e11be6a3.png)

regisrob:
Another example here: http://demos.biblissima-condorcet.fr/florus/manuscrits-florus/mirador/index.html
IMO having the full label (at least if it's not too long) is quite critical.

aeschylus:
Yep. I have a re-design of this that will take care of this issue. I'll post some images here today.

regisrob:
Cool :wink: I'll have a look tomorrow

aeschylus:
![load menu](https://cloud.githubusercontent.com/assets/145874/7397234/eb48c440-ee5b-11e4-9887-d55a12e2af14.png)

### Hover States
![mockup](https://cloud.githubusercontent.com/assets/145874/7397277/23a2d6aa-ee5c-11e4-9eac-b205db5bfc61.png)

regisrob:
Good re-design, seems appropriate to me

jvine:
I had been playing with this layout in parallel - here is what I came up with, and at least some of the rationale.

![mirador-load-window-layout](https://cloud.githubusercontent.com/assets/6945691/7761319/3b2a2e20-ffd9-11e4-9a32-924e2db3c672.png)

- I think the use of the left sidebar is a poor use of space, and closely groups two input controls that actually behave differently. So:
 - Move the controls above the list
 - Align the filter input with the content that it filters
 - Remove the redundant placeholder text from the filter input
 - Separate the load feature to the right
 - Easy to disable/hide the load feature without unbalancing the page, if that's appropriate in certain contexts
- IMO the background transparency introduces visual noise and makes it harder to read and focus on the elements of this page. Try using box-shadow to communicate layers - maybe bring the window in a bit further from the top and side borders. 
[...]

PR #522, opened 2015-04-15
Manifest thumbnail is read from manifest if available.

hashimmm:
This could be improved so that if the thumbnail URI is an IIIF image service, then a thumbnail of the required size should be requested; but it seems sensible this way too.


azaroth42:
:+1: to this PR from me! Thanks Hashimmm

hashimmm:
Thanks. I only meant to include one commit. :$ The second one adds description tooltips to the manifest titles in the manifest list panel.

aeschylus:
I would like to merge this, but not with the tooltips. However, I understand why you would want them. In fact, there's a whole issue about it at #287.

hashimmm:
Sure, I made a separate branch and another pull request for this ( #533 ).
Thanks.

Issue #414, opened 2015-02-12
metadataView: linking fields do not support object or list of objects


regisrob:
Mirador seems to only supports strings in related/seeAlso/within fields.

For example:
<pre>
"related":{
    "@id": "http://example.org"
}
</pre>
will print "@id: http://example.org" in metadata view

and:
<pre>
related: [
  {
      "@id": "http://example.org"
  },
  {
      "@id": "http://example2.org"
  }
]
</pre>
will print:
[
@id: http://example.org
,
@id: http://example2.org
]

Presentation API says "A label and the format of the related resource should be given to assist clients in rendering the resource to the user": http://iiif.io/api/presentation/2.0/#linking-properties
Question (cc @azaroth42): How it is supposed to be written in JSON-LD?

The complete example in the spec gives:
<pre>
"related":{
    "@id": "http://www.example.org/videos/video-book1.mpg",
    "format": "video/mpeg"
}
</pre>

But if I want a label as recommended in the spec, is it allowed to write?:
<pre>
"related":{
    "@id": "http://www.example.org/videos/video-book1.mpg",
    "format": "video/mpeg",
    "label": "A video about Book 1"
}
</pre>

If so, I guess the expected rendering in Mirador would be:
Related:    [A video about Book 1](http://www.example.org/videos/video-book1.mpg)
(with an optional tooltip or icon, depending on the value of "format")

... or if there are two links: 
Related:&nbsp;&nbsp;[A video about Book 1](http://www.example.org/videos/video-book1.mpg)<br />&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[Catalogue record](http://www.example.org/record/12345)

Is it the expected display? And is it correct with respect to the spec?

azaroth42:
That is the correct JSON-LD :)

regisrob:
Ok, so if it's correct I will use that syntax in our Gallica manifests (for the links to the catalogue records). 
The stringifyObject function will need some refinements ([metadataView.js#53](https://github.com/IIIF/m2/blob/master/js/src/widgets/metadataView.js#L53)). I'll try to work on this ASAP.
Thx

mikeapp:
Ignore the reference above, it would impact other values with an @id and profile, such as a logo image specified via IIIF.

rsinghal:
[...]

PR #523, opened 2015-04-15
Apply table model to metadata view: changed template markup + css

### LINKED ISSUE #519: Display issues in metadata panel
Floats are not properly cleared in "metadata-listing" definition list. It appears when the last label breaks in two lines, and it raises the following bugs:

* The "sub-title" of the following section is not aligned properly (see "Rights"): 
selection_005

* Or the margin between the next section is not properly set:
selection_004

This widespread clearing method does the trick:
<pre>
.mirador-viewer dl.metadata-listing:after {
content: "";
display: table;
clear: both;
}
</pre>

There is another bug that is quite embarrasing because it affects the proper reading of the metadata (label/value pairs are not aligned with each other when there are multiple labels breaking in two lines):
selection_003

**Suggestion: would it be better to have divs instead of definition lists and apply a tabular layout with css?**
The markup could be:
```
<div class="metadata-listing">
  <div class="row">
    <div class="label cell">Label:</div><div class="value cell">Value:</div>
  </div>
</div>
```

regisrob:
This PR fixes #519. 
Tested on Mac OS 10.8.5 in the latest versions of FF (37.0.1), Safari (6.2.4) and Chrome (41.0.2272.118)

capture d ecran 2015-04-15 a 10 23 54

Extra comments:
- I choose to apply "vertical-align:bottom" to metadata value when the label breaks in two lines (see screenshot)
- still a responsiveness issue when there are 3+ slots in the viewport
- #414 still pending for a fully proper display of metadata view

The manifest causing the last bug listed in #519 is this one: http://bvmm.irht.cnrs.fr/iiif/2309/manifest (the server is not cors-enabled yet so you need to copy the manifest and test it locally ; and it's still a work-in-progress so the uri and contents might change...)

Feel free to amend the markup or css :wink:


azaroth42:
:+1:

aeschylus:
Nice. Looks good locally and doesn't require any new tests, as we currently do not test CSS. Merging now.

PR #533, opened 2015-05-06
Manifest thumbnail is read from manifest if available.

hashimmm:
Closing the old pull request with the same title because it contained irrelevant commits.

This could be improved so that if the thumbnail URI is an IIIF image service, then a thumbnail of the required size should be requested; but it seems sensible this way too.

Also, as for the test, it could also have been done by comparing the manifest list item's `manifest.jsonLd` with its `tplData.repoImage` property, but I figured comparing the img tag's attribute was more useful.


aeschylus:
Looking at this more closely, this is not actually the correct behaviour. The image/icon that is being brought into the manifest listing is meant to be the logo of the hosting institution. This is important for institutions making their resources available. See the specification for the distinction between the thumbnail and logo. 

I would gladly accept this if either the logo is replaced for thumbnail in this change, or an option is provided that lets the Mirador deployer choose which to display as the main visual cue for the list item (with logo being the default). 

Thanks for the pull request. I look forward to merging it after a few changes.

hashimmm:
Oh. Thanks for letting me know. Alright, sure, I'll make the relevant changes.

hashimmm:
Sorry I left this hanging for a while. I've made the recommended edits (now reading the manifest's logo property instead of thumbnail).
My own use case required me to use "thumbnail". I may do as you suggested and turn it into a setting; won't bother anyone else with it.

PR #538, opened 2015-05-09
Add logo image for University College Dublin

jbhoward-dublin:
Add logo image for University College Dublin


rsinghal:
@jbhoward-dublin We are moving away from this architecture and are encouraging users to use the [logo field] (http://iiif.io/api/presentation/2.0/#rights-and-licensing-properties) on the manifest level which the code will automatically look for.

PR #540, opened 2015-05-14
Annotations z-index not set - set to 2: above OSD, below panels.

Wedjaa:
On Chrome and Firefox it seems that the annotations are not visible, setting their z-index to 2 will make them visible over OSD and invisible below panels, like the info one. I could be the only one with this issue, or maybe not!

### DIFF
--- a/css/mirador-combined.css
+++ b/css/mirador-combined.css
@@ -2079,6 +2079,7 @@ text {
   border: 2px solid white;
   box-shadow: 0px 0px 5px rgba(0,0,0,0.5);
   box-sizing: border-box;
+  z-index: 2
 }
 .annotation.selected {
   border: 3px solid orangered;
--- a/css/mirador.css
+++ b/css/mirador.css
@@ -1134,6 +1134,7 @@ text {
   border: 2px solid deepSkyBlue;
   box-shadow: 0px 0px 5px white; 
   box-sizing: border-box;
+  z-index: 2;
 }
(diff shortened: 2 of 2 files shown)


rsinghal:
I'm not sure I understand the problem.  Annotations should only be visible when the annotations layer is turned on, via the annotation button.  By default, annotations should not be visible.  Is this not working for you?

Wedjaa:
Hi! With the current CSS the annotations - when the annotation layer is enabled - are not visible; by setting the z-index - when the annotation layer is enabled - the annotations are visible. With this PR I'm just making sure that the z-index is explicitly set.

aeschylus:
Strange, what browser and system are you using?

Wedjaa:
Both Chrome and Firefox dev channels - as I mentioned in the comments to the QTip display PR: it may be due to the  way inheritance changed; this PR just makes sure that z-index is set explicitly, so that the browser have no say in the layering of the annotations.

aeschylus:
Okay. I can't test this because it doesn't seem that your configuration includes any annotation endpoint, but I am willing to merge if I see it behaves correctly after the change. Makes sense that they should be between the canvas and other control panels. 

Can I close this request and can we carry this discussion into the other one that subsumes this one?

Wedjaa:
Absolutely yes - I was trying to keep them separate, but my git gremlin thought better.

PR #541, opened 2015-05-14
Annotations Endpoint is not correctly created for each window

Wedjaa:
When an endpoint module is specified in the configuration the way it is initiated when there are multiple windows open is odd. In the current implementation the is a bit of confusion in the way _this.endpoint and endpoint are checked and populated.

For the window instance _this.endpoint is always null when first called, on the other end: endpoint is present if another window, with annotations, has been already initialised.

What happens is that the second window will create a new instance of the endpoint and assign it to the endpoint variable, things get ugly in the deferred, because when the deferred is resolved in that scope, the endpoint from which is getting annotations - via the annotationsList - is no longer the one that is resolving the deferred, but the last one created, possibly with a radically different opinion of what the annotationsList should look like, resulting in a _this.annotationsList that could be different from what is expected.

If it doesn't make sense immediately, just try and step in the code and see for different windows what the _this.endpoint and endpoint variables look like.

This first part of the fix is still using a single global annotations endpoint, I think it's very efficient. What we do is check for endpoint not being assigned - with a very counterintuitive try/catch - and create only one.  The second call to this function, from another window, will get the same endpoint and use the search function to update the annotationsList of the endpoint.

**TODO**: as it is the `search` function will overwrite the existing one, I'm not particularly happy with that because it may have some unintended consequences; I have a fix - will push it later - where I change the API contract of the search, so that it accepts  `resolve, reject` parameters and returns the list of annotations in a deferred fashion. Since it breaks the `search` contract I think we must see first who's creating annotations endpoints and what it would mean for them.

### DIFF
--- a/js/src/workspaces/window.js
+++ b/js/src/workspaces/window.js
@@ -595,28 +595,28 @@
         module = $.viewer.annotationEndpoint.module,
         options = $.viewer.annotationEndpoint.options;
 
-	// At this point 'endpoint' can be either null
-	// or already created. If we try and check for
+        // At this point 'endpoint' can be either null
+        // or already created. If we try and check for
         // undefined - as in 'if (endpoint)' we will
-	// crash and burn, since it's a null - not an
-	// undefined.
-	try {
-	  // If we can assign endpoint is defined,
-	  // use the search function of the endpoint
-	  // to search.
-	  endpoint = endpoint;
-	  endpoint.set('dfd', dfd);
-	  endpoint.search(_this.currentCanvasID);
-	} catch(error) {
-	  // We failed, endpoint is not assigned
-	  // yet - so we need to get the endpoint
-	  // module instance.
+        // crash and burn, since it's a null - not an
+        // undefined.
+        try {
+          // If we can assign endpoint is defined,
+          // use the search function of the endpoint
+          // to search.
+          endpoint = endpoint;
+          endpoint.set('dfd', dfd);
+          endpoint.search(_this.currentCanvasID);
+        } catch(error) {
+          // We failed, endpoint is not assigned
+          // yet - so we need to get the endpoint
+          // module instance.
           options.element = _this.element;
           options.uri = _this.currentCanvasID;
           options.dfd = dfd;
           options.windowID = _this.id;
           endpoint = new $[module](options);
-	}
+        }
 
         dfd.done(function(loaded) {
           _this.annotationsList = _this.annotationsList.concat(endpoint.annotationsList);
--- a/js/src/workspaces/window.js
+++ b/js/src/workspaces/window.js
@@ -595,31 +595,21 @@
         module = $.viewer.annotationEndpoint.module,
         options = $.viewer.annotationEndpoint.options;
 
-        // At this point 'endpoint' can be either null
-        // or already created. If we try and check for
-        // undefined - as in 'if (endpoint)' we will
-        // crash and burn, since it's a null - not an
-        // undefined.
-        try {
-          // If we can assign endpoint is defined,
-          // use the search function of the endpoint
-          // to search.
-          endpoint = endpoint;
-          endpoint.set('dfd', dfd);
-          endpoint.search(_this.currentCanvasID);
-        } catch(error) {
-          // We failed, endpoint is not assigned
-          // yet - so we need to get the endpoint
-          // module instance.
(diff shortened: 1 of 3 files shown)


azaroth42:
:+1: Great explanation of the problem and solution.

js/src/workspaces/window.js
```diff
@@ -594,17 +594,30 @@
         var dfd = jQuery.Deferred(),
         module = $.viewer.annotationEndpoint.module,
         options = $.viewer.annotationEndpoint.options;
```
aeschylus: Adding an `endpoint = _this.endpoint;` would solve the problem by eliminating the unintentional global variable. However, one must make sure to delete the `;` from line 595 and replace it with a comma.

aeschylus:
Thanks for catching this bug, but the problem is actually something different: the endpoint variable is not confined to the scope of the getAnnotations function because `var` is not used anywhere. An oversight with terrible implications, since `endpoint` is now a global variable floating around in any webpage mirador lives in. That's why it's being set by other windows. 

Each window is supposed to create its own endpoint, not use a global application level one. Annotations are still an evolving topic (what with talk of dynamic annotation services, multiple lists on a single object, different sources for different pages), so I apologise for the state of this whole subsystem. If we were to have one endpoint instance for the whole viewer, it should be passed into the window as the configuration This patch allows the global variable to continue to survive while allowing windows to use each other's endpoints and remove each other's annotation lists.

I don't agree with checking for the existence of something using a try/catch.

I added a comment that should help. 

If you make this change I'll accept the pull request. Thanks for the catch.

Wedjaa:
Sorry for having kept the endpoint global alive - I didn't know if there was a reason for it or not. With this version of the PR I removed it completely, always referencing `_this.endpoint` for clarity. There is now an instance per window.

Still not clear on the `search` method of the endpoint: would it make sense to have it return the annotations via a deferred or is it a kind of initializer and makes sense to keep it as it is - changing the `annotationsList` of the instance. But this is another conversation! :)

aeschylus:
Yes, the `search method` is another conversation. Someday! The method overwrites the annotation list because the list only pertains to the current manifest page, so it needs to be undone and refetched whenever the use changes pages. It's basically just GET-ing the annotationList for this page from the endpoint. 

One problem with this is that there is no well-insulated local cache of the annotations, so the request flow is a bit tied up with the display/updating code. There is some effort to extract and abstract this [here](https://github.com/IIIF/manifesto).

I tried testing this and your other PR locally, and it seems you don't have an endpoint configured, so it would be very strange if any annotations appeared. What are your configuration options?

Wedjaa:
This is how I setup the Mirador instance:
```
	      Mirador({
		"id": "viewer",
		"layout": "1x2",
		"mainMenuSettings" :
		  {"show": true,
		   "buttons" : {"bookmark" : true, "layout" : true},
		   "userLogo": {"label": "IIIF", "attributes": {"href": "http://iiif.io"}}
		 },
		'showAddFromURLBox' : false,
		"saveSession": true,
		"data": [
                  { "manifestUri": manifestUri, "location": "The Frick Collection"},
                ],
		"windowObjects": windowObjects,
		/** Testing Annotations **/
	        annotationEndpoint: {
				name: 'Frick Collection Annotation',
				module: 'FrickAnnotationsEndpoint',
				options: {
					url: '/api/annotation',
					storeId: 'comparison',
					APIKey: 'user_auth'
				}
		}	
	      });
```
The annotation endpoint module is what we are using at the Frick internally, it's based on an ElasticSearch backend - but that's a long story! I'll put up the endpoint as a gist...

aeschylus:
Also, I should add about the search method that it already does return a promise, but for the whole list, not atomic updates to it.

aeschylus:
Ah, I see, you have another endpoint adaptor. Excellent! I look forward to seeing it. I am working on one for triannon over here:https://github.com/IIIF/triannon-adaptor

Wedjaa:
I had to do a bit of guesswork and try and figure out the annotation endpoint from the way mirador called it - and I saw that the search was really passed only a deferred, so I used that one. Didn't know it was supposed to return a promise - also because Mirador doesn't seem to be using it.

This is the endpoint I created: https://gist.github.com/Wedjaa/1f8a1fd0c0c4a904572d

Wedjaa:
Need to rebase this PR - will open a new one. I messed up my repo.

PR #542, opened 2015-05-14
Keep QTip2 annotations in sight

Wedjaa:
When an annotation is close to the edge of the viewer it will be displayed outside the visible screen, possibly reduced to fit. 

This change takes advantage of the viewport positioning available in QTip. The annotation will shift as needed to be displayed inside the viewport.

### DIFF
--- a/css/mirador-combined.css
+++ b/css/mirador-combined.css
@@ -2079,6 +2079,7 @@ text {
   border: 2px solid white;
   box-shadow: 0px 0px 5px rgba(0,0,0,0.5);
   box-sizing: border-box;
+  z-index: 2
 }
 .annotation.selected {
   border: 3px solid orangered;
--- a/css/mirador.css
+++ b/css/mirador.css
@@ -1134,6 +1134,7 @@ text {
   border: 2px solid deepSkyBlue;
   box-shadow: 0px 0px 5px white; 
   box-sizing: border-box;
+  z-index: 2;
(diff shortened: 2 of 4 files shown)


Wedjaa:
Forgot to remove the z-index for annotations - that's another PR. Sorry.

aeschylus:
I am deleting the other pull request (#540) since this subsumes it. I like the display tweak. There have been advocates of the popup being at the top level (above menus), but I agree it is more logical between the canvas it annotates and the window that controls that canvas.

However, you seem to have said that annotations were invisible for you. Is this true, or did I misunderstand your comment from the other request? Were they not appearing at all?

Wedjaa:
Without a  specific z-index the annotations where not visible - with the github master version; it could be due to the versions of Chrome and Firefox I use - both the bleeding edge kind, these may have a different inheritance method for calculating the z-index, so - if not specified - the actual z-index of the annotations would be below the OSD layer, making them invisible.

That PR - accidentally ended up in here as well - just makes the z-index explicit, reducing the chance that some browser may want to go and try to be clever.

Wedjaa:
Fixed also for when the edit window for annotations opens. With this behaviour the edit window will be shifted into view even if the corner of the annotation rectangle is close or out of the view.

aeschylus:
Tested this locally and it's a nice change. Merging.

PR #545, opened 2015-05-15
Feature - Pin annotation

Wedjaa:
I think there is someone already working on this, but I had a requirement for it, so I had to create it. This is a first stab at it, not completely happy with the behaviour, but to sum it up:

- added an header to the tooltips with a 'pin' option;
- pinning the annotation will keep it open until OSD is updated, zooming or moving or an annotation is edited/removed.

Things I'm not happy about:

- removing one annotation will unpin all the annotations on the canvas; this is due to the way the canvas is updated when the annotations are changed; can't figure out an elegant way to handle this;

Comments in line with the PR code for the inner workings of this feature.

### DIFF
--- a/css/mirador.css
+++ b/css/mirador.css
@@ -1134,6 +1134,7 @@ text {
   border: 2px solid deepSkyBlue;
   box-shadow: 0px 0px 5px white; 
   box-sizing: border-box;
+  z-index: 2;
 }
 .annotation.selected {
   border: 3px solid orangered;
--- a/css/mirador.css
+++ b/css/mirador.css
@@ -1134,7 +1134,6 @@ text {
   border: 2px solid deepSkyBlue;
   box-shadow: 0px 0px 5px white; 
   box-sizing: border-box;
-  z-index: 2;
 }
 .annotation.selected {
   border: 3px solid orangered;
--- a/css/mirador.css
+++ b/css/mirador.css
@@ -1538,6 +1538,11 @@ div.manifest-info a.mirador-btn.mirador-icon-toc {
   max-height: 185px;
   overflow: auto;
 }
+
+.qtip-header {
+   background: #cecede;
+}
+
 .annotation-tooltip {
   padding: 12px; 
 }
--- a/js/src/annotations/annotationTooltip.js
+++ b/js/src/annotations/annotationTooltip.js
@@ -99,7 +99,7 @@
                                        '{{#each annotations}}',
                                        '<div class="annotation-display annotation-tooltip" data-anno-id="{{id}}">',
                                        '<div class="button-container">',
-                                            '<a href="#pin" class="pin"><i class="fa fa fa-circle-o fa-fw"></i>Pin</a>',
+                                         '<a href="#pin" class="pin"><i class="fa fa fa-circle-o fa-fw"></i>Pin</a>',
                                          '{{#if showEdit}}<a href="#edit" class="edit"><i class="fa fa-pencil-square-o fa-fw"></i>Edit</a>{{/if}}',
                                          '{{#if showDelete}}<a href="#delete" class="delete"><i class="fa fa-trash-o fa-fw"></i>Delete</a>{{/if}}',
                                        '</div>',
--- a/js/src/annotations/annotationTooltip.js
+++ b/js/src/annotations/annotationTooltip.js
@@ -96,10 +96,10 @@
 
     viewerTemplate: Handlebars.compile([
                                        '<div class="all-annotations">',
+                                        '<div class="qtip-header"><a href="#pin" class="pin"><i class="fa fa fa-circle-o fa-fw"></i></a></div>',
                                        '{{#each annotations}}',
                                        '<div class="annotation-display annotation-tooltip" data-anno-id="{{id}}">',
                                        '<div class="button-container">',
-                                         '<a href="#pin" class="pin"><i class="fa fa fa-circle-o fa-fw"></i>Pin</a>',
                                          '{{#if showEdit}}<a href="#edit" class="edit"><i class="fa fa-pencil-square-o fa-fw"></i>Edit</a>{{/if}}',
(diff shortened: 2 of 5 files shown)


css/mirador.css
```diff
@@ -1538,6 +1538,11 @@ div.manifest-info a.mirador-btn.mirador-icon-toc {
   max-height: 185px;
   overflow: auto;
 }
+
```
Wedjaa: This sets a background for the header - it looks kind of empty; I wish I could find a non-hacky way of pushing the first edit and delete buttons into the header.

css/mirador.css
```diff
@@ -1583,7 +1588,7 @@ div.manifest-info a.mirador-btn.mirador-icon-toc {
 .annotation-display.annotation-tooltip .button-container {
   text-align: right;
 }
-.annotation-display .edit, .annotation-display .delete  {
+.annotation-display .edit, .annotation-display .delete, .annotation-display .pin  {
```
Wedjaa: Make the pin consistent in style with the existing buttons.

js/src/annotations/annotationTooltip.js
```diff
@@ -96,6 +96,7 @@
 
     viewerTemplate: Handlebars.compile([
                                        '<div class="all-annotations">',
+                                        '<div class="qtip-header"><a href="#pin" class="pin"><i class="fa fa fa-circle-o fa-fw"></i></a></div>',
```
Wedjaa: This adds the header and the pin in it.

js/src/annotations/osd-canvas-renderer.js
```diff
@@ -269,9 +269,19 @@
       });
 
     },
-    
+      hideWindowAnnotations: function() {
```
Wedjaa: When hiding annotations - in case an annotations was removed, the OSD moved or someone started editing an annotation - I want to make sure I only close the annotations on the current window. This is not an issue when you have no 'sticky' annotations, but when you do, you want to make sure you only close the ones belonging to the windows where the event happened.

js/src/annotations/osd-canvas-renderer.js
```diff
@@ -269,9 +269,19 @@
...
+          // Reset the annotations to unpinned and hide
+          jQuery.each(windowAnnotations, function(index, value) {
+              var qtipApi = jQuery(value).qtip();
+              qtipApi.set({'hide.event': 'mouseleave'});
+              qtipApi.hide();
+          });
+      },
     checkMousePosition: function() {
-      jQuery('.qtip').qtip('hide');
+        this.hideWindowAnnotations();
```
Wedjaa: Hide only annotations belonging to this window. See above.

js/src/annotations/osd-canvas-renderer.js
```diff
@@ -329,6 +342,7 @@
     removeAnnotationEvents: function(tooltipevent, api) {
       jQuery('.annotation-tooltip a.delete').off("click");
       jQuery('.annotation-tooltip a.edit').off("click");
+        jQuery('.qtip-header a.pin').off("click");
```
Wedjaa: De-register the pin event handling.

js/src/annotations/osd-canvas-renderer.js
```diff
@@ -357,6 +371,19 @@
         display.remove(); //remove this annotation display from dom
       });
```
Wedjaa: This is where the magic happens. Toggle the status of the pin and of the actual annotation, so that it's no longer closed on `mouseleave`.

js/src/workspaces/window.js
```diff
@@ -594,19 +594,22 @@
         var dfd = jQuery.Deferred(),
         module = $.viewer.annotationEndpoint.module,
         options = $.viewer.annotationEndpoint.options;
-        if (_this.endpoint && _this.endpoint !== null) {
-          endpoint.set('dfd', dfd);
-          endpoint.search(_this.currentCanvasID);
-          // update with new search
+
```
Wedjaa: Sadly I created a branch from the master that contained the fix for the endpoint. This should be merged after that.

css/mirador.css
```diff
@@ -1538,6 +1538,11 @@ div.manifest-info a.mirador-btn.mirador-icon-toc {
   max-height: 185px;
   overflow: auto;
 }
+
+.qtip-header {
+   background: #cecede;
+}
```
Wedjaa: This sets a background for the header - it looks kind of empty; I wish I could find a non-hacky way of pushing the first edit and delete buttons into the header.

js/src/annotations/osd-canvas-renderer.js
```diff
@@ -357,6 +371,19 @@
...
 
+        jQuery('.qtip-header a.pin').on("click", function(event) {
```
Wedjaa: This is where the magic happens. Toggle the status of the pin and of the actual annotation, so that it's no longer closed on `mouseleave`.

js/src/workspaces/window.js
```diff
@@ -594,19 +594,22 @@
...
+        // One annotation endpoint per window, the endpoint
+        // is a property of the instance.
+        if ( _this.endpoint && _this.endpoint != null ) {
+          _this.endpoint.set('dfd', dfd);
+          _this.endpoint.search(_this.currentCanvasID);
```
Wedjaa: Sadly I created a branch from the master that contained the fix for the endpoint. This should be merged after that.

aeschylus:
I'll have to run this by our UX team and Rashmi. We have been talking about a similar feature for awhile now, but so much is going to change after the ongoing Ian Gilman POC integration that this is probably not going to be integrated into master. 

I will nevertheless show this to our UX team for feedback.

aeschylus:
After considering this in a recent design meeting, we have decided to close the PR. The transcription branch changes annotations specifically, so we have no reason to leave this open. 

Thanks for the help, and I hope the approach was working for you locally.

PR #546, opened 2015-05-15
Fix metadata links

Wedjaa:
It could be that I'm not understanding what is being done here, but inverting the value and the link I get clickable, linked URLs. I read this as: if it's a link, create a URL if you can; the way it's written it looks: if it's a link, let it be; otherwise create a URL.

  In my specific case I want the links - like the `within` - to be clickable.


mikeapp:
Careful, the value is already an HTML link, properly constructed using the anchor text provided by the user, while the method call is looking to convert URLs in text to clickable links.  In short, the change will break things.

Wedjaa:
Ok - then my manifest could be broken. I have a `within` value with the URL of a search that generated the results (IE: 'http://search.frick.org/search/id"). When that value is displayed in the metadata panel, without this change, is displayed as plain text. With this fix - since it's in links - it will be clickable.

  I was under the impression that the content of the `within` tag was supposed to be a URL - which made sense. What value should I use to have the `within` link clickable in the metadata view? Should it be some HTML <a> element?

aeschylus:
Can you post a link to a test manifest?

Wedjaa:
The manifest looks something along the lines of this: https://gist.github.com/Wedjaa/164b287c82d874522943

The IIIF specification says that `within` is supposed to be a URI - and in the example it shows: `"within":"http://www.example.org/collections/books/"` - this is somewhat consistent with the code I saw in Mirador: `_this.getMetadataLinks` will simply take `within` and keep it as a string.

I think that the proper solution would be to treat `within` like `related`, even if the specs are not as specific, and allow using a label in it, by using `stringifyRelated`. 

Does it make sense?

azaroth42:
Yep, `within` should have either a URL or a list of URLs as its value, not
HTML.  HTML is allowed in description, attribution and in the value fields
of metadata pairs, but that's it.

Rob



-- 
Rob Sanderson
Information Standards Advocate
Digital Library Systems and Services
Stanford, CA 94305

Wedjaa:
Yep - makes sense - will move the link to the related field, where it belongs!

PR #548, opened 2015-05-26
Adds support for internationalization and localization via i18next

mattmcguirk:
This pull request adds the i18next library, creates a global handlebars helper to translate strings into the  appropriate locale and updates the unit tests i18n/localization. 

### Notes ###

* The Handlebars helper is registered in the Viewer module to make sure that all other loaded modules have access to it. This seemed more elegant that putting in the global name space while still basically achieving the same thing and not repeating it multiple times. 
* Tests will have to be updated if new Handlebars templates are added by adding the Handlebars helper to the beforeEach() function of each test suite.


rsinghal:
Need to look a bit more at this, but so far it looks great.  I think we need documentation on the wiki about 1) how to add a new language, and 2) how to update the locale files if we add new text.  Is there a way to test for missing key/value pairs in any of the locale files?

rsinghal:
And what happens if a translation is missing from a locale file?

mattmcguirk:
As far as I know, there is no way to test for missing strings from a locale file, aside from keeping the English locale canonical as far as containing every string and comparing against that. They're just JSON files, so perhaps there would be a way to do this programmatically? In any event, it's outside of the capabilities of the i18next library itself.

If a string is missing from a locale file, the application will display the "key" part of the key/value pair.

mattmcguirk:
I should add that in some cases the helper is being passed a variable rather than a key to translate. In those cases, the value of the variable will be displayed if there is no corresponding translation pair.

azaroth42:
Is there a demo site with this branch can be played with?

rsinghal:
@mattmcguirk Can you add documentation about adding a new language to the wiki?  Merging now.

mattmcguirk:
Will do. Thanks.


--

Matt McGuirk

rsinghal:
@mattmcguirk I didn't realize this because I tested from my local machine, but the locales need to be copied to build/mirador, and the code needs to look for them in there, not at the top level.  Please fix ASAP

rsinghal:
I have reverted this for now, please submit a new PR when this is fixed.

PR #554, opened 2015-06-11
Update French locale with metadata strings

regisrob:
As a side note: capitalization on metadata labels is not well suited to French strings ([mirador.css#L1377](https://github.com/IIIF/mirador/blob/master/css/mirador.css#L1377)). E.g. "Etablissement De Conservation" looks a bit strange.


rsinghal:
FYI, @regisrob, I had to revert this until there is a fix to the build with i18next in Mirador.

PR #555, opened 2015-06-11
Redesign of manifest panel (first attempt)

regisrob:
This PR is just a first proposal. Any feedback is welcome. 
Some aspects regarding responsive design still need to be addressed.

**Changelog:**

**# Panel layout and styles:**

- [x] Move the input controls above the list
- [x] Align the filter input with the content that it filters
- [x] Remove the redundant placeholder text from the filter input
- [x] Separate the load feature to the right
- [x] Get rid of background transparency
- [x] apply box-shadow to load window [1]
- [x] add dynamic padding-bottom to the list of items ([manifestsPanel.js#L33](https://github.com/regisrob/mirador/blob/manifest-panel/js/src/viewer/manifestsPanel.js#L33)) [2]

[1]: I've replaced the dark stripped background with a gray one ; the main reason was to make the box-shadow effect visible
[2]: A fixed 30px padding-bottom was not suited to the situation where the input controls are above each other. I did not manage to handle that only with css.

**# Items:**

- [x] Wrap the item title to 3 lines (using height and line-height ; according to the "Fade Out" way: https://css-tricks.com/line-clampin/)
- [x] Give the title a min and max width to maintain readability (width:45%; max-width:450px; min-width:300px;)
- [x] Replace the whole "z more" block with a subtle ellipsis (&hellip; in Handlebars template)

What still needs to be done:
- [ ] add an ellipsis to multi-lines item title => the "Fade Out" technique by css-tricks does not work with background-color on hover
- [ ] add a clickable "more…" link that increases the height of the row
- [ ] set minimum threshold (equivalent to xs device) where the images finally wrap below the text content => I guess this should be part of the work on mobile optimization ; and I suspect Drew already thought about that (see [manifestsPanel.js#L87](https://github.com/IIIF/mirador/blob/master/js/src/viewer/manifestsPanel.js#L87))

NB for devs: in [ManifestListItem.js](https://github.com/IIIF/mirador/blob/master/js/src/viewer/manifestListItem.js): 
- I've kept the code blocks concerning the old "remaining items" feature, I've only commented those involving DOM manipulation
- I've only tweaked the [maxPreviewImagesWidth](https://github.com/regisrob/mirador/blob/manifest-panel/js/src/viewer/manifestListItem.js#L33) setting (by excluding "this.remainingItemsMinWidth" from the calculation)


regisrob:
@rsinghal this PR does not include the last merge in master (Update French locale with metadata strings)

rsinghal:
This is looking really great, @regisrob.  I'm not sure if I love the new gray background, but that could just be because I'm so used to the old color.
This was an issue before your PR, but we need the manifests panel adjust the thumbnails when the screen is resized.  Before, we had the "more" icon expand when the screen was made bigger, but didn't make any adjustments when the screen was made smaller.  Ideally, we would simply adjust the number of thumbnails displayed so it always stays on one line.
It would also be nice to have a rotating icon when adding a new object from URL to indicate Mirador is working.

snydman:
Agreed, this is really looking promising @regisrob.  I generally think all of mirador needs some attention to the color palette which can be addressed during a more holistic design review and update. Overall the app is very high in contrast and could use more subtlety.  Not something that needs to be solved in this PR.  More important, I agree is the need for improved responsive behaviors on smaller screens.  The new layout certainly lends itself to this.

regisrob:
I'm not sure either for the gray background... Maybe it's not the right time to start retouching the original chrome (at least it should be part of an overall design review as Stu pointed out). For now we could simply keep the dark one and add a white border to the manifest panel (instead of the box-shadow effect). 
Agree with the idea of adjusting the number of thumbnails when the screen is resized.

regisrob:
I forgot the link to the demo for the watchers: http://dev2.biblissima.fr/mirador/

rsinghal:
@regisrob Can you submit a PR for your work to the manifest-panel branch?  I will work on merging it with all of the changes that have been made to master and see if we can finish up the last things on your list. Thanks!

regisrob:
@rsinghal Sure, thanks Rashmi. I will review the local branch and submit a PR ASAP, hopefully before Ghent

rsinghal:
@regisrob Thanks.  Since there will likely be a lot of conflicts, I think having it be in a branch in the main Mirador repo will make it easier for me to merge those conflicts.

rsinghal:
Some designs in #287

PR #562, opened 2015-06-12
Logo location

rsinghal:
be able to change location of logos through settings.  this is still horribly brittle and should just not exist.


azaroth42:
+1 to being horribly brittle, and (perhaps controversial) unnecessary given logo in the manifest.

aeschylus:
I would advocate for just removing it at this point, now that the logo can
be read from the manifest. It's only in there for demo purposes/historical
reasons.

PR #579, opened 2015-07-07
Increase to 2.0.0 release

edsu:
When building the latest from master I noticed the mirador.js still said v0.9.0 at the top instead of v2.0.0.


aeschylus:
Thanks for this catch. Merging.

PR #580, opened 2015-07-07
Update internationalization support

regisrob:


regisrob:
Isn't it ready to be merged?

aeschylus:
Hm. I believe Rashmi did this manually, and it didn't close the PR with a pull request. Let me get in touch with her about this. We'll see if we can't get this merged in an hour or so.

aeschylus:
@rsinghal

rsinghal:
Sorry, missed this PR.  Looks like this has some new translations.  Do we know who is managing the Chinese translations so we can get those updated as well? @mattmcguirk

aeschylus:
Matt says two people at the 1000 Rooms project are managing the Chinese translations. Can we merge this without Chinese and add it later?

rsinghal:
Yes, but we need to figure out a good way to alert those who manage each language when they need to update the translations, otherwise files will get out of date.

regisrob:
Good point. 
No problem if you want to wait a little bit for the Chinese translations (there's only one new string if I remember well: "objectMetadata")

mikeapp:
@mattmcguirk  can add any new strings to the Transifex site, which will ping the contributors for each language.

PR #611, opened 2015-08-06
Annotation Sources Showing in Annotations Tab

sdellis:
Work in progress... still need to resolve hidden Annotation Sources when Manifest has a TOC.


aeschylus:
Tabs and other events are being re-bound every time a new object is loaded into the window. Refactor window to not call its own init function.

PR #633, opened 2015-08-12
editor toggle now works

sdellis:


aeschylus:
I think this one can be merged. Looks good and creates no bugs that I can see testing it locally. Let's write some tests for this soon.

aeschylus:
Add more commits to this one in order to reopen it. Comments in code.

aeschylus:
[here](https://github.com/IIIF/mirador/pull/633/files#diff-1bb2d28769af970c5242c3925aa2ca0cR86) is where we should update the list. The state update should call a handler that compares the page data and updates the local state if it has changed. 

Alternatively, (and more granularly) the editorPanel can listen for the page turn event and set itself to a blank/ "loading" template, and then when the list is loaded, re-populate in response to the state update event from annotationsTab.

js/src/widgets/editorPanel.js
```diff
@@ -92,6 +97,7 @@
             jQuery.subscribe('openAnnotationList.' + this.windowId, function(event, data) {
                 _this.openAnnotationList(data);
             });
+
```
aeschylus: We might want to create a new events listener here for the page turn in order to clear or invalidate the rendering.

sdellis:
Ok, thanks for the tips.  Let me see what I can do on this front tomorrow.

PR #634, opened 2015-08-13
Transcription issue 624

### LINKED ISSUE #624: Clicking item in Annotation List tab should highlight all annotation regions in that list

sdellis:
Fixes #624 
Highlighting now works and this is how:

1. window.getAnnotations normalizes the annotation object resource to be the same for manifest and endpoint annotations.  Endpoint annotation resources were coming in as an object in an array, but now both are objects.
2. editorPanel.openAnnotationList publishes an array of the selected annotation list (annotationsListFiltered) making it easier for other components to use without filtering by listId.  (Had to wrap array in object as jQuery pubsub doesn't support passing arrays.)
3. annotationsLayer listens for the annotationsListFiltered event and uses that filtered array as the OsdCanvasRenderer's annotationsList for rendering update

### DIFF
--- a/js/src/widgets/editorPanel.js
+++ b/js/src/widgets/editorPanel.js
@@ -68,6 +68,9 @@
 
               return false;
             });
+            var annos = { data: state.annotations };
+            console.log(annos);
+            jQuery.publish('annotationsListFiltered' + this.windowId, annos);
 
             state.open = open;
             this.state(state);
--- a/js/src/widgets/editorPanel.js
+++ b/js/src/widgets/editorPanel.js
@@ -69,7 +69,6 @@
               return false;
             });
             var annos = { data: state.annotations };
-            console.log(annos);
             jQuery.publish('annotationsListFiltered' + this.windowId, annos);
 
             state.open = open;
--- a/js/src/workspaces/window.js
+++ b/js/src/workspaces/window.js
@@ -628,12 +628,11 @@
           });
 
           jQuery.each( _this.annotationsList, function( key, value ) {
-            //console.log( jQuery.isArray(value.resource) );
+
             if(jQuery.isArray(value.resource)){
                _this.annotationsList[key].resource = value.resource[0];
             }
           });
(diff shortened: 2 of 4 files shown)


js/src/widgets/editorPanel.js
```diff
@@ -68,6 +68,8 @@
 
               return false;
             });
+            var annos = { data: state.annotations };
+            jQuery.publish('annotationsListFiltered' + this.windowId, annos);
```
aeschylus: Should the editor tab be publishing this? Or should the annotationsTab be publishing it?

js/src/workspaces/window.js
```diff
@@ -603,7 +602,7 @@
         module = $.viewer.annotationEndpoint.module,
         options = $.viewer.annotationEndpoint.options;
         name = $.viewer.annotationEndpoint.name;
-        console.log("endpoint: " + _this.endpoint);
+
```
aeschylus: Thanks for cleaning up the console logs.

js/src/workspaces/window.js
```diff
@@ -627,9 +626,17 @@
             }
             return true;
           });
+
+          jQuery.each( _this.annotationsList, function( key, value ) {
+
+            if(jQuery.isArray(value.resource)){
+               _this.annotationsList[key].resource = value.resource[0];
```
aeschylus: Nice fix. Now we don't have to worry about these differences, since (I think) they are both valid.

js/src/widgets/editorPanel.js
sdellis: I struggled with which module should be publishing this too.  The reason I stuck it there was because that's where the list filtering logic already lived.  I was thinking that the annotationsLayer might provide an easier way to jump among lists as well (and that manages the non-filtered list data), but that would have required a decent amount of rewriting.

rsinghal:
@sdellis @aeschylus Since this can't be merged automatically, do you want to merge it manually or close the PR?

PR #638, opened 2015-08-13
Revert "editor toggle now works"

aeschylus:
Reverts IIIF/mirador#633


aeschylus:
I am reverting this because I realise this is where best to discuss the page turning problem.

PR #642, opened 2015-08-27
Reading Use Case Done

sdellis:
The style improvements to the "editor panel" are the only things left to do for the "Reading Use Case".  Might be nice to also highlight regions when a list is selected so users don't also have to click the speech bubble icon to view the regions?


rsinghal:
@sdellis @aeschylus Since this can't be merged automatically, do you want to merge it manually or close the PR?

sdellis:
@rsinghal I think it depends.  A lot has changed with regard to working with annotations since this PR.  The editor panel listed all Annotations for a given endpoint and allowed inline editing, except I crapped out when it came to actually writing the annotation back to the server, which was when annotations were tightly coupled to the jQuery popup.

I think the "detail panel" could be helpful in conjunction with the tabbed interface, as an "editor" panel for an Annotations tab or a "settings" panel for a Tools tab.  I don't have a lot of bandwidth to add this in manually at the moment, but it might be a good task for a pair programming session during the Mirador session in Ghent, for which I would be available. 

I would be fine with closing it if you want to clean things up.  I have all the code in my repo, if we want to refer back to it later.

rsinghal:
@sdellis My main goal is to clean up these PRs.  Since I have done quite a bit of work with the other developer at HX, we have made it possible to edit annotations within a tooltip or in a side panel and have the changes reflect in the other.  I would suggest closing these PRs and coming back to this at a later date with the code in your repo as well as the work we did at HX.

sdellis:
Sounds good to me, @rsinghal!

PR #645, opened 2015-09-10
Fix for #644 - userButtons cleanup

### LINKED ISSUE #644: Inline doc for userButtons is incomplete, also callback property is probably worthless
Inline doc doesn't cover `iconClass` property available for buttons.

In retrospect, also, many callbacks get attached to elements based on class or other attributes, so the whole "specify callback to throw an error if id is omitted is kind of theatrical rather than actually doing anything useful in the way of preventing errors.

If there's an interest in including some sort of mechanism for enforcing structure of buttons, it probably ought to be more like:

```javascript
"mainMenu": {
  "userButtonsRequiredProperties":  ["id", "attribute_x", "attribute_y"]
}
```

...but frankly, this is JS, this seems totally unnecessary in the light of day.  I'm submitting a PR that removes `callback` support and documents `iconClass`

pobocks:
Fixes #644 
  * Remove `callback` property support
  * Document `iconClass` support in comment

### DIFF
--- a/js/src/viewer/mainMenu.js
+++ b/js/src/viewer/mainMenu.js
@@ -112,14 +112,16 @@
      *          exist and MUST contain an "id" value    *
      *   li_attributes: HTML attributes to add to the   *
      *          list item containing the button.        *
+     *   iconClass: class or space-separated list of    *
+     *          classes. If present, an empty span with *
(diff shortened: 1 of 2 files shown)


pobocks:
Thanks!

PR #659, opened 2015-10-02
Drag n drop manifest

aeschylus:


css/mirador.css
```diff
@@ -531,16 +531,34 @@ li.highlight {
...
 .addItemLink {
+    display:block;
+    position:absolute;
+    top:0;
+    bottom:0;
+    right:0;
+    left:0;
+    cursor: pointer;
+}
+.dropMask {
```
aeschylus: Not sure why it's showing the diff this way, but I added CSS for the dropmask and so on, for the different states of the drop interaction.

js/src/manifests/manifest.js
```diff
@@ -25,6 +52,18 @@
...
+    initFromInfoJson: function(infoJsonUrl) {
+      var _this = this;
+      this.request = jQuery.ajax({
+        url: infoJsonUrl,
+        dataType: 'json',
+        async: true
+      });
+
+      this.request.done(function(jsonLd) {
+        _this.jsonLd = _this.generateInfoWrapper(jsonLd);
```
aeschylus: The idea here is to swap out the manifest request for the info.json request and the actual manifest for a dummy manifest structure created as a wrapper around the info.json as its service. The dropped image is the one service on the one resource on the one image on the one canvas in the one sequence of the dummy manifest.

aeschylus:
@rsinghal 

Hey Rashmi. I'd like to start reviewing code before merging, so I'll wait for you to have a look and you can merge once you've had a chance to get any questions answered.

More info on demoing/testing here: https://groups.google.com/d/msg/iiif-discuss/lrR8OJMl_qk/wamI9j02CwAJ

rsinghal:
@aeschylus The drag and drop is working for me, but I noticed this branch no longer stores the session to local storage.  Did you disable that?

aeschylus:
Ah, yes, I forgot about that. I think there is a bug in master concerning the save system. When I pulled the fresh master, the app wasn't loading and the error was caused by [these lines](https://github.com/IIIF/mirador/pull/659/files#diff-288efcdffef147e82e36d3a1070c4e9eR82). Rather than debug it, I commented it out for the time being. I can uncomment those lines, but I think this needs to be dealt with in another branch.

I will try to reproduce it and create another ticket for it.

PR #660, opened 2015-10-02
non-iiif image support

thehabes:
This version includes the image manipulation tools and workarounds for non-iiif images.  The workarounds run in parallel with the rest of the code, so working IIIF images ignore the workaround so there should be no changes for working manifests (same for the annotations). *Note: only oa:commenting annotations that do not need to be drawn will work with non-iiif images since OSD draws the annotations, but for non-iiif images OSD does not receive rectangle data and cannot draw the annotations.)

For a working demo, visit http://165.134.241.141/brokenBooks/demoManifest.html

I also did some work with ranges and the TOC module.  There can be multiple "root" ranges now, so ranges can exist without being grouped together and they will display (for example, three pages from Chapter 3 without an aggregating chapter 3 range).  

The only pitfall is that I could not attach OSD tools to the OSD instance I create without a IIIF image, but they come through with all other tools besides zoom and pan.


azaroth42:
No client side zooming?  Did LegacyImageSource not work?

thehabes:
The problems come when I have to do the following in imageView.js

_this.osd = $.OpenSeadragon({
          'id':           osdID,
          'tileSources':  [], //This is the consequence of not getting the JSON.  It creates the viewport on  which the OSD functions are called.  Without it, OSD does not work.
          'uniqueID' : uniqueID
         
        });

When someone does not give a IIIF image, then jQuery.getJSON(infoJsonUrl) fails and there are no tile resources.  This is the only difference when making an OSD instance without a IIIF image, but it means OSD cannot create the actual osd.canvas or the osd.viewport object and cannot draw annotations.  It also fails to attach zoom and pan for the images since there is no real osd.viewport to use them on.  I assume this happens in the Open SeaDragon code, which I did not alter.  This is the wall I have hit and need help from the community on attaching this functionality.

azaroth42:
Did you actually try using: https://openseadragon.github.io/examples/tilesource-legacy/ ?

thehabes:
In our application, users upload an image by just giving us the URL of where it is hosted.  This could be dropbox, a user's own image server, a IIIF image server, etc.  We cannot expect the server will respond when we ask for an image with anything but that image, and it may not necessarily respond with a JSON object that tells me height and/or width.  We take that URL and create an image resource out of it and attach it to a canvas's canvas.images[] field.  At that point, that is the only thing I know from a user.  I could create the pyramid inline, but  I would only be able to make one level with one image, and I would have to lie to set height and width.  If I am wrong about this, let me know.  To answer the question directly, I did not know this existed and I did not try it.  I may be able to have a go at it today.

azaroth42:
So just wrap the image from @id in the manifest in the Legacy Tile Source JSON and hand it to OSD?  I haven't tried this either, FWIW, but that was the approach that Ian recommended.

```
{
    type: 'legacy-image-pyramid',
    levels:[{ url: canvas.images[x]['@id'] }]
}
```

thehabes:
Thanks for help with this approach, I was able to gain a lot of ground on this!  Here are my findings...
When I do not have a IIIF image to resolve, I define the osd like so:

_this.osd = $.OpenSeadragon({
          'id':           osdID,
          'tileSources':  [{
              'type': 'legacy-image-pyramid',
              'levels':[{ 
                'url': imageUrl,
                'height': 1000, //this is a lie
                'width': 885 //this is a lie
              }]
            }], 
          'uniqueID' : uniqueID
        });

With a little extra work, I could get the canvas from which the imageUrl comes, but I have also lied about the height and width of the canvas since at the time of creation, the user was unable to give me a height and width (which will be common in the application).
**There will never be any annotations drawn onto these canvases within the application besides the image resource.  It is a one-button click action that creates a range with two canvases inside that have holder image resources.  I do not know any information (like height or width) at this point, the user fills that out later IF they know the information.  Most times, users will not.  

I am able to get zoom and pan functionality, but my images look funny since the height and width were taken very literally.  When i do not supply a height or width or give height:0, the images will not load to the screen saying 

[Viewport.setHomeBounds] bounds.height must be greater than 0.
**A fix here:  resolve the image and use its natural height/width to set these bounds in the OSD code. The zooming will at least zoom to the level served by the server.  
**Another fix:  In my application when a user supplies an image, store its resolved height/width as the canvas height/width.  However, this will not solve this problem in Mirador for users with similar issues.
**Anotherother fix:  Lie better.  

For the demo that is up, I simply lie better so most the images look OK, but realize when you look through it that I Iied to Mirador so one or two images do not look OK.  

For development I plan to set the image's resolved height/width as the canvases height/width when a user saves an image so I can use it here, and that should take care of my funny looking image problem.  

The next situation is that if the numbers I lie with exceed the actual height/width of the image, zooming results in blurry/pixelated images (which is expected, but I tested it for fun anyway).  However, giving the canvas the height/width of the resolved image would solve this.  

The existing version with working zoom and pan is still at
http://165.134.241.141/brokenBooks/demoManifest.html

aeschylus:
Thanks for this Bryan. I think this is a useful demo, so I'm going to keep it here for reference, but close the PR. In particular the controls will be interesting to integrate once the sidepanel is added.

thehabes:
The new working version is at http://165.134.241.141/brokenBooks/getManifest.html

PR #665, opened 2015-10-09
moves annotations layer button to top left and replaces it with conte…

rsinghal:
…xt controls if there is an endpoint, otherwise sets the button to active


rsinghal:
@aeschylus can you merge when you have a moment?  I changed the design of the annotation buttons a bit.

aeschylus:
Sure, I'll have a look.

PR #666, opened 2015-10-12
added tabbed side panel

sdellis:
This is a subset of the recent, incomplete transcription work that allows for multiple content areas in the side panel through a tabbed interface.  It currently has an "indices" tab and an "annotations" tab.  To add a tab, one must edit the the tabs.js rendering template.  I'm not sure if the maintainers would prefer tabs to be set in the configuration object, or if the appearance of certain tabs should be based on other criteria.

### DIFF
--- a/css/mirador.css
+++ b/css/mirador.css
@@ -4,7 +4,7 @@ a {
 }
 
 /* jquery-ui-dialog theming/customisation (overrides to reset neutral defaults)
----------------------------------------------------------------------------- 
+----------------------------------------------------------------------------
  */
   .ui-dialog-titlebar {
     display:none;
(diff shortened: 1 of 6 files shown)


sdellis:
@rsinghal, these are my plans for additions re: our talk at the standup today:

1. Hide any tabs with no content
2. Hide title if only one tab is present
3. Only display Annotations tab in "image view" (disable in thumbnails and book view)
4. Tabs except for TOC should be suppressed by default. Add settings to configuration object to display tabs.

I have a note to disable icon in titlebar, but I'm not sure when.  I think when there is no tab content (including toc) and tabs are suppressed.  The "minimize" tab setting should apply to all tabs, correct?

sdellis:
@aeschylus not sure if you want to chime in on this as well...

rsinghal:
Yes, I think when a user minimizes the side panel, it should minimize all tabs (treating them as a single unit).

Currently, if an object doesn't have a TOC then the side panel icon is disabled, so we could continue that to include annotations, but I can see the argument to have a different UI change to show the user that there is nothing to populate the side panel.

PR #684, opened 2015-11-04
tabbed interface - take 2

sdellis:
This approach is a bit different than we had discussed in #666 in order to get the change in quickly (but without disruption) as a starting point for adding tabs to the side panel.  Knowing if a tab is empty and should be hidden would require the tabs.js widget to listen for all possible tabs' state changes, which starts to look a bit like a "plugin" type pattern since the listener may know know what to listen to without its own code getting updated with every possible tab.  But this PR stops short of a full plugin implementation since that requires more "big" decisions than I feel comfortable making on my own, although I could easily see a "tabs plugin" directory that builds a tab list based on the tab widgets that exist inside.  

Instead, I took a simplified approach so we could get this accepted/merged before we start getting conflicts due to other commits.  If you want to add a tab using this code, there are two steps:

1. Add additional tabs through the tabs.js state object like so:  
    tabs: [{id:'tocTab', label:'Indices'}, {id:'annotationsTab', label:'Annotations'}],
2. Add a "tab" widget that has an id property that matches the id supplied to the tabs.js widget (see annotationsTab.js)

To hide a tab, just remove it from the list referred to in step 1 (the tocTab is currently the only one).  You will not see the tabs until you add one since it hides the tab/label if there is only one tab.  I understand that the "tabs" array I refer to in step 1 could be defined in the configuration window object, but I'm struggling to remember how to set that up.  I don't mind adding it, but need some guidance.

### DIFF
--- a/css/mirador.css
+++ b/css/mirador.css
@@ -1,8 +1,8 @@
-body { 
-  padding: 0; 
-  margin: 0; 
-  overflow: hidden; 
-  font-size: 70%; 
+body {
+  padding: 0;
+  margin: 0;
+  overflow: hidden;
+  font-size: 70%;
 }
 a {
   color: inherit; /* blue colors for links too */
@@ -10,7 +10,7 @@ a {
 }
 
 /* jquery-ui-dialog theming/customisation (overrides to reset neutral defaults)
----------------------------------------------------------------------------- 
+----------------------------------------------------------------------------
  */
   .ui-dialog-titlebar {
     display:none;
@@ -59,7 +59,7 @@ a {
 
 /*----------------------------------------------------------------------------
 /* Mirador styles
----------------------------------------------------------------------------- 
+----------------------------------------------------------------------------
  */
 
   .mirador-viewer, .mirador-main-menu-bar, .mirador-status-bar {
@@ -213,7 +213,7 @@ a {
 }
 
 
-/* main menu select menu 
+/* main menu select menu
 ---------------------------------------------------------------------------- */
   #manifest-select-menu, #workspace-select-menu, #workspaceContainer {
     position: absolute;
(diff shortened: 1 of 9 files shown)


aeschylus:
Taking a look at this now.

sdellis:
Great! Let me know if you have any questions or suggestions.

rsinghal:
@sdellis The side panel is still visible (without the option to hide it) if you are looking at an object that doesn't have a TOC.

When there is a TOC the animation for minimizing it looks a little weird - anyone else getting that?

screen shot 2015-11-05 at 11 19 22 am

sdellis:
Nice catch, @rsinghal.  I will fix that.

sdellis:
Ok, just looking back into this.  The goal was to allow the inclusion of any number of tabs in the panel, which have content panes (aka, widgets like toc.js and annotationsTab.js).  

What I got stuck on, and what is necessary to know if a pane is empty, is this:  How does the sidePanel.js know what to listen for if the widgets are not predetermined?  Should there be a tab "plugin" convention that after a tab widget is instantiated, it emits a generic "tabbedContent" event with the widget id, and then the sidePanel can disappear if all tab panes are empty?  I want to be clear that we are not talking about setting the "open" property to false, but removing the option of toggling/maximizing the sidePanel at all.  I have having a hard time tracking down the code that does this currently when there is no toc or other tabs.

Any pointers or feedback?

sdellis:
I just want to follow up on this PR to clarify expectations.  @aeschylus asked me to contribute back the tabbed interface portion of the transcription work as others would like to add tabbed content and this provides a starting point.  I don't have much incentive to build out a "plugin" approach here since my project is done and I have to spend my time on other things at the moment.

I have provided the code we worked on with some additional features that @rsinghal suggested.  Unfortunately, after some attempts I'm having trouble hiding the SidePanel when there's no content in the TOC.  This is almost ready for someone to run with if they want to build out a plug-in approach here, but requires a bit more discussion before further coding can happen.  

The purpose was to allow people to add tabbed content, but I don't know what the term "allow" means to the maintainers and what the expectations are. Now there's a merge conflict since a number of PRs have been merged since. I would like to see this either merged or scrapped.

rsinghal:
@sdellis Sorry for the delay.  We are at a conference this week, but I plan to work on this later this week and early next week to help you resolve some of the issues you are having.  The side panel is really important and we want to get it into Mirador ASAP.

sdellis:
Thank you, @rsinghal.  Ping me when you're ready, and I'd be happy to work with you to wrap this up.

rsinghal:
@sdellis @aeschylus this is nearly ready.  Let's get the tests in before merging with master.

aeschylus:
I'm working on the tests for this today. Shouldn't be too long. I will force push them to this branch, so they update the PR here. @sdellis, we can catch up on this on the standup today if you're able to join. Basically we finished off the last bits of weirdness last Thursday.

sdellis:
@aeschylus , I have been really busy today getting things wrapped up before the break for a number of projects and Code4Lib.  I'm sorry I couldn't make the standup.  I will make a point to be on next week, and I will take a look at the code. I'm curious as to how @rsinghal solved it.

rsinghal:
@aeschylus What is the status of the tests?

rsinghal:
@aeschylus I don't want this PR to get out of sync with master, and since we are going to be refactoring to make testing easier, I don't think it is pressing to have tests in for this before we merge, so I'm going to go ahead with the merge.

sdellis:
:+1:

PR #689, opened 2015-11-12
Css fixes

rsinghal:
1. Continue to fix font sizes in mirador.css
2. Moved overlay div so it is a child of content-container instead of view-container.  This means the size of the overlay does not change if the side panel is open.
3. No longer use table-row and table-cell for items in MetadataView so the text doesn't get really scrunched up when multiple slots are open.
4. Correctly code for a metadata object.  Still defaults to English, for now.
5. Display logo image, rather than URL.


aeschylus:
Hm. Can you explain 2) above a bit more? Is it the metadata panel that's disrupting the layout of annotations somehow?

I'm ready to merge otherwise.

rsinghal:
If you look at an object with a table of contents on http://oculus-dev.harvardx.harvard.edu/demo/ and have the overlay open as well.  Minimize the left side panel and you will see the width of the overlay changes because originally it was a child of content-container and the width is set to 45%.  With this PR, the overlay element is no longer a child of content-container, so its width is not affected by the visibility of the side panel.  Now the overlay's width is basically determined by the size of the slot, not the OSD container.

PR #698, opened 2015-11-23
added sae module

aeschylus:
Just adds the module. Haven't tested if it works yet, though it should with no modification. Stanford set up a server for our own annotation demos at the end of last week, and we're planning on using SimpleAnnotationServer with Sesame (or possibly the as yet unperfected elasticsearch module), so I should know soon.

Configuration example here: https://github.com/IIIF/mirador/blob/transcription/index.html#L53-L61

### DIFF
--- a/js/src/annotations/simpleASEndpoint.js
+++ b/js/src/annotations/simpleASEndpoint.js
@@ -1,0 +1,188 @@
+/*
+* Edited version of https://github.com/IIIF/mirador/blob/9e3c6bbb894e044d01ad51aae1b70309939de5a9/js/src/annotations/catchEndpoint.js 
+* This module tries to store the annotation as is in a RDF store but some fiddeling is required. Fidles are:
+* 
+* - delete annotation fails if id has a / in it so have to send sanatised ids to mirador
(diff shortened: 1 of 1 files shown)


js/src/annotations/simpleASEndpoint.js
```diff
@@ -0,0 +1,188 @@
...
+          'update': [this.userid],
+          'delete': [this.userid],
+          'admin':  [this.userid]
+        }
+      };
+      this.search(this.uri);
+    },
+
+    //Search endpoint for all annotations with a given URI
+    search: function(uri) {
```
rsinghal: @aeschylus The biggest thing that needs to change is the search function will be passed JSON object of options, which includes the uri (which affects line 71).  This function can also take success and error callbacks, but not required.  See also [the catch adaptor in master](https://github.com/IIIF/mirador/blob/master/js/src/annotations/catchEndpoint.js#L63)

rsinghal:
@aeschylus What is the status of this?

aeschylus:
Finally got this working locally. There were some build and caching problems that were tripping me up.

Ready to merge.

PR #714, opened 2015-12-06
Added configuration options for JSONBlob

Wedjaa:
This PR will add the following options for the JSONBlob. This changes will allow the use of locally installed JSONBlob services. The plugin can be added to the configuration of Mirador like this:

```
'jsonStorageEndpoint': {
    'name': 'JSONBlob API Endpoint',
    'module': 'JSONBlobAPI',
    'options': {
        'ssl': false,
        'host': 'localhost',
        'port': '8080'
    }
}
```

This are the options:

*host* - is the hostname or IP where the service is running
*port* - is the port the service is listening to
*ssl* - can be *true* or *false*, this determines is http or https is used


aeschylus:
We should also adjust the default settings to have this (and hence, bookmarking) turned off.

PR #721, opened 2015-12-14
Accessibility Roles and aria-labels

lduarte1991:
@rsinghal: 
While using mirador within the HarvardX Annotation Tool, the HarvardX Accessibility Project Manager pointed out some simple labeling/roles issues that affected users with a screen-reader. In particular, the hud and window buttons were done using font-awesome meaning that screen readers would not actually know what it meant. Adding some simple aria-labels to guide the user through the buttons and adding the role "button" should let users with screen readers know that they can "click" on the buttons and know what action it will perform. 

Not necessarily comprehensive but I didn't want to touch any items that I couldn't directly see from the HxAT. 

Here is the zoom in button without an aria-label (i.e. what a screen reader would read):
zoom in button without aria-label
Here is what it should look like now with an aria-label:
zoom in button with aria-label


js/src/workspaces/window.js
```diff
@@ -766,11 +766,11 @@
                                  '</ul>',
                                  '</a>',
                                  '{{#if ThumbnailsView}}',
-                                 '<a href="javascript:;" class="mirador-btn mirador-icon-thumbs-view thumbnails-option"><i class="fa fa-th fa-lg fa-rotate-90 fa-fw"></i>',
+                                 '<a href="javascript:;" class="mirador-btn mirador-icon-thumbs-view thumbnails-option" role="button" aria-label="Change to Thumbnails Mode"><i class="fa fa-th fa-lg fa-rotate-90 fa-fw"></i>',
                                  '</a>',
                                  '{{/if}}',
                                  '{{#if MetadataView}}',
-                                 '<a href="javascript:;" class="mirador-btn mirador-icon-metadata-view" title="{{t "objectMetadata"}}"><i class="fa fa-info-circle fa-lg fa-fw"></i></a>',
+                                 '<a href="javascript:;" class="mirador-btn mirador-icon-metadata-view" title="{{t "objectMetadata"}}" role="button" aria-label="View Information/Metadata about Image"><i class="fa fa-info-circle fa-lg fa-fw"></i></a>',
```
rsinghal: Can you change this to "View Information/Metadata about Object"?  There will eventually be another info button that will allow users to view information about a canvas, so using Image in this description will be confusing.  And, that info button refers to Object-level metadata.

lduarte1991: Can do!

js/src/workspaces/slot.js
```diff
@@ -184,7 +184,10 @@
                                  // '<li class="add-slot-below"><i class="fa fa-caret-square-o-down fa-lg fa-fw"></i> {{t "addSlotBelow"}}</li>',
                                  // '</ul>',
                                  // '</a>',
-                                '<h1 class="plus">',
+                                 // accessbility-wise it's really bad practice to have h1 tags be used like this
```
rsinghal: Good to know.  Can you ticket this so we can change it?

PR #715, opened 2015-12-07
Redesign of manifest panel (revised PR #555)

regisrob:
This PR replaces #555, it is aligned with master branch.
It mainly includes changes in css and handlebars templates. I've removed the new light gray background from the first PR.
It remains a few tweaks and commented code in javascript that need to be reviewed or cleaned:
* [manifestListItem.js#L33](https://github.com/regisrob/mirador/blob/4e0e08aa4217ab80ac147db4f30ebb4deabc2c36/js/src/viewer/manifestListItem.js#L33)
* [manifestListItem.js#L39](https://github.com/regisrob/mirador/blob/4e0e08aa4217ab80ac147db4f30ebb4deabc2c36/js/src/viewer/manifestListItem.js#L39)
* [manifestListItem.js#L145](https://github.com/regisrob/mirador/blob/4e0e08aa4217ab80ac147db4f30ebb4deabc2c36/js/src/viewer/manifestListItem.js#L145)
* [manifestListItem.js#L169](https://github.com/regisrob/mirador/blob/4e0e08aa4217ab80ac147db4f30ebb4deabc2c36/js/src/viewer/manifestListItem.js#L169)
* [manifestListItem.js#L211](https://github.com/regisrob/mirador/blob/4e0e08aa4217ab80ac147db4f30ebb4deabc2c36/js/src/viewer/manifestListItem.js#L211)

See my comment on #555 for more details about the changes and what still needs to be done to reflect the mockups.

PR #726, opened 2015-12-17
New manifest gallery display

rsinghal:
This is mostly @regisrob's work from #555 and #715.  I cleaned up some commented out code and made sure that the number of preview thumbnails change as the screen size changes (and the ellipsis is added and removed as needed).

Regarding the manifest label - for now, I simply added that label to the ```title``` attribute of the ```h3``` element.  Visually, the label is cut off after 3 lines using ```overflow: hidden```, which shouldn't impact screen readers.  Adding the label to the ```title``` attribute allows a user to see the full label if they hover over the element.  This is not a solution on mobile devices, but there is so much that needs to be re-thought for mobile, that I didn't want to get bogged down by this.  

I thought about adding a more/less link to change the height so it displayed the full label, but there were some issues with that.  Until the manifest gallery is visible on the screen, it is not possible to get height, width, etc of ```display:none``` items.  We also set a min and max width for the label area, which means it's not even possible to hard code the calculations.  We can use a nicer looking tooltip library instead of using the ```title``` attribute, but programmatically adding a more/less link felt like it was more of a headache than it was worth, at the moment.  If having the 3 lines of text is still not good enough, then we can revisit this part of the manifest gallery.

One last thing to do is add "Load" to the translation files and update the template to use that.


aeschylus:
I think we should put it in and keep additional improvements on other branches/issues. It looks great. I really like how well the resizing works.

PR #747, opened 2016-01-29
Application state

rsinghal:
1. Create a global application state object that can be passed between modules
2. Remove references to a given module's parent.
3. More work can be done to remove function calls to a given module's child's functions, but at least the dependencies only go down the tree now.
4. New events are in ALL_CAPS
5. Tried to split out bindEvents and listenForActions.  Another check of all the files would be useful, since I may not have updated every file to follow this pattern.


aeschylus:
Boom!

aeschylus:
Cool. I will try to review this this evening. Can you leave a few comments in the commit through the github line comment system to guide me a bit? I am not going to be able to get to this until tonight.

aeschylus:
I mostly want to understand where/how the global object is being passed around.

aeschylus:
Also, I really think this is going to be super useful in the "state atom"/saveController https://egghead.io/series/getting-started-with-redux.

The library is 2kb and is made exactly for this.

js/src/mirador.js
```diff
@@ -5,8 +5,8 @@ window.Mirador = window.Mirador || function(config) {
   // application.
   Mirador.saveController = new Mirador.SaveController(config);
 
-  config = Mirador.saveController.currentConfig;
+  //config = Mirador.saveController.currentConfig;
 
   // initialise application 
-  Mirador.viewer = new Mirador.Viewer(config);
+  Mirador.viewer = new Mirador.Viewer({'state' : Mirador.saveController});
```
rsinghal: Global state starts here, pass the saveController object that is created on initialization of the object to the Viewer.

js/src/viewer.js
```diff
@@ -23,8 +23,10 @@
         'bookmarkPanelVisible': false
       },
       manifests:             [] 
-    }, $.DEFAULT_SETTINGS, options);
+    }, options);
 
+    this.id = this.state.getStateProperty('id');
+    this.data = this.state.getStateProperty('data');
```
rsinghal: Here are a few examples where we are getting state information from the saveController

js/src/viewer.js
```diff
@@ -77,43 +78,41 @@
...
 
       // add workspace configuration
-      this.layout = typeof this.layout !== 'string' ? JSON.stringify(this.layout) : this.layout;
+      this.layout = typeof this.state.getStateProperty('layout') !== 'string' ? JSON.stringify(this.state.getStateProperty('layout')) : this.state.getStateProperty('layout');
       this.workspace = new $.Workspace({
         layoutDescription: this.layout.charAt(0) === '{' ? JSON.parse(this.layout) : $.layoutDescriptionFromGridString(this.layout), 
-        parent: this, 
-        appendTo: this.element.find('.mirador-viewer')
+        appendTo: this.element.find('.mirador-viewer'),
+        state: this.state
```
rsinghal: Now we pass the state on to Workspace, which can access any information it needs, and pass the state on to Slot.

js/src/workspaces/window.js
```diff
@@ -103,7 +103,7 @@
         _this.currentCanvasID = _this.imagesList[0]['@id'];
       }
 
-      this.annoEndpointAvailable = !jQuery.isEmptyObject($.viewer.annotationEndpoint);
+      this.annoEndpointAvailable = !jQuery.isEmptyObject(_this.state.getStateProperty('annotationEndpoint'));
```
rsinghal: Here is an example of getting rid of a reference to the Viewer object, and using the global state.

js/src/utils/saveController.js
```diff
@@ -80,6 +80,48 @@
...
+      return this.get(prop, 'currentConfig');
+    },
+
+    get: function(prop, parent) {
+      if (parent) {
+        return this[parent][prop];
+      }
+      return this[prop];
+    },
+
```
rsinghal: A bunch of helper functions to get at certain pieces of information in the global state.

js/src/utils/saveController.js
```diff
@@ -127,6 +169,21 @@
...
+        jQuery.publish('annotationListLoaded.' + options.windowId);
+      });
+
+      jQuery.subscribe('WINDOW_ELEMENT_UPDATED', function(event, options) {
+        if (!_this.windowsElements) {
+          _this.windowsElements = {};
+        }
+        _this.windowsElements[options.windowId] = options.element;
+      });
+
```
rsinghal: Some items are not saved to localStorage because they take up a lot of space, and are only needed during runtime.  This information will be regenerated from a localStorage config, so there is no need to include it in that.

js/src/widgets/imageView.js
```diff
@@ -71,20 +78,195 @@
...
+      });
+
+      jQuery.subscribe('HUD_FADE_OUT.' + _this.windowId, function(event, elementSelector, duration, complete) {
+        _this.element.find(elementSelector).fadeOut(duration, complete);
+      });
+      //Related to Annotations HUD
+    },
+
+    bindEvents: function() {
+      var _this = this;
```
rsinghal: A lot of the actions on the HUD in ImageView is duplicated in BookView (except for the annotation-specific events).  This is a necessary evil, but the duplication will be removed when we use OSD to manage the switch between views.

js/src/workspaces/window.js
```diff
@@ -395,7 +451,8 @@
 
       if (this.sidePanel === null) {
         this.sidePanel = new $.SidePanel({
-              parent: _this,
+              windowId: _this.id,
```
rsinghal: There are a lot of examples of passing the window id and the global state from Window.  Widgets can be less dependent on Window, and simply publish events that trigger things in their Window.

js/src/viewer.js
```diff
@@ -250,8 +280,9 @@
       if (!slotAddress) {
         return;
       }
+
       var windowConfig = {
-        manifest: this.manifests[options.loadedManifest],
+        manifest: this.state.getStateProperty('manifests')[options.loadedManifest],
         currentFocus : options.viewType,
         focusesOriginal : options.availableViews,
         currentCanvasID : options.canvasID,
```
rsinghal: Here is an example where I did not refactor.  All of this information is coming from the initial configuration for any individual window objects.  We wrote Window so it sets all of the defaults on initialize, and overwrites those defaults by extending them with this configuration.  That means, we have to pass certain parameters, such as ```annotationCreationAvailable```, from Window to ImageView to Hud.  Ideally, Hud would get this information from the global state, but that means either 1) moving the default values out of Window and into SaveController, or 2) updating the SaveController with all of these values once Window has initialized (right now we only publish the value to SaveController if it changes from the default).

rsinghal:
I put in a number of comments - let me know if you have any other questions.

PR #748, opened 2016-02-01
New annotation tools

stanislavkostadinov:
Paper.js library was integrated within Mirador. All annotation shapes are rendered within a single canvas (front layer of the OSD). Five drawing tools were implemented: rectangle, circle, polygon, freehand and pin.


rsinghal:
@stanislavkostadinov Thanks!  I am going to look through this and merge it with my refactoring work as soon as possible.

aeschylus:
Very exciting. I look forward to diving in. One quick impression is that we should probably .gitignore `bower_components/` to clean up this PR for review.

rsinghal:
@stanislavkostadinov Can you add ```bower_components``` to .gitignore and update the PR?

stanislavkostadinov:
Done.

lunactic:
I would really love to try out this PR but I can't get the tools showing up. Do I need to set some specific settings for this to work?

stanislavkostadinov:
I'm using https://github.com/glenrobson/SimpleAnnotationServer as an annotation server and I these are my configurations:
```
"availableAnnotationDrawingTools": ['Rectangle', 'Ellipse', 'Freehand', 'Polygon', 'Pin'],
"annotationEndpoint": {
    "name": 'Simple Annotation Store Endpoint',
    "module": 'SimpleASEndpoint',
    "options": {
        "url": 'http://localhost:8888/annotation',
        "storeId": 'comparison',
        "APIKey": 'user_auth'
    }
},
```

lunactic:
thanks I got it to work now. 
Looks really nice :+1: 

I'll offer a few suggestions: 

- I find some of the icons a bit confusing (e.g. the pencil and the lightning)
- The behavior of the rectangle drawing tool is a bit confusing as it is created from the center. I would suggest to assume that starts drawing from a corner (we did something similar here: https://github.com/mbaertschi/DIVAServices-Spotlight/blob/master/web/app/modules/algorithm/diaAlgorithmHighlighter/tools/rectangleTool.service.coffee)

And I would also have a question:

I did not check yet the data model used here. But as there should be models for rectangles, circles and, polygons. Do you think it would be easily possible to retrieve this information from an external source?

To specify a bit in more detail: I'm currently working on a project that exposes Document Image Analysis Methods as a RESTFul WebService (see: https://github.com/lunactic/DivaServices). Some of these methods can produce rectangles or polygons as output (e.g. text line extraction methods). And I think it would be really nice if users could use such information as starting point for making annotations.

Sorry for getting so long ;-)

stanislavkostadinov:
_** I find some of the icons a bit confusing (e.g. the pencil and the lightning) **_-> I haven't spent much time on seaching icons. We can replace them with other font awesome icons (or use other icons).
_**The behavior of the rectangle drawing tool is a bit confusing as it is created from the center.** _-> It won't take much time to refactor that.
_**Do you think it would be easily possible to retrieve this information from an external source?**_ -> Theoretically, it is possible to import/export any of these shapes: polyline, polygon, line, circle, ellipse, rect and path. It is also possible/easy to add other drawing tools (e.g. regular hexagon). It only depends on the time (development time) and user requiremen

lunactic:
Thanks for your replies.
I think I will wait until this PR is merged and then probably create a fork from there and run some experiments.

Very nice features anyways!

js/src/annotations/osd-region-draw-tool.js
```diff
@@ -0,0 +1,424 @@
...
+    },
+
+    render: function() {
+      this.svgOverlay.paperScope.project.clear();
+      var _this = this;
+      _this.annotationsToShapesMap = {};
+      var deferreds = jQuery.map(this.list, function(annotation) {
+        var deferred = jQuery.Deferred();
+        if (typeof annotation.on === 'object') {
+          var shapeArray = _this.svgOverlay.parseSVG(annotation.on.selector.value, annotation);
```
rsinghal: @stanislavkostadinov For backwards compatibility (and to handle simple rectangular, non SVG annotations) this needs to be able to handle annotations where ```annotation.on.selector.value``` can look like this: ```xywh=2164,1161,695,368```.  This is the ```media-frag``` Fragment Selector: http://www.openannotation.org/spec/core/specific.html#FragmentSelector.

js/src/annotations/osd-svg-overlay.js
```diff
@@ -0,0 +1,544 @@
...
+    parseSVG: function(svg, annotation) {
+      var paperItems = [];
+      var svgParser = new DOMParser();
+      var svgDOM = svgParser.parseFromString(svg, "text/xml");
+      if (svgDOM.documentElement.nodeName == "parsererror") {
+        return; // if svg is not valid XML structure - backward compatibility.
+      }
+      var svgTag = this.paperScope.project.importSVG(svg);
+      // removes SVG tag which is the root object of comment SVG segment.
+      var body = svgTag.removeChildren()[0];
```
rsinghal: Every time I try to create a new annotation (say, rectangle) and save it, svgTag is null

rsinghal:
Regarding the rectangle drawing, I agree, it should start from the corner as that was the behavior we had previously.

I am looking for some alternative icons.  We may have to use something other than font awesome, since it doesn't seem to have anything that can better represent the different types of drawing.

js/src/settings.js
```diff
@@ -89,7 +89,9 @@
     ],
 
     'availableAnnotationDrawingTools': [
-
+      /** possible drawing tools:
+       * 'Rectangle', 'Ellipse', 'Freehand', 'Polygon', 'Pin'
+       **/
```
rsinghal: I think the default here should probably not be an empty array.  We could probably set it to all of them, and then a developer can choose to narrow down the options from there. @aeschylus what do you think?

js/src/widgets/contextControls.js
```diff
@@ -69,11 +175,37 @@
                                    '<a class="mirador-osd-edit-mode hud-control" role="button" aria-label="Make a new annotation using mouse">',
                                    '<i class="fa fa-lg fa-edit"></i>',
                                    '</a>',
-                                   '{{/if}}',
-                                   '<a class="mirador-osd-refresh-mode hud-control" role="button" aria-label="Refresh annotations">',
+                                   '<a class="hud-control draw-tool" style="color:#abcdef;">',
+                                   '|',
+                                   '</a>',
+                                   '{{#each tools}}',
```
rsinghal: There needs to be some error checking before calling the template.  If there are no available tools then it displays two horizontal bars next to each other, with nothing in between, and it is then impossible to create any sort of annotation because you can't select a shape to draw.  Perhaps, if ```availableAnnotationDrawingTools``` is empty, then we always default to at least rectangle?  @aeschylus, your input would be appreciated

js/src/annotations/osd-svg-overlay.js
```diff
@@ -0,0 +1,544 @@
...
+        tolerance: 5
+      }
+    });
+
+    // Initialization of overlay object.
+    this.tools = $.getTools();
+    this.currentTool = null;
+    // Default colors.
+    this.strokeColor = 'red';
+    this.fillColor = 'green';
```
rsinghal: These default values should be set in settings.js so a developer can change them through the configuration.  By default, fillColor should be empty, and strokeColor should be ```deepSkyBlue```, to use the values we used with the previous annotation creation functionality.

rsinghal:
screen shot 2016-02-02 at 2 34 50 pm

The ```Cancel``` and ```Choose``` buttons need to look a bit better, and perhaps follow the design of the ```Cancel``` and ```Save``` buttons on the annotation editor box

js/src/annotations/osd-svg-overlay.js
rsinghal: This seems to be an issue with how I send the annotation to my store (I am not using simpleASEndpoint).  I will investigate.

js/src/annotations/osd-svg-ellipse.js
```diff
@@ -0,0 +1,154 @@
...
+            overlay.mode = 'deform';
+            overlay.segment = null;
+            overlay.path = null;
+            var segments = hitResult.item.segments;
+            for (var idx = 0; idx < segments.length; idx++) {
+              if (segments[idx] == hitResult.segment) {
+                if (idx % 2 === 0) {
+                  document.body.style.cursor = "move";
+                } else {
+                  document.body.style.cursor = "url('../images/rotate.png') 12 12, auto";
```
rsinghal: Can you change this from the PNG file to the font awesome rotate icon? http://fortawesome.github.io/Font-Awesome/icon/repeat/

js/src/annotations/osd-svg-overlay.js
rsinghal: This is an issue with my endpoint code, and not your code.  Please ignore.

rsinghal:
If I move the mouse too quickly from annotation to annotation, the tooltip does not show up.  You have to move your mouse out of the annotation and then back into it in order to see it.  The mouse move event should be more sensitive to this.

rsinghal:
When editing a shape, it should show the editor/annotation text associated with that shape.  Sometimes, if you double click a shape, you are given a new editor box, but it doesn't save the new text you put in there, it just saves a new annotation to that shape with the same text as before.

js/src/annotations/osd-region-draw-tool.js
stanislavkostadinov: Is it appropriate to load 'xywh=`2164,1161,695,368`' and the update action transforms it as an SVG tag (SVG Selector) or 'backwards compatibility' means that the selector should be the same as the original (Fragment Selector

js/src/widgets/contextControls.js
stanislavkostadinov: If the default `availableAnnotationDrawingTools ` is the full set of tools we can ignore inappropriate developer settings like no array or empty array (else it is a sub-set of the tools).

js/src/annotations/osd-region-draw-tool.js
rsinghal: @aeschylus Can weigh in as well.  I think it's fine to transform it to an SVG tag.  In the case of manifest level annotations, Mirador can't modify the annotation stored in the manifest.  But if a user edits the annotation and saves it, then having it updated to an SVG tag in the database is acceptable, IMO.

stanislavkostadinov:
The idea of edit/delete/refresh buttons was some kind of shortcut which we can use for manipulation of shapes (without the editor of annotation text).
This is one way to delete a single shape from a set of shapes associated to 1 annotation.
If it should show the editor with edit & delete buttons - there will be 2 buttons delete:
- one for the annotation + all the associated shapes
- one for deleting shape (that button calls update functionality which removes selected shape from the set of associated shapes and saves the annotation with N-1 shapes)
Isn't it a little confusing to provide 2 sets of buttons at the same time?

js/src/widgets/contextControls.js
rsinghal: We still need to check for a bad value or empty array, since the developer could pass those in for ```availableAnnotationDrawingTools``` through the configuration

rsinghal:
There needs to be some association of the text with the shape, otherwise there is no context when you are editing a shape.  @aeschylus @beaudet @sdellis might have some ideas, but at the very least, it should show the annotation viewer, if showing the editor makes the UI too complex.

On a related note, the annotation viewer should probably now have ```Edit Text``` and ```Edit Shape``` buttons.  ```Edit Shape``` should be a shortcut to the clicks it currently takes to edit a shape and focus on the particular one you want to edit.  ```Edit Text``` would have the same behavior it currently has.

js/src/widgets/contextControls.js
stanislavkostadinov: e.g.:

```
    if (options.availableAnnotationDrawingTools) {
      options.availableAnnotationDrawingTools = $.grep(options.availableAnnotationDrawingTools, function(element, index) {
        return $.inArray(element, $.DEFAULT_SETTINGS.availableAnnotationDrawingTools) >= 0;
      });
      if (options.availableAnnotationDrawingTools.length > 0) {
        this.availableAnnotationDrawingTools = options.availableAnnotationDrawingTools;
      }
    }
```

(within viewer.js)

rsinghal: This doesn't account for when ```options.availableAnnotationDrawingTools``` is set to ```[]```

stanislavkostadinov: Yes, I meant if the default settings contain all of the tools.
`['Rectangle', 'Ellipse', 'Freehand', 'Polygon', 'Pin']`

stanislavkostadinov:
tools
Is there any ideas about the icons of the different tools? (I can change the lighting to a polygon, but in my point of view pencil look user-friendly.)

rsinghal:
I think you can leave the icons as they are for now, until we can identify ones that are more representative.

rsinghal:
@aeschylus @sdellis What do you think of using some of the Google Material icons?  Perhaps: https://design.google.com/icons/#ic_gesture, https://design.google.com/icons/#ic_border_color, https://design.google.com/icons/#ic_format_color_fill?

stanislavkostadinov:
(drawing tools)
rectangle - https://design.google.com/icons/#ic_check_box_outline_blank
circle - https://design.google.com/icons/#ic_radio_button_unchecked
polygon - https://design.google.com/icons/#ic_timeline
pin - https://design.google.com/icons/#ic_room
freenahd - https://design.google.com/icons/#ic_gesture
(color pickers)
border color - https://design.google.com/icons/#ic_border_color
color fill - https://design.google.com/icons/#ic_format_color_fill

sdellis:
Apologies for just getting around to this.  It has been a busy week!  First of all, @stanislavkostadinov, this is wonderful and welcome work.  Thank you for your patience in this process.

@rsinghal, in order to circumvent my annotation server woes, I updated the ```index.html``` file configuration to match yours and made sure that the ```js/src/annotationscatchEndpoint.js``` module matched what is in master so I could use Catch.  I was able to get @stanislavkostadinov's PR to build using ```grunt --force``` as there is a warning in manifestListItem.  When I try to create an annotation I get the following error: ```Uncaught TypeError: Cannot read property 'removeChildren' of null```. The POST to the annotation server seems to have worked and returns a 200, however the annotations never visibly appear after they are created.  

As a sidenote, (and it might not solve this problem) I wonder if perhaps there should be a localStorage fallback for annotations that would support both annotation testing as well as offline annotation use cases as we get more into developing out Annotation features?  The whole annotation server thing seems like an unnecessary obstacle for front-end devs to have to deal with as a prereq to development.

At any rate, I can make a few comments on the behaviors of the tools and the choice of icons:

### Multi-region support
The tools seem to allow for multiple regions for a single annotation.  For example, I can draw a polygon and instead of edit the text in the annotation popup, I can create another polygon. Is this intentional?  From the discussion here it seems that multiple regions for a single annotation is supported (are there use cases for this project?).  However, I cannot zoom out or pan if I want to draw a region outside the viewport, which could be somewhat frustrating.

### Drawing Tool Conventions
I do not see most of these as show-stoppers, but the way shapes are drawn break common conventions used in popular graphical editors (Adobe products, Sketch, etc.) detracting from the experience

- As @lunactic mentioned, the initiation point for rectangles and circles should be a corner and not the center.  
- Click-drag is the convention used to scale shapes with mouseUp indicating the end of the scaling process.  Here, it is a click, mouseMove, and another click to end the process. This is may have been intentional to accommodate other functions for conflicting "drag-like" gestures on mobile, but it's not what I'm used to.
- Holding Shift while drawing should create a regular polygon (equiangular and equilateral) or perfect circle.
- Using the freedraw pencil and lightning bolt (should be a lasso IMO, although the Material "timeline" icon is probably fine) is hard because there is no "snap to point" feature that allows a user to easily close the shape.

### Trash, Save, and Rotate (or Refresh?)
The Trash, Save, and Rotate icons do not seem to work for me.  Perhaps if they are meant as tools for existing annotations? If that's the case they should be shown only on selection of an annotation.  Furthermore, they get mixed with the Full Screen icon at smaller screens, which made it hard to tell what they were related to (the window, image, all annotations, the annotation I have selected?).

### Default Tool
I wonder if clicking the Pencil editing tool present in the "View Annotations" mode should just default to a rectangle?  

### Material Icons
The Material icons Rashmi and Stanislove pointed out would definitely work better for some of those tools.  It would be nice to see how well they integrate with the current icons.

Hope this helps!

gigamorph:
Silly user question, but how do I close the polygon so I can see the edit window? Thanks.

stanislavkostadinov:
Drawing polygon ends with double click. If you use double click around (default distance is 5 pixels) the first point you will create closed shape.

yuricus:
There's a problem with json-ld generated by this component:

`{
  "@context": "http://iiif.io/api/presentation/2/context.json",
  "@type": "oa:Annotation",
  "motivation": [
    "oa:commenting"
  ],
  "resource": [
    {
      "@type": "dctypes:Text",
      "format": "text/html",
      "chars": "<p>a polygon</p>"
    }
  ],
  "on": {
    "@type": "oa:SpecificResource",
    "source": "http://oculus-dev.harvardx.harvard.edu/manifests/huam:165773/canvas/canvas-17810920.json",
    "selector": {
      "@type": "oa:SvgSelector",
      "value": "<svg xmlns='http://www.w3.org/2000/svg'><path xmlns=\"http://www.w3.org/2000/svg\" d=\"M387.24168,318.25369l-39.56547,0l-39.56547,0l0,-135.64158l0,-135.64158l39.56547,0l39.56547,0l0,135.64158l78.86089,423.73487l335.19886,-70.45581l115.29133,-226.31261l-441.9501,-104.61621l0,0z\" data-paper-data=\"{&quot;rotation&quot;:0,&quot;annotation&quot;:null}\" id=\"rectangle_e4a9effb-c884-44a7-94fc-812027eddfd1\" fill-opacity=\"0\" fill=\"#00bfff\" stroke=\"#00bfff\" stroke-width=\"1\" stroke-linecap=\"butt\" stroke-linejoin=\"miter\" stroke-miterlimit=\"10\" stroke-dasharray=\"\" stroke-dashoffset=\"0\" font-family=\"sans-serif\" font-weight=\"normal\" font-size=\"12\" text-anchor=\"start\" mix-blend-mode=\"normal\"/></svg>"
    },
    "scope": {
      "@context": "http://www.harvard.edu/catch/oa.json",
      "@type": "catch:Viewport",
      "value": "xywh=-1018,0,3830,2450"
    }
  }
}`

in '"source": "http://oculus-dev.harvardx.harvard.edu/manifests/huam:165773/canvas/canvas-17810920.json"' there's no "source" attribute in specified json-ld context http://iiif.io/api/presentation/2/context.json, you should use "full" instead to produce json-ld that conforms to Open Annotation model.

See examples on oa:SpecificResource at http://iiif.io/api/presentation/2.0/#non-rectangular-segments.

rsinghal:
Thanks @sdellis for your detailed response.  I like your suggestion of using the local storage as an "annotation" store for the purposes of demo and testing.  This came up separately at Harvard, and I am going to work to implement that because I think it will be very useful.

Being able to create multiple shapes for one annotation is a valid use case.  I can pan and zoom using the controls on the bottom left, but the mouse does not allow for those controls during "editing" mode.  Do the controls not work for you, @sdellis , either?  The only problem is the editor box might block the area of the image where you want to draw an image, so it may be useful to be able to move the editor box around during drawing.

Under "Drawing Tool Conventions" - I agree with your first three points, and hopefully they can be addressed.  I can't find a good lasso icon, so we will have to keep looking, but hopefully the timeline icon can work for now.  Per @stanislavkostadinov 's comment, it sounds like if you double click within 5 pixels of the first point, then it closes the shape.  I'm not sure what feature you are looking for, @sdellis .

The last icon (refresh/rotate), is for refreshing from the annotation store.  This is useful in our MOOCs when students might be annotating at the same time.  Rather than have Mirador constantly ping the db, we have the user's explicitly request the latest annotations.  I can see how it would be confusing, and I'm not sure how exactly to best represent this.  At the very minimum, we can have that be off by default, unless someone chooses to make that functionality available in their instance of Mirador.  The delete icon does work for me.

On being able to edit the shapes: this should not be triggered by clicking inside a shape, but when you click on a border.  That means the border will likely need to be a bit thicker to be more sensitive to user's clicks.  Right now, it's impossible to edit a shape that is fully surrounded by another shape.  Also, the current behavior is when you click inside a shape the first time and you move the mouse, it immediately starts moving the shape.  The initial click on the border should select the shape, but the user should have to click and move the mouse to move the shape.

Another issue I sometimes come across is if I create an annotation when I am zoomed in and reload the page and turn on annotations when I am fully zoomed out, the annotations are not in the same location as when I initially drew them.  I can't always create this bug, but if I can figure out the steps, I will post them.

I like the idea of having the Default Tool be rectangle so user can immediately start drawing an annotation.

I think, in general, it is best to follow the conventions in image-editing software, since that is what users are used to.  The work you have done is incredibly impressive, and now we just have to nail down some of these more specific things.

sdellis:
@rsinghal I hadn't thought of using the OSD controls for panning and zooming, but yes, those do work for me. I can see that the mouse controls are disabled to allow one to add multiple annotation regions.  This is probably fine and not a showstopper, but I will note that the convention in image-editing software is to hold down spacebar (cursor changes to hand) and drag.  There are some other keyboard commands for zooming (+/-) that may be helpful.

One other thing I noticed... the freehand "pencil" tool should use path simplification ([see example](http://paperjs.org/examples/path-simplification/)) available in Paper.js.  It currently produces a lot of coordinates that are going to unnecessarily bloat the annotations (and it's also a bit unsightly and unwieldy with so many points). 

Speaking of "drag", in the example above, notice how you have to click-and-drag to draw the shape.  That is what I am used to.  I am also concerned about tablets where touch-and-drag would be equivalent.  I will set up my MacBook to be able to share its local server on a network and test this on tablets.  My hunch is that it won't work in its current state.

I feel that "snap to point" to close shape rather than double click is more intuitive. This means that the cursor essentially "snaps" to a point when your cursor gets within x number of pixels depending on the granularity you want making it very easy to close shapes.  At the very least, the convention in image-editing tools is for the cursor to change (a little circle in the bottom right) when you are over a point or region that will close the shape.  Custom cursors are possible, but we'd have to come up with one.

(I wish I could help with the localStorage, but I'm in the middle of helping to organize Code4Lib... once it's over in mid-March I may be able to spend a bit of weekend time helping out with things like this that aren't in anyone's critical path, but will help developers get on-boarded.)

rsinghal:
@stanislavkostadinov I added tickets for individual issues that have been brought up in the comments to make it easier to have a conversation on each one.  If anyone else has anything else, please create an issue and reference this PR in the body so it will be referenced here

yuricus:
_Preface: @rsinghal introduced separate issues for the things being discussed here, this text was created just before that, we'll decide what to introduce as new issues or update existing ones soon._

We’re trying to adopt annotation tools at ResearchSpace and we’ve made these observations concerning user experience and integration possibilities. Some of them are repeating of what already had been mentioned, but we decided to present full list to stress those points.

For easy reference of current progress on this PR I’ve set up a live demo you can access at: http://mirador.apps.vismart.biz/

Here are the notes:

1. If you resize browser window, all annotation shapes are misplaced
2. It looks like annotation shape coordinates are taken against screen and not translated to coordinates of an image, this may cause (1), does anyone have strong opinion of what coordinates should be used in annotation?
3. Rectangle manipulation tool is not what users expect, combined zoom-rotate is confusing. In my mind, user should have separate tools for editing a rectangle (moving a corner or a side, most used one) and rotating a region (less used).
4. It is not trivial how to edit the body of the annotation - you have to disable annotation tools, hover over annotation and push ‘edit’. Is there any other way?
5. The sequence of selecting tools/manipulating annotations is hard to carry out. One idea could be to follow the example of image editing software, e.g. Photoshop, Gimp, Corel Draw, etc. 
6. Preferable behaviour, in my mind, should be:
 * Introduce a  select tool: when the user selects an annotation with a select tool, options for modifying selected annotations become available: shape edit, color change, annotation text change, deletion of annotation.
 * The user should be able to edit selected annotation without switching to particular shape tool, you already have selected shape as a context.
 * Save behaviour: when user is editing shapes, they are modified, but not saved. When user clicks ‘save’ button all modifications are sent to annotation storage.
 * When ‘edit mode’ of annotations is not activated, user should not be able to modify annotations at all.
7. Instead, current behaviour is such
  * Without entering edit mode, user can change a body of annotation or delete it. (Is this the only way to do it? I’m not 100% positive here)
  * To modify a shape of an annotation, you have to select the tool according to annotation shape, click on annotation shape, the software enters into edit mode and you can modify it. If you select another tool by mistake, new annotation will be drawn over the previous one, confusing the user.
8. When integrating this set of annotation tools to another platform, there is a number of considerations.
  * We would like to use our own annotation editor, so it would be desirable, if the editor of the annotation body is plugable.
  * TinyMCE is complex with it’s own themes, plugins, etc and it has it’s own resource loader, which is hard to use with webpack, more simple choice by default could be an option.
  * We’d like to provide our own list of annotations to the image and to be able to talk to mirador when annotation is selected/deselected in our list and visa versa. Right now there’s no ‘select’ tool and no ‘selected’ state and it requires quite a lot of refactoring. In general, a clean states and events would make it much easier to extend and customize image annotation.

We would like to help on those issues and I suggest to have a talk on the next development standup how to address them.

rsinghal:
Merging this into the branch on the Mirador repo.  Anyone who wants to submit fixes to this code, should submit PRs to the annotation-work branch.  Eventually, this will be merged with the refactor branch.

PR #752, opened 2016-02-10
Fixed jshint error: variable "image" was being used "out of scope".

gigamorph:
Probably not fatal but bad style.


gigamorph:
Let me know if I should rather create an issue than a pull request for this kind of small fixes. Thanks.

rsinghal:
In the future, it would be best to make a ticket first and then submit the PR.  Also, if your PR also makes significant changes to the white space, then it would be best to separate that out from code changes so we can more quickly review the code changes.

PR #756, opened 2016-02-19
Failsafe to get currentConfig anyway if jsonBlob call fails

yuricus:
This patch will prevent currentConfig in saveController from not being initialized when jsonBlob call fails.


rsinghal:
Thanks @yuricus

PR #773, opened 2016-03-01
Bug-fix of resizing annotations

stanislavkostadinov:
Annotations change position and size when image is resized or when the OSD is in fullscreen mode


rsinghal:
This looks good to me, so I am going to merge.  @yuricus please feel free to test from annotation-work since you noticed this issue as well.

Issue #776, opened 2016-03-01
How to finish a polygon shape?


rsinghal:
@stanislavkostadinov 

It's not clear to me how to finish drawing a polygon so that I can get the text box to appear.  If I double click I end up with an extra circular shape.

stanislavkostadinov:
May be the extra circle is the one that is used as a clue when user closes the polygon. I've made an animated gif of how it should be working (test env.: Windows 7 OS, browsers - Firefox 44.0.2, Chrome 48.0.2564, IE 11.0.9600).
![polygon](https://cloud.githubusercontent.com/assets/4114894/13454989/63e4c846-e063-11e5-9dc6-2615969f3708.gif)

PR #778, opened 2016-03-02
Bug-fix of zoom (images with width/height bigger than 2^13)

stanislavkostadinov:
When image is zoomed the annotation figures are not visible.


stanislavkostadinov:
http://stackoverflow.com/questions/6081483/maximum-size-of-a-canvas-element

stanislavkostadinov:
Latest commit actually fixes #776

stanislavkostadinov:
- One trivial big-fix of circle tool: when resizing circle with height 1px (or less) the image flashes.
cicle_tool

PR #780, opened 2016-03-03
Ability to provide manifest as json content instead of URI

yuricus:
When embedding Mirador as JS component to a page, sometimes it makes sense to provide Manifest object to the Mirador directly, rather then force it to be loaded or constructed from IIIF Image API's info.json 

I know there was some work done on providing generalized way to deal with manifests, but it looks like it was unfinished.

We're using this functionality in ResearchSpace, I'm not 100% sure it's for everyone and this code should go to this branch only if it's useful. Glad to join discussion on the subject.


rsinghal:
@yuricus Can you give an example of how this would be configured upon initializing Mirador?

yuricus:
We're preparing manifest on the client side using jsonld.js transformations like framing and passing manifest to Mirador directly, rather than providing URL to download manifest from back end.

Like, having ready manifest in `manifest`, and it's URI in info.imageIRI.value we initalize Mirador the following way:

```
Mirador({
          "id": "mirador", // The CSS ID selector for the containing element.
          "layout": "1x1", // The number and arrangement of windows ("row"x"column")?
          "saveSession": false,
          "data": [
            {"manifestUri": info.imageIRI.value, "location": "British Museum", "manifestContent": manifest}
          ],....
});
```

Right now it's only one-time initialization of Mirador, no way of dynamically alter running instance, e. g. providing new manifest do display.

beaudet:
I am interested in this as well and there might be an existing Mirador ticket for this.  ConservationSpace was originally thinking along these lines as well but we ended up opting for a URL based load for several reasons:


1. there didn't seem to be any existing Mirador APIs for injecting the manifest directly so one would have to be written

2. there seemed to be a dependency on the "Manifesto-like" refactoring of Mirador to complete #1 which hasn't started yet and is therefore incompatible with our delivery schedule

3. and perhaps most importantly, we were originally thinking that we would use Mirador to modify the manifests to generate "views" for inclusion in the conservation documents and then save the modified manifests in ConservationSpace but after discussions with the Mirador developers, we decided against that approach.  Since ConservationSpace is seeking to create a long term archive of conservation documents, relying on support for older versions of the IIIF presentation API represents a risk against future compatibility.


Dave Beaudet, NGA

________________________________

rsinghal:
@yuricus Can you redo this PR and submit it [refactor](https://github.com/IIIF/mirador/tree/refactor) instead?  I have merged all of the other annotation-work code into refactor and would like to delete annotation-work, especially since this PR isn't related to annotations.

yuricus:
Annotation-work was merged to refactor merged to master, I will create PR to master instead.

PR #801, opened 2016-03-23
Added ability to provide manifest as json content instead of URI

yuricus:
Supersedes #780 

If user prepares manifest json object on the client side, he now can pass manifest to Mirador directly, rather than providing URL to download manifest from back end.

Like, having ready manifest in `manifest`, and it's URI in manifestURI you could initialize Mirador the following way:

```
Mirador({
          "id": "mirador", // The CSS ID selector for the containing element.
          "layout": "1x1", // The number and arrangement of windows ("row"x"column")?
          "saveSession": false,
          "data": [
            {"manifestUri": manifestURI, "location": "British Museum", "manifestContent": manifest}
          ],....
});
```

Right now it's only one-time initialization of Mirador, no way of dynamically alter running instance, e. g. providing new manifest do display.

### DIFF
--- a/js/src/manifests/manifest.js
+++ b/js/src/manifests/manifest.js
@@ -1,7 +1,15 @@
 (function($){
 
-  $.Manifest = function(manifestUri, location) {
-    if (manifestUri.indexOf('info.json') !== -1) {
+  $.Manifest = function(manifestUri, location, manifestContent) {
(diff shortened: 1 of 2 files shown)


rsinghal:
@yuricus This is great.  Thanks for changing the PR

PR #802, opened 2016-03-23
Make annotation body editor pluggable, existing TinyMCE being default choice

yuricus:
# Intro

This change brings pluggable annotation body editor into Mirador, existing TinyMCE remains default implementation.

# Motivation

1. Different use-cases require different data to be associated with an annotation, from rich text, tags to other references etc.
2. When integrating Mirador with existing complex front-end systems developers could reuse some editors that's already exist in system.
3. TinyMCE is complex component with it's own themes, plugins, dynamic resource loading, it's hard to package into build system like webpack, rich text editing may not be not enough reason to deal with the complexities and it's easier to use something less advanced.

# Design
* AnnotationTooltip - manages annotation tooltip lifecycle, controls, container for image editor
* OsdSvgOverlay - service for managing annotations
* TinyMceAnnotationEditor - component for viewing/editing annotation.

## AnnotationTooltip
Annotation tooltip handles annotation dialog - creating, handling events, showing edit/save/cancel buttons, etc.

## EditorComponent
Editor is configured on Mirador init with editor plugin name and parameters:
```
'annotationBodyEditor': {
  'module': 'TinyMCEAnnotationBodyEditor',
  'options': {}
},
```

Editor itself must have following members:
show
isDirty
createAnnotation
updateAnnotation
For more information, parameters and docs consult TinyMceAnnotationEditor.

# Further development ideas
At this point minimal steps were taken to make editor pluggable. A lot of jQuery events are fired along the way, priority was not to change them, just rearrange the code. Still, something is worth improving:

* Cleaner interfaces (no upgrade to edit, etc.) between layers.
* Make annotation viewer part of editor to be customizable
* Abstract annotation tooltip, make it pluggable to make different UIs for annotations possible
* Revise/enrich event/callback system to make things like zoom to given annotation possible

### DIFF
--- a/js/src/annotations/annotationTooltip.js
+++ b/js/src/annotations/annotationTooltip.js
@@ -3,9 +3,10 @@
   $.AnnotationTooltip = function(options) {
 
     jQuery.extend(this, {
-      element:   null,
+      targetElement: null,
       annotations: [],
-      windowId:  ""
+      parent: null,
+      windowId: "",
     }, options);
 
     this.init();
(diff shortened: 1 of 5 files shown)


stanislavkostadinov:
I've implemented similar functionality by using events. The default behaviour remains the same as the current one. Default comment editor and comment tooltips can be replaced with only few lines. I'm useing these events:

'windowSlotAdded' - slotAddress ->  provides slot info (id for all the other events)
'annotationList.' + id (new one) -> provides all the comment related to this id
'annotationMousePosition.' + id (new one) -> notifies external application to show/hide 1 or many annotations data
'annotationShapeCreated.' + id (new one) when new shape is created and external application should open popup
'annotationCreated.'+id, external applications publish that event when new comment is created
'annotationUpdated.'+id, external applications publish that event when comment is undated
'annotationDeleted.'+id, external applications publish that event when comment is undated
'annotationListLoaded.' + id and external applications should subscribe to this event if they want to be notified when create,edit,delete actions are done

yuricus:
@stanislavkostadinov yes, introducing events may be an option to completely replace existing annotation environment with anything developer wants. 

Yet sometimes changing only text editor would suffice, it's much easier than writing event handling code and doing all update-create-edit logic by hand.

I think both solutions complement each other and could live inside Mirador side-by-side, with I suggest the way to follow in our case.

In fact, I think we would use event system around annotations in near future.

rsinghal:
@yuricus I haven't tested this yet, but there are few things to point out before we can continue:  we are moving away from accessing data in Mirador using `$` and `parent`.  So, instead of passing `parent: this` to child modules, we are passing `this.state` from parent to child and accessing properties with something like this `this.state.getStateProperty('windowObjects')`

Therefore, lines like 18 and 19 in `AnnotationTooltip` should be changed to access the `state` in this new way.

yuricus:
Updated to match refactor config

rsinghal:
@yuricus Sorry, I was renaming `refactor` to `release2.1` and didn't realize that would close this PR.  Can you please redo?

PR #804, opened 2016-03-24
New border type tool, resize bug-fix & functionality used to plug external comment popups

stanislavkostadinov:
New border type tool: solid line, dashes, dots and dashes


rsinghal:
@stanislavkostadinov  Sorry, I was renaming `refactor` to `release2.1` and didn't realize that would close this PR.  Can you please redo?

Issue #653, opened 2015-09-29
Mirador thumbnail label displays [object object] for multilingual canvas labels


azaroth42:
For example:

```
{"@type": "sc:Canvas",
 "label": [{"@value": "English", "@language":"en"}, {"@value": "Francais", "@language":"fr"}]
 ...
}
```

rsinghal:
@azaroth42 Do you have a fixture object we can test against?

azaroth42:
Via:
    https://github.com/UniversalViewer/universalviewer/issues/188#issuecomment-176296418

A manifest in the wild that demonstrates this:
    https://iiif.riksarkivet.se/arkis!C0000263/manifest

sdellis:
Just curious, how does the client know which language label to display by default when multiple are provided as in this instance?  (I'm not talking about the metadata panel here.)

tomcrane:
Hi @sdellis - the UV supports a list of locales in preferred order as part of its configuration - which can be used to pull the appropriate metadata from the manifest as well as control the language of the user interface. The user can override the locale. There's also a "toggle" mode in use at the National Library of Wales - so that even if many locales are available to choose from in "settings", the UV draws attention to a toggle between two of them, in this case English and Welsh.

Another option is to use the browser locale, but the UV doesn't do this.

sdellis:
@tomcrane yes, I had a conversation with Ed about this recently, and I like how UV handles it as a config setting/user preference.  I'm sorry, I typed "client" but meant "Mirador" as it was a Mirador issue.  My bad.

yuricus:
We've got intermediate result for fixing this: #806 
This pull request fixes [object object] display to more sensible one (prefers English labels).
To encapsulate language handling new function is created and it's called from different parts of Mirador.

Our small effort creates foundation of working with multilingual labels, I hope that some one will pick it up from here.

rsinghal:
Also related: the label on the manifest view page

Issue #798, opened 2016-03-18
Incorrect metadata value displayed


mikeapp:
The metadata display repeats the value for a previous label (accession number appears again under bibliography).  The screen shot shows the current version of Mirador vs. an earlier version.  The demo Manifest is http://manifests.britishart.yale.edu/manifest/5009

<img width="776" alt="screen shot 2016-03-18 at 3 25 46 pm" src="https://cloud.githubusercontent.com/assets/1987952/13889717/6642bf36-ed1e-11e5-9e65-4512de45bea6.png">

rsinghal:
We are making the assumption that if the value is an array, then it is an array of objects that provide the same value of that label in multiple languages.  That manifest has an array for `Bibliography` with one string in it.

Before I make the correct fix, this is probably also a question for @azaroth42 since I can't find anything specific one way or the other in the api description.  Is this structure in the manifest valid?

azaroth42:
An array of values with no language associated (e.g. where the values are not objects with @value/@language keys) then it's just a set of values.

The structure in the manifest:

```json
{
 "metadata": [
    {
      "label": "Bibliography",
      "value": ["...", "...", "..."]
    }
  ]
}
```

Is correct.  This is just a case of http://iiif.io/api/presentation/2.0/#repeated-properties, where the property being repeated is `value`.

The content of those strings is not correct, as according to http://iiif.io/api/presentation/2.0/#property-values-in-html if the value is HTML, then it MUST start with `<` and end with `>` but the client should be displaying all of them.

Yes ... the test here is a bit wishy washy to distinguish between a choice based on language and only one should be displayed, and a list where all of the values should be displayed.  The alternative is to always distinguish a choice as a separate resource, which makes it harder to have the strings for the 95% case, and the decision was to be slightly ambiguous.  We could clarify this in http://iiif.io/api/presentation/2.1/#language-of-property-values

beaudet:
Seems reasonable and not questioning the existing spec but why would text with HTML need to start with open bracket and end with close bracket?  What about cases where a string contains simple embedded formatting codes for italics, underline, bold, etc.? Should we start with empty span tag and end with same?

-------- Original message --------

azaroth42:
The requirement to start and end with brackets is to make it clear (or as clear as possible) to distinguish HTML from plain text that might include angle brackets. So the resolution to whether a `< `is a real `<` or the beginning of an HTML tag is the test on the first and last characters of the string.

mikeapp:
[...]

PR #806, opened 2016-03-28
Change multi-lingual thumbnail label display to sensible defaults

yuricus:
As mentioned in #653 [Object object] is displayed where thumbnail label should be when labels are multi-lingual.

As discussion and ongoing work progresses in original issue, this is intermediate result addresses two things:

1) JsonLd.getTextValue(property, language) is introduced to fetch text from a property. Defaults to English. If label in specified language is not found, function tries to get label in any language (non-deterministic).
2) Two calls to this function are introduced in metadataView and window to uniformly cope with multilingual label.

We laid foundations to solving the original issue, further work will be introducing language selector and propagating language property to JsonLd.getTextValue call, but we will not be able to implement that in near future.

### DIFF
--- a/js/src/widgets/metadataView.js
+++ b/js/src/widgets/metadataView.js
@@ -126,8 +126,8 @@
   getMetadataDetails: function(jsonLd) {
       // TODO: This should not default to English
       var mdList = {
-          'label':        '<b>' + jsonLd.label + '</b>' || '',
-          'description':  jsonLd.description || ''
+        'label': '<b>' + ($.JsonLd.getTextValue(jsonLd.label) || '') + '</b>',
+        'description':  jsonLd.description || ''
       };
       var value = "";
       var label = "";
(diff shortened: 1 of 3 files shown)


js/src/widgets/metadataView.js
```diff
@@ -126,8 +126,8 @@
   getMetadataDetails: function(jsonLd) {
       // TODO: This should not default to English
       var mdList = {
-          'label':        '<b>' + jsonLd.label + '</b>' || '',
-          'description':  jsonLd.description || ''
```
rsinghal: I think `description` can also be treated the same way as label, and therefore it is valid to call `getTextValue` on it as well, per: http://iiif.io/api/presentation/2.0/#language-of-property-values.

rsinghal:
@yuricus Can you change this PR to master?  This is related to #798, and I would like to incorporate your work into that bug fix.  I will then merge it into refactor after.

rsinghal:
@yuricus Sorry, I was renaming `refactor` to `release2.1` and didn't realize that would close this PR!

PR #809, opened 2016-03-30
Adaption of the thumbnail

fe-kre:
Problem was a thumbnail that would be bigger than a usual screen.

I set up fixed values as limits. If the aspect ratio exceeds one limit, than calculate the width or height again with a correction variable. After that point, recalculate the aspect ratio and the new width.

Because of quality reasons for the thumbnail, I set the width to a fixed value.


rsinghal:
Thanks @fe-kre.  Can you include an example manifest that caused the problem originally?

fe-kre:
http://iiif.ub.uni-leipzig.de/0000000001/manifest.json is an example manifest for our problem.

rsinghal:
@fe-kre The code changes are messing up the thumbnails for existing images.

Before:
screen shot 2016-04-04 at 11 51 14 am

After:
screen shot 2016-04-04 at 11 51 27 am

rsinghal:
@fe-kre Please fix and reopen the PR when you are ready.

fe-kre:
We have taken a new approach to our problem and have resolved it with the help of css properties.

I set up a new div around 'preview-images' and gave him the properties position: absolute and display: inline-flex. So our outsized thumbnail shows up.

Hopefully this is an acceptable way to deal with our problem.

Issue #609, opened 2015-08-06
Make Annotation editors pluggable


rsinghal:
So developers can choose which editor to use (and to remove dependency on TinyMCE)

PR #814, opened 2016-03-30
Make annotation body editor pluggable, existing TinyMCE being default choice

yuricus:
# Intro

This change brings pluggable annotation body editor into Mirador, existing TinyMCE remains default implementation.

# Motivation

1. Different use-cases require different data to be associated with an annotation, from rich text, tags to other references etc.
2. When integrating Mirador with existing complex front-end systems developers could reuse some editors that's already exist in system.
3. TinyMCE is complex component with it's own themes, plugins, dynamic resource loading, it's hard to package into build system like webpack, rich text editing may not be not enough reason to deal with the complexities and it's easier to use something less advanced.

# Design
* AnnotationTooltip - manages annotation tooltip lifecycle, controls, container for image editor
* OsdSvgOverlay - service for managing annotations
* TinyMceAnnotationEditor - component for viewing/editing annotation.

## AnnotationTooltip
Annotation tooltip handles annotation dialog - creating, handling events, showing edit/save/cancel buttons, etc.

## EditorComponent
Editor is configured on Mirador init with editor plugin name and parameters:
```
'annotationBodyEditor': {
  'module': 'TinyMCEAnnotationBodyEditor',
  'options': {}
},
```

Editor itself must have following members:
show
isDirty
createAnnotation
updateAnnotation
For more information, parameters and docs consult TinyMceAnnotationEditor.

# Further development ideas
At this point minimal steps were taken to make editor pluggable. A lot of jQuery events are fired along the way, priority was not to change them, just rearrange the code. Still, something is worth improving:

* Cleaner interfaces (no upgrade to edit, etc.) between layers.
* Make annotation viewer part of editor to be customizable
* Abstract annotation tooltip, make it pluggable to make different UIs for annotations possible
* Revise/enrich event/callback system to make things like zoom to given annotation possible

@stanislavkostadinov made event-based handling of annotation body editor to replace annotation body editor interface completely #808 . Both solutions can be integrated and complement each other.

### DIFF
--- a/js/src/annotations/annotationTooltip.js
+++ b/js/src/annotations/annotationTooltip.js
@@ -3,9 +3,10 @@
   $.AnnotationTooltip = function(options) {
 
     jQuery.extend(this, {
-      element:   null,
+      targetElement: null,
       annotations: [],
-      windowId:  ""
+      parent: null,
+      windowId: "",
     }, options);
 
     this.init();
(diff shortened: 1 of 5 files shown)


yuricus:
Found existing issue for that #609

Issue #789, opened 2016-03-07
Collapse all canvas controls (except basic navigation) into a single hierarchical menu.


aeschylus:
<img width="936" alt="screen shot 2016-03-07 at 10 35 43" src="https://cloud.githubusercontent.com/assets/145874/13579137/9b77eefa-e450-11e5-9afa-dcff62e1562a.png">
<img width="924" alt="screen shot 2016-03-07 at 10 35 54" src="https://cloud.githubusercontent.com/assets/145874/13579139/9d7ab8b8-e450-11e5-856e-66e3d9b0cadd.png">

aeschylus:
We should probably discuss what this means for configuration, since many people will want very specific sets of these tools in very specific orders to be displayed on the canvas.

stanislavkostadinov:
The order of the tools depends on the order of the `availableAnnotationDrawingTools`.

PR #816, opened 2016-03-31
Allow defining metadata for a canvas.

Wedjaa:
## Canvas Specific Metadata

This PR implements a way of defining metadata that's specific to the canvas that is currently in focus. We implemented this as a way of allowing a collection of heterogeneous  group of images in a collection - that we create on the fly based on the user search results in our specific case - to be described individually rather than collectively.
## Using the manifest to point to canvas metadata

In the manifest - in the section with the canvas where you want to add some metadata - you can add an annotation of type `oa:describe` that will point to the service providing the information. In other words - since an example is worth a thousand words:

```
....
....
          "otherContent":[  
            {  
              "@context":"http://iiif.io/api/presentation/2/context.json",
              "@id":"http://dfc.frick.org/api/search/info/collid/canvasid",
              "@type":"sc:AnnotationList",
              "resources":[  
                {  
                  "@type":"oa:Annotation",
                  "motivation":"oa:describing",
                  "resource":{  
                    "@id":"http://dfc.frick.orf/api/search/info/collid/canvasid",
                    "@type":"dctypes:Dataset",
                    "format":"application/json"
                  },
                  "on":"http://dfc.frick.org/compare/collid/sequence/normal/canvas/canvasid"
                }
              ]
           }
        ]
....
....
```

  In plain English we are telling Mirador to connect to `http://dfc.frick.orf/api/search/info/collid/canvasid` to retrieve information regarding this particular canvas - specified on the _on_ property.
## Metadata format

The service endpoint should return the metadata in a form that can be used by Mirador to extract and display it - something along these lines:

```
{  
  "resources":[  
    {  
      "@context":"http://iiif.io/api/presentation/2/context.json",
      "@id":"http://dfc.frick.org/compare/collid/sequence/normal/canvas/canvasid",
      "@type":"oa:Annotation",
      "on":"http://dfc.frick.org/compare/collid/sequence/normal/canvas/canvasid",
      "motivation":[  
        "oa:describing"
      ],
      "resource":{  
        "FileName":"image_001.jpg",
        "Creator_culture":"Italian School. ",
        "Creator":"Palma, il Giovane, 1544-1628.",
        "WorkTitle":"St. Cyriacus"
      }
    }
  ]
}
```

The information under _resources_ will be displayed in the metadata panel under _Media Details_.


js/src/annotations/osd-region-draw-tool.js
```diff
@@ -115,6 +115,18 @@
       this.svgOverlay.restoreEditedShapes();
     },
 
+    isImageAnnotation: function(annotation) {
```
Wedjaa: This function will check if the annotation that was passes is an image annotation - one to be displayed on the canvas - or a metadata annotation. The logic is that if we have an annotation with motivation `oa:describing` we don't want to add it to the list of annotations that will be displayed.

js/src/annotations/osd-region-draw-tool.js
```diff
@@ -122,25 +134,31 @@
...
       _this.annotationsToShapesMap = {};
       var deferreds = jQuery.map(this.list, function(annotation) {
-        var deferred = jQuery.Deferred();
-        var shapeArray;
-        if (typeof annotation.on === 'object') {
-          if (annotation.on.selector.value.indexOf('<svg') != -1) {
-            shapeArray = _this.svgOverlay.parseSVG(annotation.on.selector.value, annotation);
+
+        var deferred;
+
```
Wedjaa: We create the deferred only if this is not a metadata annotation. See function above.

js/src/widgets/metadataView.js
```diff
@@ -20,11 +20,18 @@
           tplData = {
             metadataListingCls: this.metadataListingCls
           };
-          
-      _this.manifest = _this.manifest.jsonLd;
+
```
Wedjaa: We don't want to change the manifest if it's already defined.

js/src/widgets/metadataView.js
```diff
@@ -208,30 +219,77 @@
...
-        if (stateValue) { 
-            this.show(); 
+        if (stateValue) {
+            this.show();
         } else {
             this.hide();
         }
     },
-    
+
```
Wedjaa: This is where most of the logic is: find metadata annotations for the canvasid we have now and use it to populate the metadata.

js/src/widgets/metadataView.js
```diff
@@ -208,30 +219,77 @@
   },
 
     bindEvents: function() {
```
Wedjaa: Subscribe to detect changes in the annotations list, IE: the focused canvas may have changed.

js/src/widgets/metadataView.js
```diff
@@ -252,9 +310,17 @@
 
       return textWithLinks;
     },
-    
+
     template: Handlebars.compile([
-    '<div class="sub-title">{{t "details"}}:</div>',
```
Wedjaa: We add a section for the canvas specific metadata.

js/src/workspaces/window.js
```diff
@@ -755,7 +755,10 @@
               value['@id'] = $.genUUID();
             }
             //indicate this is a manifest annotation - which affects the UI
-            value.endpoint = "manifest";
+            if (typeof value.endpoint ===  'undefined') {
```
Wedjaa: Change the annotation endpoint to 'manifest' only if it's not defined already.

locales/it/translation.json
```diff
@@ -0,0 +1,37 @@
+{
```
Wedjaa: I'm Italian - I felt compelled to add this....

rsinghal:
@Wedjaa Why did you choose to make the canvas-level metadata an annotation instead of using the `metadata` field on a canvas?

We are working towards making the different panels more consistent.  Therefore, the bar with the manifest title and other navigation, should be all manifest-level information (including the `i` button, which displays only metadata on the manifest level).  Canvas-level metadata should be displayed on the canvas level in Mirador, and we will be following the pattern in #789, where the canvas menu will have a button for the canvas-level metadata.

iptet:
Well - we are taking advantage of the service endpoints of annotations to allow on one side to keep a manifest light and stable, while being able to update the metadata for the canvas. Think about it like this: in the manifest I want information that the viewer can use to go and get the 'final' information about some media, extended -- and possibly variable - information should be requested when displaying.

  More to the point: in our use-case we are generating the manifests on the fly, depending on the results of a search and media that was picked by the user -- I submitted a talk proposal on this, but haven't heard back yet. The user can bookmark and share the results of the search and the manifest itself.

  If some information about the media changes - maybe we added a field, maybe some scholar found out that the date of the item is wrong - we don't need to change anything in the manifests people have saved and bookmarked, since the information is picked up dynamically. If we stored the canvas metadata in the manifest this would not be possible, and people that have saved a manifest would not see the updates.

  At an higher level: I think that the metadata about a canvas can be either volatile or stable, and ideally you would include stable metadata in the manifest and volatile information as annotations of a specific type -- oa:describe in our implementation.

 About where to display the metadata: if displaying a well defined collection - say a book, or a collection of homogeneous items - it makes sense to have the (i) button display information on the collection, I agree. But when handling and heterogeneous group of media the (i) button would make more sense to show information about the specific canvas that's being displayed. Showing both - currently one after the other, maybe later simply in two different tabs in the information panel - does make sense. I have a phobia of button proliferation - so a simple button showing the relevant information about a canvas and a collection is a soothing interaction for me, having two buttons on the same interface that do two slightly different things is a bit less soothing. But that's just me! :)

rsinghal:
There is also the [service](http://iiif.io/api/presentation/2.0/#linking-properties) property.  I'm not sure if this use case would be the correct usage of that.  Perhaps @azaroth42 can give more insight.

We are moving towards a true separation of concerns within Mirador, and all of the information in that navigation bar has to do with the manifest, and everything that has to do with the canvas has to be on the actual canvas.  The same thing will happen with Ranges, when we add functionality to support multiple ranges.  Metadata about each range will only be viewable in the Range area. @aeschylus

azaroth42:
In terms of the Presentation API, metadata about canvases should be given with the `metadata` label/value pairs in the manifest.

Pulling in remote content would be a good use case for `service`, there's also `seeAlso` which seems a closer (more semantically accurate) fit.

iptet:
@azaroth42 I was using the 'oa:describing' just for that reason: it tied up very well with the existing code, allowing to more or less make the adding of more information describing the item seamless and keeping with the semantics of the motivation choses:  'The motivation that represents a description of the target resource(s), as opposed to a comment about them...'.

@rsinghal Is there a branch where a display for the information about the canvas is already in the works or would I need to start from scratch and try and implement the ideas from #789 ?

For the moment we are going into production with this solution at the Frick, I'll try and see what I can do later to comply with the new direction of Mirador and integrate it better in the layout. I must say that this was initially implemented because most of our users expected to see information about the image they were looking at when clicking on the (i).

azaroth42:
Yes, there's nothing wrong with doing that :) It's just not in the scope of the IIIF side of things.  Would be very happy to hear your thoughts on the new W3C specs: https://www.w3.org/TR/annotation-model/
(though in a different venue, rather than polluting the issue here)

rsinghal:
@Wedjaa Can we work to add this into release2.1?  I want to make sure that we can also get data from the `metadata` field on a canvas, if it exists.  I am also thinking it would only be helpful to show canvas level metadata when a user is on `ImageView` or `BookView`, because it might be confusing to see individual metadata on `ThumbnailsView`.  If it is easier, I can create a new branch for this, and you can PR your stuff to that, and I can work on some of these other features.

Wedjaa:
I will try and make the PR more 'presentable' - let me know when the
branch is ready so that I can start the PR,

  Fabio


2016-05-31 9:41 GMT-05:00 Rashmi Singhal <notifications@github.com>:

rsinghal:
Done.  `canvas-level-metadata` now exists and is even with `release2.1`

rsinghal:
@Wedjaa bump

Wedjaa:
@rsinghal Sorry - I am currently off the project -- will try and commit the PR as soon as I can.

rsinghal:
@Wedjaa no problem - even if there are conflicts, can you recreate the PR to submit it to `canvas-level-metadata`?  I can work to resolve those conflicts.  Thanks!

aeschylus:
Closing this for now. It is an interesting discussion but can still be kept around in the closed PRs section.

Also, I think the "metadata for all levels of object" question has shifted more toward a sidebar and manifesto+metadata-component in 2.2. Feel free to re-open if there is any strong disagreement.

Issue #305, opened 2014-10-22
Support for virtual ruler


snydman:
parity with 1.0rc

markmatney:
The organization that is funding UCLA Library's Mirador project ("Sinai Palimpsests") are interested in this feature. Are the "help wanted" label and "2.1 milestone" designation accurate (the [Mirador Plan 2016](https://docs.google.com/document/d/1bVwrOHiM7EvVxl4Ogy7stIH0-R_GBXF7vuAJvYxPhG0/edit#heading=h.6csfyv1s2tkj) does not list the feature as a release priority), and can anyone speak on the possibility of the feature being added to an upcoming release?

markmatney:
I've implemented a prototype (using [this OpenSeadragon plugin](https://github.com/NIST-ISG/OpenSeadragonScalebar) as the base) to fit the use cases for UCLA LIbrary's Mirador project. You can see a demo of it [here](https://drive.google.com/file/d/0B_xDW0hVa9VzS3ZIZEhqdmVnSVE/view?usp=sharing). The numbers on the ruler are useless right now, since I haven't configured it to grab the physical dimension data yet.

In the interest of future development, I also created a [design document](https://docs.google.com/document/d/19QQPV4DTMCa4BZohrmTEyIbkM2EkiXcWqo79ePcozQA/edit?usp=sharing) that describes a UI for the ruler that _I'd_ like to use. Please feel free to share, comment, or edit. If there's time at an upcoming community call, I'd like to discuss this doc.

azaroth42:
:+1: ! 

The physical dimension data should be provided in the manifest as per:  
  http://iiif.io/api/annex/services/#physical-dimensions

markmatney:
Got it. I am working on adding that functionality, as well as reorganizing my code.

aeschylus:
Very cool. I can see this being merged in soon from a PR if we can get the new menu in. I left a few comments in the design doc. See: https://docs.google.com/document/d/19QQPV4DTMCa4BZohrmTEyIbkM2EkiXcWqo79ePcozQA/edit#heading=h.4c7gfqbeegc

markmatney:
Update: I have reorganized my code into a "plugin" (https://github.com/UCLALibrary/mirador-ruler), making it one of two ruler plugins (that I know of) (https://github.com/dbmdz/mirador-plugins/tree/master/PhysicalRuler)

camillevilla:
[...]

Issue #97, opened 2014-07-07
[2.0] Physical size ruler support


azaroth42:
Process the physical size service from either the Image API info.json response or the Presentation API in order to rescale the ruler to the appropriate size.
See: http://iiif.io/api/annex/services/#physical-dimensions

markmatney:
Can we close this issue and direct discussion to https://github.com/ProjectMirador/mirador/issues/305? The two issues seem to be duplicates, and 305 has more comments.

PR #820, opened 2016-04-05
Physical Ruler (Issues #305 and #97)

markmatney:
This commit enables support for physical ruler overlay.

As far as IIIF spec compliance goes, right now it only supports the single-image-per-canvas case, where the [physical dimension service](http://iiif.io/api/annex/services/#physical-dimensions) is provided as a top-level key-value pair in the `info.json` for that image.

Also, only millimeter rulers are currently supported. So, if a physical dimension service is provided with units with another unit type (e.g., "in" or "cm"), the correct physical dimensions will still be reflected by the ruler, but in millimeters.

If no physical dimension data is available for an image, the UI buttons that control the ruler simply don't do anything when pressed (although they are still visible to the user).

Settings are found in `js/src/settings.js`, under the `scalebar` key.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -31,6 +31,7 @@ module.exports = function(grunt) {
     'js/lib/tinymce.min.js',
     'js/lib/handlebars.js',
     'js/lib/openseadragon.min.js',
+    'js/lib/openseadragon-scalebar.js',
     'js/lib/d3.v3.min.js',
     'js/lib/pubsub.min.js',
     'js/lib/URI.min.js',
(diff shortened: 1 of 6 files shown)


beaudet:
Maybe rather than hiding the button or having a non-functional one the ruler could display pixels instead of real dimensions in cases where no real dimensions are provided?

-------- Original message --------

markmatney:
I can see why that would be valuable (relative measurements are better than no measurements).

aeschylus:
Just leaving some notes for later:

In order to merge this we need to resolve: https://github.com/IIIF/mirador/issues/789, and merge IIIFML, since there is some coupling between osd and the ruler.

PR #828, opened 2016-04-08
New border type tool: solid line, dashes, dots and dashes + bug-fixes

stanislavkostadinov:
Bug-fix: https://github.com/IIIF/mirador/issues/764

### DIFF
--- a/js/src/viewer.js
+++ b/js/src/viewer.js
@@ -93,7 +93,10 @@
       });
 
       this.manifestsPanel = new $.ManifestsPanel({ appendTo: this.element.find('.mirador-viewer'), state: this.state });
-      this.bookmarkPanel = new $.BookmarkPanel({ appendTo: this.element.find('.mirador-viewer'), state: this.state });
+      //only instatiate bookmarkPanel if we need it
+      if (showMainMenu && this.state.getStateProperty('mainMenuSettings').buttons.bookmark) {
+        this.bookmarkPanel = new $.BookmarkPanel({ appendTo: this.element.find('.mirador-viewer'), state: this.state });
+      }
 
       // set this to be displayed
       this.set('currentWorkspaceVisible', true);


rsinghal:
A few initial thoughts:
- The dropdown menu for border types is not quite aligned with the other two dropdown menus
- The dropdown menu for the border types does not have the same hover behavior that the other two dropdown menus have
- Can you remove the shadow on the icons in the dropdown menus as we don't use those anywhere else so they are not consistent with the Mirador design
- You can't refer to an image location using `background-image: url('../images/border_type_3.png');` because a developer can change the `buildPath` or `imagesPath` in the settings.  Therefore, it must be set in the JS until we can switch to LESS, which would then allow us to change the path dynamically.  Ideally, we really shouldn't have icons for these in Mirador, but point to them in a font file.  Do these exist anywhere? @aeschylus @davebeaudet
- In edit mode, I think it would be helpful for the mouse cursor to change to the pointer when the mouse is over a border - to indicate that the user can click and select it

stanislavkostadinov:
```
.mirador-main-menu .icon-bookmark-workspace,
.mirador-main-menu .icon-load-window,
.mirador-main-menu .icon-window-options,
.mirador-main-menu .icon-clear-local-storage {
  background-image: url('../images/mirador-icons.png');
```
This is part of the mirador.css and I was just trying to edit the CSS file in a way that the paths will be consistent.

stanislavkostadinov:
I've updated the source with fixes for the first three bullets. What should  I do with the URL paths and do I have to implement the fifth bullet or it can be done as a new PR?

rsinghal:
Yes, the fifth bullet can be done as a new PR.  I have added a ticket for it.  I would like @aeschylus to comment on the fourth bullet first before we decide to how to fix it (either JS or font file)

PR #830, opened 2016-04-09
Update utils.js

leedyjq:
Unless I'm misunderstanding the purpose of the lazyLoadingFactor, it appears that in the ThumbnailsView, thumbnails currently on screen are not loaded as expected, while most off the screen are. Reversing the sense of the boolean return value of the isOnScreen function seems to cause these thumbnails to be loaded as expected.

This behavior is seen with the default value of the lazyLoadingFactor (1.5).

### DIFF
--- a/js/src/utils/utils.js
+++ b/js/src/utils/utils.js
@@ -196,7 +196,7 @@
     bounds.bottom = bounds.top + el.height();
     bounds.right = bounds.left + el.width();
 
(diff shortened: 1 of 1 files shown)


rsinghal:
@leedyjq This breaks the main thumbnail view and if you view one of the first pages in a left-to-right manuscript.  Are you primarily working with right-to-left objects?  In general, this logic needs to be rethought, and it will after we have merged some other code that changes how the main thumbnail view is displayed.  That will allow us to write code and logic that is specific to the filmstrip.

Issue #764, opened 2016-02-22
Being able to edit a shape should be triggered when clicking on its border


rsinghal:
Rather than inside the shape, otherwise it's impossible to select a shape that is completely inside of another.

Also, the current behavior is when you click inside a shape the first time and you move the mouse, it immediately starts moving the shape. The initial click on the border should select the shape, but the user should have to click and move the mouse to move the shape.

#748

rsinghal:
@stanislavkostadinov Triggering a shape to be edited should occur when you click on a border.  Otherwise, it's 1) impossible to select a shape inside of another shape and 2) impossible to draw a shape inside of an existing shape.  

Please fork and work off the [refactor](https://github.com/IIIF/mirador/tree/refactor) branch as I have merged all of your annotation work into there along with refactoring that I did.  I had to move some code around, but it should generally look as you expect it to.

stanislavkostadinov:
I'm working on that problem and I'll create PR asap.

aeschylus:
What are the statuses of the annotation issues. Do any of the ones tagged "annotation" need to be updated?

rsinghal:
This hasn't been merged

Issue #784, opened 2016-03-07
Restyle dropdowns to have an arrow indicating their behaviour.


aeschylus:
<img width="462" alt="screen shot 2016-03-07 at 10 29 05" src="https://cloud.githubusercontent.com/assets/145874/13578921/7c36317e-e44f-11e5-8b69-1588e4fbd935.png">

PR #840, opened 2016-04-15
New border type tool: solid line, dashes, dots and dashes + bug-fixes

stanislavkostadinov:
Bug-fix: #764
CSS customization


rsinghal:
A few more things:
- The three drop downs for border color, fill color, and border type should all be drop downs that are triggered by hover/mouse enter.  It is inconsistent for two of them to have the downward caret but require clicking, while the third one does not require clicking but has no indicator that is a dropdown menu.  Following #784, all three should be dropdown menus that are triggered using hover/mouse over.  The border type menu does not need the pencil icon again, but instead needs the downward caret.
- Since the border type and border color are related, they should be next to each other. 
- Compared to the border color and fill drop down menus, the border type drop down menu has space between the icon and the actual menu.  These all need to be perfectly consistent.
- The texts for `Cancel`, `Choose`, and the border type names all need to use i18n, and cannot be hardcoded to English.

stanislavkostadinov:
I think the only think that I haven't worked on is the background URL. This type of URLs exist before my commit and I don't know what to do with them.

rsinghal:
The way we are currently handling background URL is to do something like this in the JS: https://github.com/IIIF/mirador/blob/master/js/src/viewer.js#L42.  It's not pretty and we will refactor it using LESS at some point in the future, but at least doing this guarantees that the images won't be broken for people who must put the `images` folder in a non-standard location.

stanislavkostadinov:
Done.

rsinghal:
A small typo in the English translation file: it needs to be `&amp;`, otherwise it doesn't display the text properly for the border types.

I'm also getting weird behavior where the color pickers still stay half visible when I hover away from the drop down menus:
screen shot 2016-04-25 at 4 35 13 pm
The border drop down doesn't have the same problem.  I can make it hidden by either clicking Choose/Cancel or moving my mouse onto a different icon in the annotation selection of icons.

Also, I'm encountering this weird behavior:

1. I click a shape to edit it
2. It's border and fill colors are different than the colors set in the two drop down menus
3. I hover over the two color drop down menus and hover away (I do *not* click Choose)
4. My shape's colors get changed to those that were in the drop down menus despite not actively Choosing them for that shape
5. I do not get this behavior with the border type drop down.

Everything else is working as expected.

stanislavkostadinov:
Weird effect was removed and hover effect over the color pickers works fine.

rsinghal:
I found something else that is inconsistent:

1. I refresh Mirador in my broswer and I go to each of the drop down color menus and click on one of the pre-populated colors, but I don't click `Choose`.  The color in the icon changes so it *looks* like I have selected those colors for my shape.
2. I draw a shape - it retains the original border color/fill (deepSkyBlue for border, no fill) despite the fact that the drop down menus show different colors.
3. If, after choosing the color, I click `Choose`, then I get the correct colors when I draw the shape.
screen shot 2016-04-26 at 2 32 39 pm
I compared this to `release2.1` and it properly changes the color if a user clicks on a color and doesn't click on `Choose`.

stanislavkostadinov:
Yes, that is because I removed changing color on close (the previous weird effect).

stanislavkostadinov:
I've added some modifications and the result is as described below:
- the only way to change the color of selected shape is by click on 'Choose' button.
- if user selects different color from default palette or by changing transparency, or color the color of selected shape is not updated by the next created shape will have the new changes.
I think that is the most flexible solution.

gigamorph:
There are couple of lines that fail when the window/object config has "annotationCreation" set to false:

https://github.com/stanislavkostadinov/mirador/blob/master/js/src/widgets/contextControls.js#L62
https://github.com/stanislavkostadinov/mirador/blob/master/js/src/widgets/contextControls.js#L118

since `fillPicker.find(".sp-cancel")` is empty when "annotationCreation" is false.

rsinghal:
Closing this since it will be redone as a PR to `annotation-work` per @beaudet

PR #847, opened 2016-04-18
Feature/2.1 search within

jeffreycwitt:
I don't think you'll want to accept all of this yet.
 
But commit: 4c08a7d fixes a general tab problem when multiple windows are up. 

John Abrahams really deserves credit for this fix: 

He writes: 

Hey Jeff,

> If this is still bugging you, there is an easy fix:
> 
> Original code in tabs.js (https://github.com/jeffreycwitt/mirador/blob/feature/2.1-search-within/js/src/widgets/tabs.js#L69):
> jQuery.publish('tabSelected.' + _this.windowId, jQuery( ".tabGroup li" ).index( this ));
> 
> Change to:
> jQuery.publish('tabSelected.' + _this.windowId, jQuery( this ).index());
> 
> The original code is selected all tabs on the screen, which in your example would be 4 total across both windows. The CSS selector is just too general in the original case and picks up more elements then it should. The modification is instead of selected ".tabGroup li", select the clicked element and take its position among sibling elements.


jeffreycwitt:
I'm going to re-submit this with just the change on line 69. So just ignore this. Sorry.

PR #884, opened 2016-05-05
Fixed New annotation dialog not showing up in fullscreen mode

yuricus:
Fixed positioning of qtip in DOM to fix annotation dialog not displayed in fullscreen mode.


aeschylus:
Thanks for the fix Yuri.

PR #885, opened 2016-05-05
Fixed DEFAULT_SETTINGS overwrite in SaveController initialization

yuricus:
This is extremely important when reusing Mirador withing a page multiple times, otherwise you've got your DEFAULT_SETTINGS rewritten.

### DIFF
--- a/js/src/utils/saveController.js
+++ b/js/src/utils/saveController.js
@@ -22,8 +22,7 @@
       config.mainMenuSettings.buttons = {};
     }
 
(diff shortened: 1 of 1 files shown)


aeschylus:
Good catch. We should look across the app to see where this is not used. It's a but confusing in that including the true/false argument updates the function signature, making the second argument the target (instead of the first argument). 

It's fairly often that we've run into issues of data mutating unintentionally in the application, and it looks like this will solve that (create a new object each time, instead of mutating the original).

PR #912, opened 2016-05-08
Drag and drop not working in Chrome

### LINKED ISSUE #911: Drag and drop not working in Chrome
DnD does not work with Chrome as source and target.

mikeapp:
Fixes #911   Prefers 'text/plain' content from drag and drop event, falls back to Chrome-specific method otherwise.

We might want to rethink our drag-and-drop recommendation, it would be best if the data were explicitly passed from the drag source (e.g., set as a URI).

### DIFF
--- a/js/src/workspaces/slot.js
+++ b/js/src/workspaces/slot.js
@@ -132,8 +132,20 @@
       var _this = this;
 
       e.preventDefault();
-      var url = e.originalEvent.dataTransfer.getData("text/plain");
-      if (url) {
+      var text_url = e.originalEvent.dataTransfer.getData("text/plain");
+      if (text_url) {
+        _this.handleDrop(text_url);
+      } else {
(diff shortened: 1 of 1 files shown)


aeschylus:
Nice. Thanks Mike. Should have added a fallback. I do wish we could understand why this doesn't work; I don't want to relinquish my fantasy that Chrome is a standards-compliant browser.

I was just doing some experimentation with this, and realised that it's quite easy to embed drag-and-droppable links anywhere (including emails). Try dragging these into Mirador (both from the internet archive, which doesn't present dnd icons yet).

*Image*
[image](https://archive.org/details/mma_irises_436528?image=https://iiif.archivelab.org/iiif/mma_irises_436528/info.json)
*Book*
[image](https://archive.org/details/platowithenglish04platuoft?manifest=https://iiif.archivelab.org/iiif/platowithenglish04platuoft/manifest.json)

PR #919, opened 2016-05-10
added more materials from various institutions

aeschylus:


rsinghal:
I think we should reduce the number of Harvard and Stanford samples and make an effort to represent more institutions in the example index.html.  This PR feels a bit of an overkill on Stanford objects, IMO.

regisrob:
Since Gallica is now exposing his own manifests, I think we could replace the two Biblissima manifests with these one (or any other one if you prefer):
http://gallica.bnf.fr/iiif/ark:/12148/btv1b84539771/manifest.json
http://gallica.bnf.fr/iiif/ark:/12148/btv1b10500687r/manifest.json

rsinghal:
I should also change the oculus-dev links to point to the actual Library and Museum manifests since those are the "official" manifests.  I am going to close this PR and we can think through a more comprehensive list of manifests, as well as trim the number of manifests from Harvard and Stanford.

PR #920, opened 2016-05-10
updated project with locales for GB and 'dev', prevented the viewer i…

aeschylus:
…nitialisation from being synchronous, which was causing warnings in Chrome and Firefox.


aeschylus:
After reading through some documentation on i18next, it seemed best not to make them symlinks for now, since they are likely to change separately. Git can handle symlinks if we decide we do decide to go down that route.

rsinghal:
We need to figure out how to make this more efficient.  Having to update 3 files for English is not great.

PR #922, opened 2016-05-11
Adaption of the thumbnail

fe-kre:
I am sorry to open a new PR, but I am not sure how to open a closed PR. I refer my point of view to #809.

We have taken a new approach to our problem and have resolved it with the help of css properties.

I set up a new div around 'preview-images' and gave him the properties position: absolute and display: inline-flex. So our outsized thumbnail shows up.

Hopefully this is an acceptable way to deal with our problem.

### DIFF
--- a/js/src/annotations/annotationTooltip.js
+++ b/js/src/annotations/annotationTooltip.js
@@ -6,6 +6,7 @@
       targetElement: null,
       annotations: [],
       windowId: "",
+      eventEmitter: null
     }, options);
 
     this.init();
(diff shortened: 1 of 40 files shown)


rsinghal:
screen shot 2016-05-11 at 3 13 26 pm

This is how it looks on my screen, does it look like this for you?  In your previous PR the thumbnail was on the same horizontal level as the title.  It may also be worth turning off the scaling of the thumbnail for these really wide images.

fe-kre:
viewer

On my screen, it works as expected. Maybe `display: inline-flex` does not work correct with some browsers? For me, it works with Firefox and Chromium.

css/mirador.css
```diff
@@ -400,6 +400,10 @@ a {
   margin-right:15px;
   height: 80px;
 }
+.previev-thumb {
```
rsinghal: Aha!  There is a typo in this class name

fe-kre:
I'm so sorry for this little mistake. I correct it.

PR #926, opened 2016-05-12
fixed flashing by switching out jquery maps for native filters+maps. …

aeschylus:
…This causes some some mis-match between the open and close and selected items, and requires some refactoring.


aeschylus:
The problem is fixed. Please merge.

PR #931, opened 2016-05-17
Thumbnail

fe-kre:
I am sorry again for my approach, I am really new to Github and learn their procedures.

I correct my mistake at https://github.com/IIIF/mirador/pull/922.

Hopefully, now everything works as expected.

### DIFF
--- a/css/mirador.css
+++ b/css/mirador.css
@@ -400,6 +400,10 @@ a {
   margin-right:15px;
   height: 80px;
 }
+.previev-thumb {
+  display: inline-flex;
+  position: absolute;
+}
 .preview-images {
   margin-right:0;
 }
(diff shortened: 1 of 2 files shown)


fe-kre:
Is there any problems with my PR? Shall I correct something?

rsinghal:
I haven't had a chance to look at it yet, but plan to today.

rsinghal:
@fe-kre It looks fine.  In general, I'm not sure what the best behavior is when the user resizes the window  down.  The default behavior is to have the `...` display when it is not possible to display a full thumbnail, but there is an argument to be made that it should follow the behavior of your large scroll - always show a little bit of the first thumbnail.  I am going to ticket that separately and merge this.

PR #942, opened 2016-05-30
Added fake local JSONBlob endpoint.

dicksonlaw583:
This is designed to replace the default JSONBlob endpoint for bookmarking in demo setups, removing the dependency on jsonblob.com.

### DIFF
--- a/js/src/utils/localJsonBlobApi.js
+++ b/js/src/utils/localJsonBlobApi.js
@@ -15,21 +15,19 @@
     
     $.LocalJSONBlobAPI = function(opts) {
           this.options = {
-              storage: "localStorage"
+              storage: window.localStorage
           };
-          for (var key in opts) {
-              this.options[key] = opts[key];
-          }
+          jQuery.extend(this.options, opts);
     };
 
     $.LocalJSONBlobAPI.prototype = {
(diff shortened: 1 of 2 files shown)


aeschylus:
This seems like a good addition that I'm happy to merge. Could you write a few tests for this (coverage is decreasing)?

Long-term I think we need to figure out how to remove jsonblob entirely from the codebase and just let people get the internal state from the API, perhaps following that up with a guide on how to persist said state in various ways.

Thanks @dicksonlaw583.

dicksonlaw583:
Tests added for the local JSON Blob API.

aeschylus:
https://github.com/IIIF/mirador/wiki/Configuring-Persistent-Bookmarking-in-Mirador

dicksonlaw583:
Wiki page: https://github.com/IIIF/mirador/wiki/Configuring-Persistent-Bookmarking-in-Mirador

aeschylus:
I made some minor edits to the wiki page, but otherwise this looks good for now.

We will probably remove the feature entirely in the upcoming release, to facilitate different methods of persisting the state, but for now this will help development, and provide a way forward if that falls out of the release.

PR #945, opened 2016-06-02
Add workaround to fix #929

### LINKED ISSUE #929: FF: Mirador in second tab is empty after loading
[mirado2.zip](https://github.com/IIIF/mirador/files/295017/mirado2.zip)

Loading mirador in iframe inside `display:none` parent causes the image to be loaded but not shown.

Steps to reproduce
Open test.html
Wait until the image is loaded.
Go to tab 2 
See that there is no image shown, although it loads the image and if you go to the element which displays the image and set its style to `display:block` it will show it.

It happens under Firefox only. Maybe it is because mirador does not handle the event that sets its canvas to be visible.

radpet:
Add workaround to fix #929 by manually setting the display to block inside the complete animation callback. I tried to find all .show() functions that might have been affected and add the fix.

I would like to ask what code beautifier to use in order to have consistency with the other contributors.

### DIFF
--- a/js/src/viewer/workspacePanel.js
+++ b/js/src/viewer/workspacePanel.js
@@ -94,7 +94,15 @@
     },
 
     show: function() {
-      jQuery(this.element).show({effect: "fade", duration: 160, easing: "easeInCubic"});
+      var onComplete = function () {
+        // Under firefox $.show() used under display:none iframe does not change the display.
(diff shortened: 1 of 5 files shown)


rsinghal:
@radpet Can you resubmit your PR without all of the whitespace changes?  I can't see what changes you made.

Issue #680, opened 2015-10-27
TOC module does not order correctly


thehabes:
Hey all,
I came across a bug today in the viewer.  I have a range assortment like so

Table of Contents (root)
     Section 1
     Section 2
     Section 3
     Section 4
     Section 5
     Section 6

Each of those sections have varying amounts of child ranges.

Mirador is displaying it as

Table of Contents(root)
      Random orders, depending on the last changes I made to the data set.

When I look into the Network tab for the response of the manifest, all of the data relationships are correct.  However, Mirador populates the children based on the order they were returned from the server.  My response for the manifests contains the structures field which outputs

    0: Table of Contents
            ranges:[Section 1, Section 2, Section 3, Section 4, Section 5, Section 6]
            within: root
    1: Section 1
            within: Table of Contents
    .
    .
    .
    12 : Section 3
            within: Table of Contents
    13: Section 4
            within: Table of Contents
    14 : Section 2
            within: Table of Contents
    .
    .
    .
    25 : Section 6
            within: Table of Contents
    26: Section 5
            within: Table of Contents

Mirador then displays the data set as:

Table Of Contents
        Section 1
        Section 3
        Section 4
        Section 2
        Section 6
        Section 5

This happened many different ways, in many different orders depending on the last changes I made to the data set.  One thing is always consistent;  it always orders things based on the response from the server, not the assortment the data is providing.  

I suspect that this is because the 'within' property is used to build the structure from the lowest children to the root. I have a page that does this same thing but builds the structure from the root down to the children.  I use the ranges:[] property to build the order so that how it is returned from the server does not affect the order of display.  Users are asking why the order is different in the viewer than as it is displayed on my other page.

rsinghal:
@thehabes, do you have a sample manifest we can test with that does this?

thehabes:
Yes, use this API call to get the manifest:
http://165.134.241.141/brokenBooks/getManifest?username=debra

The Mirador rendered version can be found at
http://165.134.241.141/brokenBooks/getManifest.html

We had to trick the server into making it mostly appear in order.  Any time we save a new range or canvas and order it, the order is not preserved.  You will see this happen with a range named BB-Llang_0065, which appears before BB_Llang_0001 in the range named Temporale even though BB_Llang_0065 is 64 positions in the array after BB_Llang_0001 but is returned by the server before BB_Llang_0001.  You will also see that its canvases appear last in the thumbnail strip, which does not reflect the order displayed in the TOC. 

[...]

Issue #527, opened 2015-04-27
Presentation API 2.0 support for ranges and table of contents display


snydman:

thehabes:
I have written an algorithm using recursion that takes a collection of ranges, in any order, and displays them from largest collection to smallest in a 'table of contents' style display so that there are no repeats and the proper parent-child relationship is maintained throughout.  I had to do this for one of the projects that SLU has taken on.  The canvases contained within the ranges display with all of their annotations. I planned on showing it at the Monday, 5/4 IIIF event.

In conjunction with this piece, I am working on a "re-ordering" interface so that ranges can be removed from or included in other ranges in a 'drag-and-drop' style, but this will not be prepared in time for the event.

thehabes:
http://165.134.241.141/brokenBooks/getManifest.html
--There's a sneak peek.  The code I have on my local machine is already improved.  The things that are out of order are so on purpose to I can work on my re-ordering interface.  Things will make much more sense when I demo.

aeschylus:
Here is some more data provided by Sean Martin with the Vatican. The relevant code is in a recursive method in the toc module. I'll have a look at how it can be adapted to handle 2.0 manifests better than currently, but it looks like there may be other issues besides. Paste follows:
The issue with display and navigation of table of contents with Mirador 2

Preamble

The web site video at https://www.youtube.com/watch?v=9b1ReCZh9-E   has at around time 1:00 

mins a good looking table of contents (TOC) with indents - see attached screen shot below:

I have located the Harvard manifest and been able to reproduce the table of contents (TOC) with a 

similar user experience. The indents can be controlled to collapse and expand, and also one can 

navigate by clicking on an entry. So far so good …

However I have tried but failed to reproduce a similar behaviour with a TOC for one of the Vatican 

manuscripts.

Rob has confirmed that Mirador has so far implemented version 1.0 of the IIIF spec regarding ranges, 

but has not yet implemented version 2.0. The structure of these versions is different: version 1.0 is 

essentially bottom-up and version 2.0 is top-down.

Background and timing

The Vatican pilot project funded by Mellon will conclude at the end of June. It comprises ten 

manuscripts which will be viewed by Vatican staff. Most of them have rich structural information, 

hence a good TOC presentation and navigation will be important. There is thus a timing issue about 

when the issue might be resolved - in essence the sooner the better.

It will be desirable for us to have a working version by mid May, so that we can conduct testing, and 

then make a release with the set of ten manuscripts by the end of May.

[...]

Issue #375, opened 2015-01-08
Simple Image Manipulation Requests


thehabes:
Hi all. I have users that have requested simple 90 degree image rotation as a tool (both positive and negative) for your viewer on a site I am building. There are also request for toggling greyscale and color invert as well as requests for brightness and contrast sliders. I was able to successfully build these requests into the viewer using your coding schema and introduced buttons into the UI.
Do you feel that these bits of functionality are worthy of a pull request?

azaroth42:
Hiyas!  I'm very curious as to how you implemented these :)  Could you give a link to an installed instance to look at? Thanks

thehabes:
I forked the project and have everything on my github you can find the code here: https://github.com/thehabes/m2

Everything I am doing is installed locally and on my work machine and is in the beginning stages, so there is no link I can give you at this moment for testing.  If you take my mirador and build it on your machine, you will see the buttons for rotation in a <div> element with the class mirador-pan-zoom-controls hud-control towards the bottom right corner of whatever SINGLE image is active.  I have not implemented any rotation in the side by side or book view but I am thinking about it.  The changes I made were to hud.js and the mirador.js that came out of building with grunt as well as mirador.css which had to be 'grunted' into mirador-combined.css to take effect.  

Specifically, in mirador.js in bindEvents (line 6617) I added code starting at 6682 and in hud.js I altered the template object on line 203 to add my buttons.  I did the same thing to the object in mirador.js on line 6808.  

If you need something more than what is on my github account, let me know.  I will commit everything now.

aeschylus:
This is exciting. I have cloned your fork and had a look. I see your controls (nice job modifying the h.u.d.), but they don't seem to cause a rotation (not in single mode either).

thehabes:
Ok that is not good.  It is working on my machine.  I can positive and
negative rotate ninety degrees and do 180 degree flips as well.  I have it
working in Chrome and Firefox.  Does it happen to give you an error?

On Fri, Jan 9, 2015 at 11:36 AM, aeschylus <notifications@github.com> wrote:

> This is exciting. I have cloned your fork and had a look. I see your
> controls (nice job modifying the h.u.d.), but they don't seem to cause a
> rotation (not in single mode either).
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/IIIF/m2/issues/375#issuecomment-69368860>.
>



-- 
-------------------------------------------------------------
Bryan J. Haberberger
[...]

PR #948, opened 2016-06-03
Release 2.1 TOC Ranges Ordering Fix

### LINKED ISSUE #939: TOC child range toggle not working in release 2.1
http://165.134.241.141/brokenBooks/getManifest.html

When I click on a caret, _this.render() always fires

toClose = _this.previousSelectedElements.filter(function(rangeID) {...})

and never 

toOpen = _this.selectedElements.filter(function(rangeID) {...})

It seems like no matter what, it is always trying to close the first listed range and the variable toOpen is always an empty array.  See the web console log for additional information.

thehabes:
**The main fix for this pull request:**
This fix is all in toc.js with some select mirador.css changes, see the function extractV21RangeTrees and the toc-caret click event handler.

This fixes #939, #680 and #527 and an issue not yet brought up:  what happens when a range is within more than one range?  Right now, the within property in mirador doesn't support that.  The ranges[] property does.  The V2 algorithm would just ignore it, the V21 algorithm will support it and place it correctly.

**Extras included with this pull request:**
This code also has the fix for #375.  For a demo on how I have placed the image manipulation tools in with the other hud controls, visit [this page](http://165.134.241.141/brokenBooks/getManifest.html)
***anything with the word bbAnno in it can be stricken from the pull request, it is specific to making it work for my application.  

This code also supports Non-IIIF images (that is, images not fed via a tiling service or from a IIIF server).  For this code, see the .fail() function in imageView.js (line 954)

This code supports default images for "Image Not Found" and "Image Not Resolved" situations.  See manifest.js, iiif.js, utils.js.  The images used for these are hard linked, feel free to replace them or ignore this code entirely.


rsinghal:
@thehabes This is a bit too much for one PR.  

Can you please resubmit with just the isolated fixes to individual items?  For example, one PR for non-IIIF images, another PR for 2.0 ranges, another PR for "Image Not Found" situations.  Having smaller, incremental changes is the best way to determine that the small changes work in isolation.  Also, the current PR reduces coverage by 5.4% - can you please include tests with the individual PRs?

Also, I see in your changes to `toc.js` that you commented out the `switch` statement that determines which type of range the manifest contains, and therefore, how to process it.  This cannot be removed because Mirador must be able to support 1.0 and 2.0 ranges.

thehabes:
Sure, I can break these out more.  I will un-comment the switch, that was from testing it should not have been left commented.

Issue #918, opened 2016-05-09
In-manifest annotations not being aligned properly with canvas.


aeschylus:
<img width="471" alt="screen shot 2016-05-09 at 10 02 43" src="https://cloud.githubusercontent.com/assets/145874/15128383/8c7149ac-1609-11e6-981c-28cd76183130.png">

aeschylus:
Try the Oto Ege and open annotations.

rsinghal:
The manifest is http://dms-data.stanford.edu/data/manifests/Stanford/ege1/manifest.json

stanislavkostadinov:
I'm not a 100% sure, but I think there is a bug-fix assosiated with that problem over here: #840

PR #951, opened 2016-06-09
Bug-fixes & functionality that replaces comment popup&tooltips

radpet:
Bug-fixes & functionality that replaces Mirador comment popup and tooltips with external ones

Functionality that replaces Mirador comment popup and tooltips and notifies external applications
#764
Bug-fix: in-manifest annotations not being aligned properly with canvas
#918

### DIFF
--- a/css/mirador.css
+++ b/css/mirador.css
@@ -1171,6 +1171,68 @@ li.highlight {
 /* color picker custom style
 ---------------------------------------------------------------------------- */
 
+.draw-tool * {
+  text-shadow: none !important;
+}
+
+.draw-tool>i {
+  text-shadow: 0 0 5px black !important;
+}
+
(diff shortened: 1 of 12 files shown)


radpet:
I extracted some commits from Stanislav's work in order to review them faster. I need to fulfill one more todo task but you can check the other code.

rsinghal:
I have merged this and will start testing it and create tickets for anything I find.  @radpet @beaudet

radpet:
@rsinghal 
Okey I'll prepare another one soon. I need to organize them a little bit. I think it would be better if you check this one before we merge the other in order to track any issues easier.

rsinghal:
Sounds good.  I am also fixing some things and will be pushing those changes to this branch as well.

PR #952, opened 2016-06-10
Fixed unit tests & remove useless pin handlers

radpet:


rsinghal:
@radpet is there a way to make the pin a bit more pointed, like it was before?  It seems less like pin drop and more like an oval that is slightly pointed on one side.  @beaudet

radpet:
@rsinghal  @beaudet Yup no problem

radpet:
@rsinghal  I can make the pin like before but I cannot remove the 4 point handles.However what do you think about returning the old pin and making it bold when selected. I does not need those handles because a pin cannot be resized or rotated so they are useless, aren't they.

beaudet:
Radoslav,

Yes.  No handles would probably be better than one.  However, there needs to be some way to highlight the pin when the annotation is selected.  That is not modeled in the current designs for the annotation editing modifications.  How do you propose to highlight pin annotations?  Animating the bounding box with no handles seems like it would be consistent with the rest of the proposed design.

Thanks,

Dave

radpet:
@beaudet  I think they should act like all other shapes except that they won't have handles when focused. As far as highlighting is concerned I think that there won't be any difference if you highlight rectangle or pin or any other shape.

@rsinghal If everything else is okey and you are good to return the old pin, merge this and Ill include it in the new pull request.

rsinghal:
@radpet Yes - that sounds good.

Issue #941, opened 2016-05-30
drag and drop for a one level deep collection


jeffreycwitt:
Is there any reason Mirador shouldn't be able to handle a drag and drop collection for a one level deep collection? Likewise, I was under the impression that if I had a one level deep collection, I should be able to add the link directly to the manifestList page and see the results. 

Neither are currently working for me.

Here's where I'm currently displaying the drag and drop link.
[http://scta.lombardpress.org/text/questions/plaoulcommentary](http://scta.lombardpress.org/text/questions/plaoulcommentary) (note that I've I change this drag and drop link to a simple manifest everything works. 

And here's the actual collection: 
[http://scta.info/iiif/plaoulcommentary/collection](http://scta.info/iiif/plaoulcommentary/collection)

PR #956, opened 2016-06-13
modified slot.js to allow for collection drag and drop

jeffreycwitt:
modified slot.js to allow for collection drag and drop that then moves user to manifest list view.

Should resolve issue: #941

### DIFF
--- a/js/src/workspaces/slot.js
+++ b/js/src/workspaces/slot.js
@@ -149,6 +149,7 @@
 
         url = url || text_url;
         var manifestUrl = $.getQueryParams(url).manifest,
+            collectionUrl = $.getQueryParams(url).collection,
             canvasId = $.getQueryParams(url).canvas,
             imageInfoUrl = $.getQueryParams(url).image,
             windowConfig;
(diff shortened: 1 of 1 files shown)


js/src/workspaces/slot.js
```diff
@@ -171,11 +172,31 @@
...
+                  _this.eventEmitter.publish('ADD_MANIFEST_FROM_URL', mfst['@id'], "(Added from URL)");
+                }
+              });
+            }
+          });
+
+          //TODO: 
+          //this works;
+          //but you might want to check if some "publish" action would be better
+          _this.addItem();
```
aeschylus: Good point, but I think this will be okay for now. We need to think about how the different top-level screens can be represented in the data structure, and how to update them accordingly without calling the methods of things. 

Any update needs to ultimately be done through an "action", so that the state manager can send out notifications to subscribers(UI components) that depend on that state.

js/src/workspaces/slot.js
```diff
@@ -171,11 +172,31 @@
...
             _this.eventEmitter.publish('ADD_MANIFEST_FROM_URL', imageInfoUrl, "(Added from URL)");
           }
-        } else {
+        } 
+        else if (typeof collectionUrl !== 'undefined'){
+          jQuery.getJSON(collectionUrl).done(function (data, status, jqXHR) {
+            if (data.hasOwnProperty('manifests')){
+              jQuery.each(data.manifests, function (ci, mfst) {
+                if (!_this.state.getStateProperty('manifests')[imageInfoUrl]) {
+                  _this.eventEmitter.publish('ADD_MANIFEST_FROM_URL', mfst['@id'], "(Added from URL)");
```
aeschylus: I'm loving this new eventEmitter stuff. It's almost like we're a real application.

aeschylus:
This seems like a good way forward using existing patterns. Although the coverage has decreased, I'm adding going ahead with it since the slot.js d-n-d code needs some serious refactoring to be testable. The different branches seriously need to be broken out into functions.

jeffreycwitt:
Hooray!!!

On Wed, Jun 15, 2016 at 3:19 PM, aeschylus <notifications@github.com> wrote:

> Merged #956 <https://github.com/IIIF/mirador/pull/956>.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/IIIF/mirador/pull/956#event-693716687>, or mute the
> thread
> <https://github.com/notifications/unsubscribe/ABF_PVg8AjgCJ0cC_dJEKogS0zyE5Fy0ks5qMFAxgaJpZM4I0t0K>
> .
>



-- 
Dr. Jeffrey C. Witt
Philosophy Department
Loyola University Maryland
4501 N. Charles St.
Baltimore, MD 21210
www.jeffreycwitt.com

Issue #955, opened 2016-06-13
Make it possible to drag-and-drop direct manifest URLs


dicksonlaw583:
Mirador should accept drag-and-drops from direct links such as this one, instead of crashing:

[Direct manifest link](http://wellcomelibrary.org/iiif/b18035723/manifest)

The current implementation complains about not finding the method `split` for `undefined`.

jrhoads:
Looks like this could be closed.

PR #957, opened 2016-06-14
Fix for #955: Ability to drag and drop direct manifest URLs

dicksonlaw583:
Links to manifest URLs can now be directly dropped into empty Mirador slots without errors.

See: #955


aeschylus:
Hi @dicksonlaw583. Can you synchronise this with release 2.1? I merged Jeffrey's first because it touched fewer files. There seems to be a mild conflict.

dicksonlaw583:
Should be good to go now.

Also, while you are at it, please also consider #959.

aeschylus:
Thanks @dicksonlaw583, merging both of these now.

PR #959, opened 2016-06-15
Fixed #958: Drag-and-drop not working on IE

### LINKED ISSUE #958: Drag and drop not working in Internet Explorer
Drag and drop in Internet Explorer replaces the content of the browser window with the dropped link.

dicksonlaw583:
Drag-and-drop no longer crashes on IE.

See: #958


aeschylus:
@dicksonlaw583 Rashmi and I talked about these and I think this could be improved a bit. The test in particular seems fairly basic (and perhaps we should use something like "example.com" rather than google.com). It would be good to add a test to assure that the URL resolves to an actual manifest/info.json/collection if possible, since different things are done with these.

dicksonlaw583:
The function I edited was just a helper for extracting URL parameters from after `?`, so the tests involved would be fairly basic. Ensuring that it resolves to something IIIF-friendly is an entirely separate task, in my opinion.

rsinghal:
Closing this so we can delete `release2.1`

PR #966, opened 2016-06-20
Vertical right2left

knagasaki:
I add a function to display thumbnails and images from right to left in some modes by use of the viewingDirection value in IIIF manifest and adjust tooltip size for vertical writing system (only delete the overriding css). It would strongly support to adopt IIIF in East Asian institutions. However, it might not be matured as a style of recent programming. Please revise it as you prefer if necessary.


aeschylus:
@rsinghal I am thinking of merging this even though some of it will be replaced by the ML/"Almagest" work, since it it fairly unobtrusive and does not affect anything else. The only thing I'm worried about is the change to the css (and test coverage - it occurs to me that we may be including the locales files as "lines of code" that the coverage tool is expecting to be tested. This doesn't seem quite right. We definitely want to test the localisation, but not the files themselves. I'll look into adjusting this.)

@knagasaki Thank you so much for this contribution. I saw your post on the IIIF-discuss mailing list and have a response incoming about the right-to-left work that is currently being done in Mirador, which we could really use your help with. In the mean time, would you consider making a separate pull request that includes only your locale file and your additions to the index.html? We can orient you to the other r-t-l work going on in another repo soon. Thanks you again; I'm very excited to see East Asian fixtures and adoption.

rsinghal:
I think we should keep it open so we can reference it as we work on iiifManifestLayouts, but I wouldn't merge it right now

aeschylus:
@rsinghal Sounds good.

knagasaki:
Dear all, thank you for your consideration and comments. I will follow you, @rsinghal.

Issue #950, opened 2016-06-09
Extended navigation


morpheus-87:
At the Munich Digitization Center (Bavarian State Library) we developed a new feature for the Mirador viewer, an extended navigation. With that extension you can not just scroll to the next   respectively the previous page, but also with an offset of 3, 5 or 10 pages and the first/last page.

A demo can be found here: https://digipress2.digitale-sammlungen.de/view/bsb10504874_01404_u001/8

For illustration purpose I also add two screenshots:
![screenshot-digipress2 digitale-sammlungen de 2016-06-09 09-41-28a](https://cloud.githubusercontent.com/assets/19190936/15922192/cf9e7076-2e26-11e6-84bf-a3fc41fd4751.png)
![screenshot-digipress2 digitale-sammlungen de 2016-06-09 09-54-19](https://cloud.githubusercontent.com/assets/19190936/15922504/4eb93e58-2e28-11e6-982c-da4d4be5e901.png)

I would like to provide this feature for the community.

aeschylus:
This is an interesting contribution and reminds me of some comments made by Eben English a the year before last about book use-cases specifically. I will bring this up to the community and see what they think on the call this Thursday.

It would be great to have you present the interface on the call this Thursday if you can make it. We announce all the details on the [mirador-tech](https://groups.google.com/forum/#!forum/mirador-tech) and [iiif-discuss](https://groups.google.com/forum/#!forum/iiif-discuss) mailing list, and on the [Mirador Gitter](https://gitter.im/IIIF/mirador).

@rsinghal @jeffreycwitt To me this is a perfect example of something that could make a "plugin", maintained in a separate repo, that people can compose into their Mirador instance. Not sure how that would work exactly, but it might make a good thought experiment for the implementers call. Perhaps @edsilv can give us some insight about that.

datazuul:
I will present this today in the Mirador call as @morpheus-87 can not participate today. He will join the call in the future. I will announce this in mirador-tech mailing list.

PR #967, opened 2016-06-21
Extended navigation

morpheus-87:
At the Munich Digitization Center (Bavarian State Library) we developed a new feature for the Mirador viewer, an extended navigation. With that extension you can not just scroll to the next respectively the previous page, but also with an offset of 3, 5 or 10 pages and the first/last page.

Fixes Issue #950, which contains example usecase.


aeschylus:
Closing this as a pull request, but we will still have it in our closed PRs. Let's discuss how to make this into a plugin for others to use as well. I would like to know where we can host work of this kind in the documentation.

PR #972, opened 2016-06-23
Implementation of new ellipse  functionality

radpet:
@stanislavkostadinov 
Implementation of new ellipse  functionality
Update of ellipse tool tests

TODO: backward compatibility? The old ellipse shapes will be broken when selecting them with the new tool. What shall we do?


radpet:
This pr includes the new rectangle deformation functionality. When we merge the rectangle I will rebase :)

rsinghal:
@radpet I'm not sure we need to worry about backwards compatibility right now since all of this work has not been officially released, so it's all still under development.  The only thing we have to make sure is that the rectangle annotation tool can support annotations that use the format xywh, rather than the SVG format, since those are valid annotations.

radpet:
No worries then I have already implemented a fix for the old rectangles, therefore there was a support for the xywh I think so everything should be okey. Check the ellipse out and lets merge it :)

rsinghal:
The ellipse stuff looks good, so I'm going to merge, but it looks like something isn't working correctly with the xywh annotations coming from this manifest, http://dms-data.stanford.edu/data/manifests/Stanford/ege1/manifest.json.  The annotations display in release2.1, but not in this PR.

radpet:
@rsinghal I will check it I have not tested it against such shapes thanks

Issue #977, opened 2016-06-27
Open mirador in full-screen mode


radpet:
When I try to activate full screen with opened image the following error is thrown

mirador.js:25756 Uncaught TypeError: _this.fullScreen is not a function

I'll open pull request if noone is working on it now

The fix I have done is to rename the function because it seems something sets the same variable to bool :)

radpet:
@rsinghal do you want to fix it by renaming the function to toggleFullscreen for example?

rsinghal:
Yeah, I'll take a look at it

radpet:
@rsinghal Should I send pr to release 2.1 with the fix?

rsinghal:
yes

radpet:
https://github.com/IIIF/mirador/pull/981

PR #981, opened 2016-06-27
Fix exception thrown when entering in fullscreen mode

radpet:
Rename the function because it seems the same variable name has already been used to hold to bool state


rsinghal:
closes #977

PR #989, opened 2016-06-30
Image manipulation tools integration with release 2.1

thehabes:
I think I did this correctly with the branching this time.

I did not write tests or anything, but I wanted to get this up since others working with it.  I am not sure how to write tests for it, I will need your help with that.  It's definitely not perfect, but its as worked in as I could get it.  For a running mock up of the UI/UX, see http://165.134.241.141/brokenBooks/getManifest.html.


thehabes:
Meh got it wrong, had the range code too!

PR #992, opened 2016-06-30
Image manipulation tools

thehabes:
This is the image manipulation tools integrated with the release2.1 branch. I have built the new objects into the handlebars template in the hud. I have accounted for their event handling in imageView.js and created the CSS necessary to handle them in mirador.css

I did not write any tests, mainly because I am not sure how to do that :( but I am happy to do so, I would need some pointers to get started. I wanted to get this up since others are working on it and I at least have a working version with an idea of how to get it all into the UI. To see how the UI/UX is implemented, see the working version at http://165.134.241.141/brokenBooks/getManifest.html

### DIFF
--- a/css/mirador.css
+++ b/css/mirador.css
@@ -1166,6 +1166,143 @@ li.highlight {
   font-weight: normal;
 }
 
+/* Below are image manipulation additions */
+.mirador-osd-contrast{
+  font-size: 103%;
+  color: #FFF;
+  width: 175px;
+  display: block;
+  position: relative;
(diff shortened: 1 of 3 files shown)


aeschylus:
We weren't able to merge this directly as is/was, due to the controls not following the new designs (#789). However, a lot of the browser API-related code (css filters) was used in the PR that addressed that.

Closing this for now. Have a look at https://github.com/IIIF/mirador/blob/release2.1/js/src/widgets/imageView.js#L305-L474

PR #996, opened 2016-07-01
select canvas by index

nicolasfranck:
I want to be able to select a canvas by index, not just by canvasID.
When both are given, canvasID is preferred.

### DIFF
--- a/js/src/workspaces/window.js
+++ b/js/src/workspaces/window.js
@@ -84,7 +84,9 @@
 
       _this.imagesList = _this.manifest.getCanvases();
       if (!_this.canvasID) {
-        _this.canvasID = _this.imagesList[0]['@id'];
+        var ci = isNaN(this.canvasIndex) ? 0 : parseInt(this.canvasIndex);
+        ci = ci >= 0 && ci < _this.imagesList.length ? ci : 0;
+        _this.canvasID = _this.imagesList[ci]['@id'];
       }
 
       this.annoEndpointAvailable = !jQuery.isEmptyObject(_this.state.getStateProperty('annotationEndpoint'));


rsinghal:
@aeschylus This was submitted 8 hours go, but Travis is still not done, so it can't be merged.  Do you know what's going on?

aeschylus:
This is really weird.

aeschylus:
I'm not sure what's going on with travis, since this build doesn't even appear anywhere in the travis dashboard.

However, I think we should close this pull request, since this is not really the place for this functionality. This adds a kind of "shadow API" that is hard and inconsistent to document. An input should have only one type, and we should not be doing type checking and coercion inside the library code. Can we add a method like `selectCanvasBySequenceIndex()` instead?

I especially don't know why this check has to go in the initial setup code. Can't the programmer get the canvas ID himself and pass that in?

On the other hand, one advantage of using the sequence index is that canvas IDs don't have to be unique within a sequence (I believe that's right, though it would be strange to repeat them), so I could be convinced to switch it over to the index instead of the ID, but having that configuration field accept either seems strange.

aeschylus:
We should continue to discuss in an issue. Closing for now.

nicolasfranck:
I created an issue for this: https://github.com/IIIF/mirador/issues/999

I totally agree: only one attribute should influence the behaviour, either canvasID or canvasIndex.

I did not create this because I thought that the canvasID wasn't unique enough (it should, it is an identifier!). Here is my use case:

In the current situation, mirador is a multi-window environment. State cannot be saved so easily in the URL because of the complexity of this structured information. But in my case, I'm using Mirador in Zen Mode, i.e. with only one manifest open, and no options available to add/remove any windows. 
In that case the state information is flat (is it?).

Of course, I cannot create a state controller specifically for Zen Mode, and assume that no windows will be added. So I disabled the state controller (because it assumes multi-window environment).

I saw that you can supply the openseadragon bounds to the windowObject configuration. So I fetched
this myself from the URL fragment added it to the configuration. But I cannot restore the canvas so easily without storing the canvasID in the URL. The canvasIndex seems like a cleaner,shorter and more safe way to do this.

This is what I propose: is there a way to influence the mirador instance before it starts rendering?
Now everything starts automatically after configuration. But if you add a method like "start", only in that case would a method like "selectCanvasByIndex" be usefull. Maybe a configuration option like "auto"?

PR #997, opened 2016-07-01
Add stroke type picker

radpet:


radpet:
@rsinghal  this is https://github.com/IIIF/mirador/pull/840 with some modifications

js/src/widgets/contextControls.js
```diff
@@ -168,6 +200,15 @@
                                    '<a class="hud-control draw-tool">',
                                    '<input type="text" class="fillColorPicker"/>',
                                    '</a>',
+                                   '{{#if showStrokeStyle}}',
+                                   '<a href="javascript:;" class="mirador-btn draw-tool mirador-line-type" role="button" aria-label="Change Line Type"><i class="material-icons solid">create</i>',
+                                   '<ul class="dropdown type-list">',
+                                   '<li><i class="fa solid"></i> solid</li>',
```
rsinghal: The text used in these `<li>` elements must use i18next and the translation files

css/mirador.css
```diff
@@ -1231,6 +1231,52 @@ li.highlight {
...
+  width: 55px;
+  height: 10px;
+  font-size: 15px !important;
+  background-repeat: no-repeat;
+}
+/*
+ TODO find font-icons for border types
+*/
+.mirador-line-type .solid {
+  background-image: url('../images/border_type_1.png');
```
rsinghal: mirador.css can't use `url('../images` because the location of the `images` folder can be set by `buildPath` and `imagesPath`.  You need to do something like this: https://github.com/IIIF/mirador/blob/release2.1/js/src/viewer.js#L60 to get those paths and set the `background-image` in the JS, not the CSS

rsinghal:
I made some additional comments in #840 about these two line notes.

css/mirador.css
radpet: Yup I saw this in the old pr,however missed coding it.

radpet:
@rsinghal What if I make all pickers dropdowns? I have changed them to open on hover but I am not really keen with that, they close if I accidentally hover something else.

rsinghal:
@radpet I think that seems reasonable and will hopefully remove any frustrations of having the dropdown disappear if they move the mouse too quickly.  Personally, I would like to remove the `Choose` and `Cancel` buttons so that once a user clicks on a color, that chooses the color and hides the dropdown (which is how the border type dropdown would also work). And if a user re-clicks on the icon, that effectively cancels it.

radpet:
@rsinghal Yup I totally agree with you, however I am planning to do this as part of another pr because this is part of https://github.com/IIIF/mirador/issues/882. With this pr I just want to add the missing border/stroke/line type functionality.

rsinghal:
@radpet sounds good.  I will review this PR again (with your updates) and merge or let you know if I find any other issues.

rsinghal:
@radpet @beaudet @aeschylus 

I wonder if the lines need to be thicker by default?  It's really hard to see the difference between the different types of borders (one is dashed, one is solid):

screen shot 2016-07-06 at 2 52 51 pm

screen shot 2016-07-06 at 2 53 56 pm

screen shot 2016-07-06 at 2 54 24 pm

locales/en/translation.json
```diff
@@ -40,8 +40,15 @@
     "fullScreenWindowTooltip": "Toggle this window to full screen",
     "logo": "Logo",
     "load": "Load",
+    "load22": "Load",
```
radpet: I should remove this sorry

radpet:
I agree I can make another pr with additional setting coming from the config file if someone wants to change the size of the dashes.

PR #1005, opened 2016-07-06
789 slider styling improvements

aeschylus:
Styles sliders on image manipulation controls so they appear underneath the caret of their respective icons. Displays the percentage represented by the slider to the right of the handle. Updates on reset.


aeschylus:
Hm. I need to write some tests for this.

rsinghal:
Yes, we need to write tests, but we changed different widgets, which will get significantly modified by iiifManifestLayouts, which is why we haven't added tests yet.

PR #1014, opened 2016-07-13
Rotation Icons/Refactoring/Fix minor bugs

radpet:
@rsinghal

I change the rotation circle to icon. The icon here is temporary until we agree on which one to use.When we choose one I will change it.

I fixed some bugs related to selecting elements. Currently, before those fixes you were able to select sometimes more than one shape at a time which was rather buggy.
 
I refactored all unit tests. Most of them still need more refactoring because they depend on other tools to test themselves which is not right. I included a dummy tool but its not used anywhere. I'll refactor them step by step with the new functionality.

After we merge this I am thinking to start implementing 2.iii.a.b.a for which I will also need a proper trashcan icon.

 For now they will be hardcoded because there is some math behind them, but in future I'll try to make them to come as configuration (both rotation and deletion icons)


rsinghal:
Until you are ready to completely remove awesome cursor from the code, can you comment out using the new icon since it is showing both right now.
http://jsfiddle.net/rna9qv6m/ is another alternative for the icon

radpet:
@rsinghal  I changed the the rotation handle to become an image. I did not changed any cursor icons yet.

What do you mean by 

> it is showing both right now.

rsinghal:
image

rsinghal:
Also, the length of the line above the rectangle (for the rotation handle) and the new rotation icon do not properly scale as you zoom in and out.  If you start off fully zoomed in and click on a shape and then zoom fully out or vice versa, you will see the issue.

radpet:
@rsinghal I saw that issue with zoom in/out Ill try to fix it. What do you suggest about the rotation cursor?

rsinghal:
For now, let's just keep using awesome cursor until we can think through this some more.  Since you also need to add the trash can icon with the shape on edit, the rotation icon can be revisited then.

radpet:
@rsinghal  Okey Ill comment in the the rotate icon which is attached to the shape?

rsinghal:
Yes, sounds good.

radpet:
@rsinghal I'll be ready in few minutes and then start working on the trash can :+1:

radpet:
@rsinghal It should be same as before + some fixes :) . Would you please give me some info from where I can get the rotation handle icon and trash can images.

PR #1020, opened 2016-07-19
Display delete button when shape is selected

radpet:
Use PointText to display trash icon
Use Raster as mask to the trash icon to make it clickable because we don't hitTest for 'fill'

@rsinghal 
I have added a property in the settings which control the color of selected shape.
I have made the pin to be dragged / fixes https://github.com/IIIF/mirador/issues/982

I have implemented the showing of the delete button of the vicinity of each shape when it is selected.
The icon auto resizes on zoom in/zoom out. 
The icon should handle all shape transformation and preserve its location where it was initially attached.

(What will be done in another pr):
Cursor changes when you mouse over icon.
Change the handle for rotation to icon.

### DIFF
--- a/js/src/annotations/annotation-utils.js
+++ b/js/src/annotations/annotation-utils.js
@@ -4,23 +4,21 @@
 
   };
 
-  var Icon = function(paperScope,opts){
-    this.paperScope = paperScope;
-    this.raster = new paperScope.Raster(opts);
-    if(opts && opts.onLoad){
-      this.raster.onLoad = opts.onLoad;
-    }
-  };
+  function PaperItem(item){
+    this.item = item;
+  }
 
-  Icon.prototype = {
-
+  PaperItem.prototype  = {
+    getItem:function(){
+      return this.item;
+    },
     translateByXY: function(x,y){
-      this.raster.position.x +=x;
-      this.raster.position.y +=y;
+      this.item.position.x +=x;
+      this.item.position.y +=y;
     },
 
     translateByPoint:function(point){
-      this.raster.position = this.raster.position.add(point);
+      this.item.position = this.item.position.add(point);
     },
 
     setOnMouseDownListener:function(callback){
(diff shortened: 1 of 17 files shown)


radpet:
@rsinghal I found that I have introduced bug when you select shape for transformation. Will need some time to find what causes it. I'll notify you when it is ready.

radpet:
@rsinghal I'll send another pr that fixes the clicking on the icon and the delete action itself because you can't do it in edit mode now (if shape is enabled for editing its part of shapes should also be :) )

rsinghal:
@radpet

I am seeing what you are describing where a user has to click on a shape twice when in edit mode.  Basically, on the very first time, if you click and drag, it works.  But if you release the mouse and try to click on one of the handles or on the border to move it, it doesn't work.  You have to release the mouse and then click/drag again.  It seems to be losing the selected state.  

This is not happening in the current `annotation-work` branch, so something in your PR seems to have changed this behavior.  I will attempt to figure out how I fixed that from my commits.

radpet:
@rsinghal I pushed some changes, when did you check it out? I am testing it now and it works fine. The only strange thing I found is when editing polygon and I'll checking it now

rsinghal:
Yes, I think I was using an old build, and it's working correctly for me now.  I only fixed ellipse and rectangle when in edit mode because I think you mentioned you had some refactoring you needed to do for polygon, so the selected state is not working properly for polygons.

radpet:
@rsinghal Ahh yes that was the problem I suppose thanks a lot. Well if you are okey with the icon lets merge this.

rsinghal:
@radpet Can you make the trash can icon white?  

I also find that when I edit a shape that is drawn near the bottom of the image I can't see the trashcan.  Or, if you draw a shape near the top of the image, you can't access the rotate point.  Lastly, the top point (used for rotation) does seem to be adjusting size when the user zooms in/out.  I'm going to create a ticket for all of this separately so as not to slow down this PR, but I wanted you to be aware of it (I will also add screenshots)

radpet:
You don't see it when its near the bottom maybe because it is black, what do you think about being in the same color as the selected shape.

The blue doth will be replaced with icon as you know and will act similarly as the trash can.

rsinghal:
Sure, it can be the same color as the border.  

As you can see from the two screenshots, the rotate point and trash can are not visible if they are too close to the edge.  

screen shot 2016-07-19 at 4 04 41 pm
screen shot 2016-07-19 at 3 54 10 pm

rsinghal:
@radpet I am getting an error after trying to zoom/pan while in edit mode and also after saving and being in normal mode:

`item.data.self.setPosition(item.data.self.getData('pivot').add(new overlay.paperScope.Point(0, 21 / overlay.paperScope.view.zoom).rotate(item.data.parent.data.rotation)));`

Specifically, the `data` of `item.data.parent` is undefined.

radpet:
Ahh yes you got me here I didn't check this use case.

Hmm strange it should not be undefined which tool were you selected?

rsinghal:
Ellipse

radpet:
@rsinghal  Ahh a small typo, I do really sorry. It should be okey now.

PR #1026, opened 2016-07-25
Deletes shape when clicked on the trash can icon.

radpet:
@rsinghal 
## This pr fixes:

Implement single shape delete action.
Deletes shape action is triggered when you click the trash can icon.
Confirmation message appears.

Add check if you try to create annotation without any shapes.

Add check if you edit annotation and you have deleted all shapes.
In case you have no shapes and you try to save it triggers confirmation that the annotation will be deleted.

Unsubscribe from all events in the osd overlay when the osd is closed (This should fix some large memory leaks)

## What will be fixed in future pr: (DONE (waiting to merge this pr in order to rebase on top of it and I'll send pr) )
@rsinghal @beaudet
All confirmation dialogs/popups will be modal implemented with the help of [http://bootboxjs.com/#](http://bootboxjs.com/#)
All messages displayed will be configured from the i18n langs

Refactor overlay to keep track of subscribed events in order to unsubscribe from all of them when needed (write unit test to watch for event subscription memory leaks)

### DIFF
--- a/js/src/annotations/osd-svg-overlay.js
+++ b/js/src/annotations/osd-svg-overlay.js
@@ -3,6 +3,7 @@
     if (this.svgOverlayTools) {
       return this.svgOverlayTools;
     }
+    options.partOfPrefix = '_partOf';
     this.svgOverlayTools = [new $.Rectangle(options), new $.Freehand(options), new $.Polygon(options), new $.Ellipse(options), new $.Pin(options)];
     return this.svgOverlayTools;
   };
@@ -35,6 +36,7 @@
       fixedShapeSize: drawingToolsSettings.fixedShapeSize,
       selectedColor: drawingToolsSettings.selectedColor || '#004c66',
       shapeHandleSize:drawingToolsSettings.shapeHandleSize,
+      partOfPrefix:'_partOf',
       hitOptions: {
         handles: true,
         stroke: true,
@@ -43,7 +45,6 @@
       }
     });
 
-    // Initialization of overlay object.
     this.tools = $.getTools(drawingToolsSettings);
     this.currentTool = null;
     // Default colors.
@@ -71,6 +72,12 @@
     this.viewer.addHandler('resize', _thisResize);
     this.viewer.addHandler('rotate', _thisResize);
     this.viewer.addHandler('constrain', _thisResize);
+
+    var _thisDestroy = function(){
+      _this.destroy();
+    };
+
+    this.viewer.addHandler('close',_thisDestroy);
 
 
     _this.eventEmitter.subscribe('toggleDrawingTool.' + _this.windowId, function(event, tool) {
@@ -107,6 +114,14 @@
           break;
         }
       }
+    });
+
+    _this.eventEmitter.subscribe('deleteShape.' + _this.windowId, function (event, shape) {
+      // TODO use global notification system
+      if (window.confirm("Are you sure you want to delete the shape?")) {
+        _this.deleteShape(shape);
+      }
+
     });
 
     _this.eventEmitter.subscribe('changeBorderColor.' + _this.windowId, function(event, color) {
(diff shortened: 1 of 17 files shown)


js/src/annotations/osd-svg-overlay.js
```diff
@@ -761,6 +796,28 @@
...
+    // Should have no references in order to be garbage collected
+    destroy:function(){
+      this.eventEmitter.unsubscribe('deleteShape.' + this.windowId);
+      this.eventEmitter.unsubscribe('toggleDrawingTool.' + this.windowId);
+      this.eventEmitter.unsubscribe('toggleDefaultDrawingTool.' + this.windowId);
+      this.eventEmitter.unsubscribe('deleteShape.' + this.windowId);
+      this.eventEmitter.unsubscribe('changeBorderColor.' + this.windowId);
+      this.eventEmitter.unsubscribe('changeFillColor.' + this.windowId);
+      this.eventEmitter.unsubscribe('changeBorderColor.' + this.windowId);
+      this.eventEmitter.unsubscribe(' toggleBorderType.' + this.windowId);
```
rsinghal: I think there is a typo here with the extra space before `toggleBorderType`

js/src/annotations/osd-svg-rectangle.js
```diff
@@ -1,13 +1,11 @@
 (function($) {
   $.Rectangle = function(options) {
-    jQuery.extend(this, {
+    jQuery.extend(this, options ,{
```
rsinghal: In the other shape files, `options` is the last parameter for `extend`, but in this file you moved it

rsinghal:
I like the work you did with being able to delete an individual shape.  However, I don't agree that saving an annotation that has all shapes deleted should, by default, delete the annotation.  The dialog that pops up could be very confusing to the user, since it is saying that saving causes deletion.  Instead, the dialog should present the user with the option to save the annotation with no shapes or delete the entire annotation, and act accordingly.

The option to have an annotation with no shapes is a valid use case in terms of an annotation and there have been discussions to add other ways to display annotations in Mirador (such as through the left side panel).  This means that even though an annotation with no shapes won't show up in Mirador right now, eventually there will be a mechanism to display those annotations.

Regarding Bootbox - that requires Bootstrap, which we don't use in Mirador.  What do we gain by adding these to Mirador (since it does make the JS and CSS files larger)?  Is there something smaller we can use?

beaudet:
Clearly, M2.1 needs to support the use cases that have been identified and I don't see any mutually exclusive use cases, so I think your suggestions for alternative behavior would be compatible with ConservationSpace R2's requirements.  Radoslav, please let me know if I'm missing anything here.

radpet:
@rsinghal @beaudet 

I agree with you. I'll remove the check and the user will be able to save annotation when there are no shapes.

As far as Bootbox is concerned, I haven't search much for additional libs. However, how about changing the default annotation tooltip to http://v4-alpha.getbootstrap.com/components/popovers/?

About Bootstrap I have done several optimizations:
The css file is rather small because I manually selected only the modal styles. I think I we can use [https://github.com/twbs/bootstrap/blob/master/js/modal.js](https://github.com/twbs/bootstrap/blob/master/js/modal.js) not the whole min.js they provide.


Another libs for creating modals I think we might use are [https://github.com/noelboss/featherlight/](https://github.com/noelboss/featherlight/) or [https://github.com/hakimel/Avgrund](https://github.com/hakimel/Avgrund)

radpet:
@rsinghal There are few more pr I have prepared on top of this one that I want to merge to annotation-work as soon as possible? If you have any objections we will discuss them and I'll fix them as separate pr if necessary :+1:

rsinghal:
I will create a new ticket for saving annotations without shapes.

I'm not seeing the reason to add a new modal library to Mirador.  The simple window dialog boxes seem to be good enough at the moment, no?

radpet:
screenshot from 2016-07-27 19-56-53

Don't you like it?

rsinghal:
Yes - it does look nicer - I agree :)

PR #1029, opened 2016-07-29
Change delete confirms to use custom dialogs

radpet:
@rsinghal @beaudet
## This pull request adds
Add dialog builder that use bootbox lib
Refactor all native confirm dialogs to use the dialog builder
Add translations
Fix more memory leaks
Add test to check the unsubscription when overlay is about to be destroyed

## What I want to disscuss
What text should be shown when the user tries to save annotation without shapes.

## What will be done in other pull request
Make the save without shapes option available as part of https://github.com/IIIF/mirador/issues/1028


js/lib/bootbox.js
```diff
@@ -0,0 +1,985 @@
+/**
```
radpet: I think that we should minify/uglify all libs on the build phrase as I proposed here:
https://github.com/IIIF/mirador/issues/1027
not in the lib folder, because it is easier to debug

rsinghal: Agreed.  As he noted, @aeschylus is changing the build process, so he will incorporate that with the rest of his work.

rsinghal:
A few notes on the Text/translation file:

You can reuse the `cancel` key in the translation file, rather than creating another one.  

I don't think you need a title (`editModalSaveAnnotationWithNoShapesTitle`).  

And I would change `There are no shapes related to this annotation. What shall we do?` to `There are no shapes related to this annotation.` since the buttons explain their choices anyway.  @aeschylus @beaudet any thoughts on the text presented to the user?

beaudet:
I don't have strong feelings about this.  So long as it's clear to the user what is happening, what their options are, and the language that is selected is consistent with "Mirador vernacular", I think it will be fine.

radpet:
@rsinghal Shall we merge this if you are okey with it, I look forward to create new pull request with a lot of new things.

rsinghal:
Yup, looks good.

Issue #1028, opened 2016-07-29
Gracefully ignore annotations without fragment or svg selectors


rsinghal:
For now, that means not attempting to create a shape since there is nothing to draw.

@radpet

radpet:
@rsinghal This should work now in https://github.com/IIIF/mirador/pull/1030

PR #1030, opened 2016-08-01
Refactor tools to handle new edit workflow

radpet:
@rsinghal @beaudet
# This pull request adds :

## Main feature

Refactor pin tool / Refactor polygon tool / Refactor freehand tool to handle to be visually highlighted in edit mode

Fix deletion of only one point (freehand and polygon tool) https://github.com/IIIF/mirador/issues/986

Disable shape selecting when out of edit mode or create new annotation mode

## Additional features
Add stroke width change on newly created shapes

@rsinghal I have commented some code in order to review it and tell me if you agree on removing it 

# What should be done in other pull requests
Have appropriate cursor change (remove awsome cursor lib, and use only native) [generalize the solution for all shapes]

Change blue doth (for rotation) to rotate icon ( Which one do you suggest to use. I like the one in the mock ups but I cannot find it)

Refactor the whole osd-overlay test suite (after we all agree on the features and workings of the annotations)
Check out the whole annotation functionality for code/memory improvements (should have good coverage and really working tests first (not only blind covering) (we have them almost) )

---

With this pull request I think that https://github.com/IIIF/mirador/issues/881 is rather done (except the rotate icon and the cursors changes?

@rsinghal @beaudet Please share with me if I am missing something or you suggest any improvements.

---
### Found issues that will be fixed as part of this pull request:

- [x] https://github.com/IIIF/mirador/issues/1028 
- [x] https://github.com/IIIF/mirador/issues/1031
- [x] 
> Have an existing annotation for the image.
On that image, create an annotation wth one shape (in this case, I chose polygon) and save it.
Edit that new annotation and delete the shape.
Click Save and choose to delete the annotation in the modal (I can't test this with Saving the annotation without shapes because of the first bug).
Click to edit the existing annotation
Click anywhere on the image (not on one of the existing shapes), it starts drawing the points of a polygon, rather than doing nothing.

- [x] Styles for newly created shapes is not preserved when going to fullscreen mode

### DIFF
--- a/js/src/annotations/osd-region-draw-tool.js
+++ b/js/src/annotations/osd-region-draw-tool.js
@@ -96,7 +96,7 @@
           var onAnnotationSaved = jQuery.Deferred();
 
           if (!_this.svgOverlay.draftPaths.length) {
-            new $.DialogBuilder().dialog({
+            new $.DialogBuilder(_this.svgOverlay.slotWindowElement).dialog({
               message: i18n.t('editModalSaveAnnotationWithNoShapesMsg'),
               buttons: {
                 success: {
--- a/js/src/annotations/osd-region-draw-tool.js
+++ b/js/src/annotations/osd-region-draw-tool.js
@@ -57,11 +57,13 @@
       this.svgOverlay.paperScope.project.clear();
       var _this = this;
       _this.annotationsToShapesMap = {};
-      var deferreds = jQuery.map(this.list, function(annotation) { // TODO should refactor to handle annotation without shapes
+      var deferreds = jQuery.map(this.list, function(annotation) {
         var deferred = jQuery.Deferred();
-        var shapeArray;
-        if (typeof annotation.on === 'object') {
-          if (annotation.on.selector.value.indexOf('<svg') != -1) {
+        if(annotation.on && !annotation.on.selector){
+          console.log('no shape attached on this annotation',annotation); // temporary until view for full list of annotations for current canvas is contributed
+          return deferred;
+        } else if (typeof annotation.on === 'object') {
+          if (annotation.on.selector.value.indexOf('<svg') !== -1) {
             shapeArray = _this.svgOverlay.parseSVG(annotation.on.selector.value, annotation);
           } else {
             var shape = annotation.on.selector.value.split('=')[1].split(',');
@@ -96,14 +98,20 @@
           var onAnnotationSaved = jQuery.Deferred();
 
           if (!_this.svgOverlay.draftPaths.length) {
-            new $.DialogBuilder(_this.svgOverlay.slotWindowElement).dialog({
+            new $.DialogBuilder(windowElement).dialog({
               message: i18n.t('editModalSaveAnnotationWithNoShapesMsg'),
               buttons: {
                 success: {
                   label: i18n.t('editModalBtnSaveWithoutShapes'),
-                  className: 'btn-success disabled',
+                  className: 'btn-success',
                   callback: function () {
-                    // TODO currently we do not support this option
+                    oaAnno.on = {
+                      "@type": "oa:SpecificResource",
+                      "full": _this.state.getWindowObjectById(_this.windowId).canvasID
+                    };
+                    //save to endpoint
+                    _this.eventEmitter.publish('annotationUpdated.' + _this.windowId, [oaAnno]);
+                    onAnnotationSaved.resolve();
                   }
                 },
                 danger: {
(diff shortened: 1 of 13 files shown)


js/src/annotations/osd-svg-overlay.js
```diff
@@ -303,60 +307,80 @@
...
-                  inArray = true;
-                  break;
-                }
-              }
-              if (!inArray) {
-                this.overlay.editedPaths.push(this.overlay.path);
-              }
-            }
-          }
+          // should check if this is used anywhere and remove it if not used
```
radpet: Used in
https://github.com/IIIF/mirador/blob/0db6f1fdd26a13d3e66a108ba924ad68667a3561/js/src/annotations/osd-region-draw-tool.js#L82 

I don't think we need it with the new workflow, do we?

rsinghal:
@radpet I found the following bugs:

-----

1. Create an annotation with one shape and save it.
2. Edit that annotation and delete the shape.
3. Save the annotation and choose `Save annotation without shapes`
4. The modal doesn't close but there are no errors in the console.

-----

1. Have an existing annotation for the image.
2. On that image, create an annotation wth one shape (in this case, I chose polygon) and save it.
3. Edit that new annotation and delete the shape.
4. Click `Save` and choose to delete the annotation in the modal (I can't test this with Saving the annotation without shapes because of the first bug).
5. Click to edit the existing annotation
6. Click anywhere on the image (not on one of the existing shapes), it starts drawing the points of a polygon, rather than doing nothing.

-----

I am still reviewing and testing, so I may have more (as well as determining if the commented out code is still needed).

js/src/annotations/osd-svg-overlay.js
rsinghal: I don't think so.  I didn't write that - I think that was already there when I started adding the new edit code, so if it doesn't appear to be used anywhere, I think it can go away.

radpet:
@rsinghal Thanks for you response and work:

About

> Create an annotation with one shape and save it.
Edit that annotation and delete the shape.
Save the annotation and choose Save annotation without shapes
The modal doesn't close but there are no errors in the console.

I made the save annotation without shapes button disabled until https://github.com/IIIF/mirador/issues/1028 is implemented

rsinghal:
@radpet I see - perhaps hide that button entirely until #1028 is implemented, and then we can make the button visible again?  In case anyone is testing/using this branch, I don't want to get unnecessary tickets about it not working (since we already know it's not implemented yet!)

radpet:
@rsinghal That is why I made it disabled I'll try to implement it as part of this pr.

Anyway about 

> Have an existing annotation for the image.
On that image, create an annotation wth one shape (in this case, I chose polygon) and save it.
Edit that new annotation and delete the shape.
Click Save and choose to delete the annotation in the modal (I can't test this with Saving the annotation without shapes because of the first bug).
Click to edit the existing annotation
Click anywhere on the image (not on one of the existing shapes), it starts drawing the points of a polygon, rather than doing nothing.

you are right. The current tool need to be reset but it is not now. Ty :)

js/src/annotations/osd-svg-overlay.js
radpet: It is rather old code from the first version of the annotation-tools. I think it was used when we used the save button icon in order to save the edited shapes

radpet:
@rsinghal Please check out the new things, I tried to fix all issues I mentioned above. I want to merge this as soon as possible because we want to trigger new build for the CS project from annotation-work :)

rsinghal:
Reviewing now

js/src/annotations/osd-svg-overlay.js
rsinghal: Makes sense - I think it can be removed

rsinghal:
@radpet I found a bug creating an annotation with multiple slots open.  I'm going to fix that in `annotation-work` now

js/src/annotations/osd-region-draw-tool.js
```diff
@@ -55,11 +57,13 @@
...
       var _this = this;
       _this.annotationsToShapesMap = {};
-      var deferreds = jQuery.map(this.list, function(annotation) { // TODO should refactor to handle annotation without shapes
+      var deferreds = jQuery.map(this.list, function(annotation) {
         var deferred = jQuery.Deferred();
-        var shapeArray;
-        if (typeof annotation.on === 'object') {
-          if (annotation.on.selector.value.indexOf('<svg') != -1) {
+        if(annotation.on && !annotation.on.selector){
+          console.log('no shape attached on this annotation',annotation); // temporary until view for full list of annotations for current canvas is contributed
```
rsinghal: Can you remove this log statement in your next PR?

radpet: Yup no problem :)

PR #1040, opened 2016-08-09
remove postinstall script

wkentdag:
fixes https://github.com/IIIF/mirador/issues/821#issuecomment-238626079, allowing users to install mirador via `npm` like so: `npm install --save IIIF/mirador#release2.1`

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,6 @@
   "scripts": {
     "start": "bower install && grunt",
     "test": "./node_modules/grunt-cli/bin/grunt cover jshint coveralls --verbose --full",
-    "postinstall": "grunt default",
     "travis": "./node_modules/grunt-cli/bin/grunt ci --verbose --force"
   },
   "dependencies": {}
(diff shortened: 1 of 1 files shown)


aeschylus:
Okay, we can remove this for now. The build process is about to be overhauled, so it makes sense to let people include it without the postinstall. I think the task is more trouble than it's worth even when it does work as intended.

wkentdag:
@aeschylus - whoops, I should've left a comment before you merged this. As it stands, running `npm install --save IIIF/mirador#release2.1` is not enough to actually manage this package via npm - the dependencies still need to be installed and `grunt` needs to run before you can access the source in the `/build` folder. I tried adding `npm start` as a new `postinstall` script on my fork, but it keeps failing because the bower dependency `qTip2#~2.2.1` is getting a 404. I'm not familiar enough with this codebase to modify those dependencies, and maybe it doesn't matter if you're already overhauling the build process. Just thought I should leave a note though for anybody else trying to use this with npm to follow these steps for a complete build:

```
$ npm i -S IIIF/mirador#release2.1
cd ./node_modules/Mirador
npm install && npm start
```

🎉

aeschylus:
Okay, see the revert above. Can you edit it and make another PR adding these back. The alternative is to put it in the documentation.

wkentdag:
sure thing, I'll do both. will try to get it in by tomorrow

PR #1043, opened 2016-08-11
Feature/integration improvements

radpet:
# This pull request fixes:

Memory leaks caused by not unsubscribing from events when object is not used any more
Handle refresh annotation action properly

# This pull request adds:

Events that handle all annotation logic inside overlay in order to remove them from the annotation-tooltip object

Option whether to proceed to confirmation screen after save beeing clicked when you edit annotation and have no shapes.

### DIFF
--- a/js/src/annotations/annotationTooltip.js
+++ b/js/src/annotations/annotationTooltip.js
@@ -1,7 +1,6 @@
 (function($) {
 
   $.AnnotationTooltip = function(options) {
-
     jQuery.extend(this, {
       targetElement: null,
       annotations: [],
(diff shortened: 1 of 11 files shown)


js/src/annotations/annotationTooltip.js
```diff
@@ -82,22 +81,11 @@
...
 
-              var returnToPointer = function(){
-                _this.eventEmitter.publish('SET_STATE_MACHINE_POINTER.' + _this.windowId);
+              var cancelCallback = function(){
                 api.destroy();
-                if (params.onCancel) { params.onCancel(); }
               };
 
-              if (_this.activeEditor.isDirty()) {
-                new $.DialogBuilder().confirm(i18n.t('cancelAnnotation'),function(result){
```
rsinghal: Looks like you are not passing the isDirty check in the event publisher so it's showing the dialog on an empty text editor.

rsinghal:
@radpet I noticed (and I don't know if this happened in this PR or a previous one), but when you go into Edit mode for an annotation and click on one of the shapes and then click outside of it, it draws that shape.  It should only draw a shape when the user explicitly selects one of the shapes by the top icons.  A user should be able to click around the rest of the image when in pointer mode without drawing a shape.  I believe this was working at one point, so I don't know when it broke.

radpet:
@rsinghal 

> @radpet I noticed (and I don't know if this happened in this PR or a previous one), but when you go into Edit mode for an annotation and click on one of the shapes and then click outside of it, it draws that shape. It should only draw a shape when the user explicitly selects one of the shapes by the top icons. A user should be able to click around the rest of the image when in pointer mode without drawing a shape. I believe this was working at one point, so I don't know when it broke.

Nah this happens in annotation-work too... I'll try to fix it in another pull request.

js/src/annotations/annotationTooltip.js
radpet: Thanks :)

rsinghal:
We need that bug fixed as soon as possible as we will be doing a code freeze early next week.

radpet:
Right I will start fixing it right now, meanwhile please checkout this pull request.

PR #1048, opened 2016-08-12
Reset current tool after each click

radpet:
@rsinghal 

Create one shape per tool selection. Reset current tool on every mousedown except when creating shape.

This is what we talked in https://github.com/IIIF/mirador/pull/1043.
I think this pull request fixes the issue.Check it out please and tell me if this is what we are looking for :)


js/src/annotations/annotation-utils.js
```diff
@@ -158,6 +158,7 @@
   DeleteActionIcon.prototype.setOnMouseDownListener = function (overlay) {
     this.mouseDown = function () {
       overlay.eventEmitter.publish('deleteShape.' + overlay.windowId, [this.getData('parent')]);
+      overlay.mode = 'delete';
```
radpet: set mode to sth different from ''

PR #1079, opened 2016-08-26
whenever window is initialized or updated, make sure we unsubscribe f…

### LINKED ISSUE #1075: Replace Object not properly destroying old events

rsinghal:
…rom any events in odd-region-draw-tool and osd-svg-overlay, since they will be re-subscribed
closes #1075

### DIFF
--- a/js/src/annotations/osd-svg-overlay.js
+++ b/js/src/annotations/osd-svg-overlay.js
@@ -68,261 +68,6 @@
     this.eventEmitter = eventEmitter;
     this.eventsSubscriptions = [];
 
-    this._thisResize = function(){
-      _this.resize();
-    };
-    this._thisRotate = function(event) {
-      _this.rotate(event);
-      _this.resize();
-    };
-    this.viewer.addHandler('animation', this._thisResize);
-    this.viewer.addHandler('open', this._thisResize);
-    this.viewer.addHandler('animation-finish', this._thisResize);
-    this.viewer.addHandler('update-viewport', this._thisResize);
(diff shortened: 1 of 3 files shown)


js/src/annotations/osd-svg-overlay.js
```diff
@@ -386,6 +131,265 @@
...
+      this.viewer.addHandler('rotate', this._thisRotate);
+      this.viewer.addHandler('constrain',this._thisResize);
+
+      this._thisDestroy = function(){
+        _this.destroy();
+      };
+
+      this.viewer.addHandler('close',this._thisDestroy);
+
+      this.eventsSubscriptions.push(this.eventEmitter.subscribe('DESTROY_EVENTS.'+this.windowId, function(event) {
```
radpet: Do you unsubscribe for this event in the destroy()?

rsinghal: Nothing needs to change in the destroy function.  window.js sends out a message to unsubscribe when it is updated (which is what happens when a user selects Replace Object)

radpet: Why don't you unsubscribe from DESTROY_EVENTS ? It gets subscribed again on the next initialization,I don't see need to keep calling it more than once, aren't we only leaking memory?

rsinghal: I don't understand.  Since it's part of the `eventsSubscriptions` array, it automatically gets unsubscribed on the next initialization (as part of `destroy), and then gets re-subscribed

PR #1082, opened 2016-08-28
modeChange event should only be triggered from hud and only listened …

### LINKED ISSUE #1080: "displayOn inappropriate in current state pointer" error when changing to annotated pages
Using the NAF 6221 manifest that ships with the demo:

1. Turn to a page containing annotations (e.g. 1r)
2. Toggle off the annotations.
3. Turn to another page containing annotations (e.g. 1v)

Notice that although the annotation controls have been automatically toggled on, no annotations actually display. In F12, "displayOn inappropriate in current state pointer" appears in the error logs.

This has never happened with non-RC versions of 2.1 before, so this is definitely a recent change.

rsinghal:
…to in annotationsLayer.  refactoring in annotationsLayer and osd-region-draw-tool.  selecting the pointer tool should not automatically cancel an active annotation.

closes #1080 

@radpet Can you check that everything is working as you expect on your end before I merge?

### DIFF
--- a/js/src/annotations/osd-region-draw-tool.js
+++ b/js/src/annotations/osd-region-draw-tool.js
@@ -24,24 +24,42 @@
       this.listenForActions();
     },
 
-    enterCreateMode: function() {
-      this.osdViewer.setMouseNavEnabled(false);
-      this.svgOverlay.show();
-      this.svgOverlay.enable();
+    enterDisplayAnnotations: function() {
+      // if a user selected the pointer mode but is still actively
+      // working on an annotation, don't re-render
+      if (!this.svgOverlay.inEditOrCreateMode) {
+        this.exitEditMode(true);
+        this.render();
+      }
     },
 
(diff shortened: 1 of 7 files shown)


js/src/annotations/osd-region-draw-tool.js
```diff
@@ -184,16 +201,6 @@
         _this.destroy();
       }));
 
-      this.eventsSubscriptions.push(_this.eventEmitter.subscribe('refreshOverlay.' + _this.windowId, function(event) {
```
radpet: We might need this how we re-render the annotations when refresh is clicked?

rsinghal: Yes, but what happens inside of this will likely be quite different, so it's better to re-code it when it's added rather than be confused by outdated code.

radpet: Do you plan to add it soon?

PR #1092, opened 2016-08-29
Add "within" property to "on" in the json-ld annotation

regisrob:
... in the same way as the Search API does: http://iiif.io/api/search/1.0/#target-resource-structure
And as per this thread on iiif-discuss: https://groups.google.com/forum/#!topic/iiif-discuss/HZtInSSs_8k


aeschylus:
👍

PR #1107, opened 2016-09-02
Add support for multiple annotation list within manifest.

xtai:
## About this PR

This PR adds support for multiple annotation lists within a single canvas.

[According to IIIF Presentation API](http://iiif.io/api/presentation/2.1/#canvas), it seems reasonable to support multiple manifest annotation lists:
> The value of both of these properties must be a list, even if there is _only one entry_.

This PR adds all annotations from different lists (urls) to the same single array: `_this.annotationsList`; therefore, Mirador can display all annotations on window load.

## Why?

We have been working on improve Mirador annotation to utilize as part of highlighting search results/newspaper article feature.

It can be really useful for us to have different types of annotation lists/layers and to display them without a annotation backend server.


xtai:
Sorry, delete the PR branch by accident.

rsinghal:
@aeschylus Why is Coveralls saying coverage decreased by 3.9%?  I don't see enough changes to warrant such a massive decrease

aeschylus:
Basically this created more cases/logic branches we need to test for, since it's now possible for the array to be empty or have multiple items.

xtai:
Hi @aeschylus, I did add tests for multiple items in an array and empty array in the test file: [spec/manifests/manifest.test.js Line 243](https://github.com/IIIF/mirador/pull/1107/files#diff-b99c04de63ade6db71e6f8bfd42f109bR243) and [spec/manifests/manifest.test.js Line 293](https://github.com/IIIF/mirador/pull/1107/files#diff-b99c04de63ade6db71e6f8bfd42f109bR293).

rsinghal:
@xtai One of the tests is failing: https://travis-ci.org/IIIF/mirador/builds/157137900#L1125.  Can you fix before we merge?

xtai:
@rsinghal Done, I guess the failing test was the reason behind of the massive -3.9% coverage decrease.

xtai:
@rsinghal Fixed.

aeschylus:
Figured out what was wrong with the other coverage/failing tests, but this one makes no sense. Merging.

PR #1109, opened 2016-09-08
Unsupported Browser Detection

thehabes:
This will detect whether or not the browser is a supported one at the window level.  Then the hide/show flags for certain elements can be overwritten based on the true/false value of the parameter set and tracked by the init() (unsupportedBrowser).

### DIFF
--- a/js/src/widgets/hud.js
+++ b/js/src/widgets/hud.js
@@ -17,9 +17,17 @@
 
     init: function() {
       this.createStateMachines();
-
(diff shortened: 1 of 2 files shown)


aeschylus:
Currently working on the related issue, so I appreciate the PR. However, it looks like this causes all the tests to fail.

aeschylus:
Also, works correctly because it implements the standard:
https://saucelabs.com/beta/tests/8d417dcc975d4f7196a43d1ebe99034b/watch

PR #1116, opened 2016-09-14
Fixed missing background on manifest loading screen

dicksonlaw583:
This pull request restores the background colour lost in commit [5b0eb0b](https://github.com/IIIF/mirador/commit/5b0eb0beee02b30ceef2ec87b85062d0c8f750a9). Both Chrome and Firefox treated the `//` comment as invalid CSS and ignored the two properties below it.

### DIFF
--- a/css/mirador.css
+++ b/css/mirador.css
@@ -233,7 +233,7 @@ a {
   top: 0;
   display: none;
   overflow: hidden;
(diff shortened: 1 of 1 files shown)


aeschylus:
This was also fixed in: https://github.com/IIIF/mirador/pull/1114, but given the status of that PR, it's good to merge this separately.

Thanks @dicksonlaw583 .

PR #1129, opened 2016-09-20
Update french locale

regisrob:


rsinghal:
@regisrob Can you switch this to `2.1.1`?

PR #1133, opened 2016-09-22
Modify JsonLd.getTextValue to match algorithm outlined in spec

jbaiter:
This PR implements the algorithm outlined in ["4.3. Language of Property Values"](http://iiif.io/api/presentation/2.1/#language-of-property-values) in the Presentation API 2.1 specification.

It maintains complete backwards compatibility with existing code (as far as I could tell from the unit tests).

In the present implementation the browser's language preferences (as available from `window.navigator.languages`) are used, although the language(s) passed to `Mirador.JsonLd.getTextValue` always take precedence over these.

### DIFF
--- a/js/src/utils/jsonLd.js
+++ b/js/src/utils/jsonLd.js
@@ -1,20 +1,66 @@
 (function($) {
-  
+  /** Get the language to use for displaying the given propertyValue.
+   *
+   * Uses the algorithm described in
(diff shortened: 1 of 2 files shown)


rsinghal:
@jbaiter Can you switch this PR to `2.1.1`?

aeschylus:
This looks good to me, but we need to rebase to 2.1.1.

jbaiter:
Done :-)

PR #1141, opened 2016-09-27
reset image manipulation controls on `updateImage`

### LINKED ISSUE #1130: Image manipulation settings are preserved between next/previous image
... without being applied to the current image.

This issue forces the user to reset the settings each time he opens the next/previous image. Moreover it causes a bug on the buttons state (e.g. Invert or Grayscale), where the active/inactive states are inverted

rsinghal:
fixes #1130


aeschylus:
This seems like something we want in the 2.1.1, no? Can you rebase with 2.1.1? @rsinghal

PR #1142, opened 2016-09-29
updated package.json and bower.json dependency and self versions, swa…

### LINKED ISSUE #1126: Problems with annotationEndpoint
Does anybody use Mirador with annotation endpoint different from the local storage? It seems to me that when used with simpleASEndpoint or any other that loads the annotations asynchronous there are problems with the annotations rendering. What I found is that it renders the shapes before it loads the new annotations list.

### LINKED ISSUE #1127: Chrome (latest): uncaught range error on projectmirador.org - reproducible by reloading page a few times in quick succession
I got this by hitting ctrl+f5 a few times after the page had already loaded successfully.  Windows version doesn't seem to matter.  This is Chrome latest version (Mirador.js: 4) 

Apparently caused by an infinite loop in JS, so I suspect it can probably be replicated on OSX as well.  And yup, I just did.  Reloading the page a few times can reproduce.  See screenshots (windows + mac) below.

http://stackoverflow.com/questions/18253431/uncaught-rangeerror-maximum-call-stack-size-exceeded-chrome-js-error
http://stackoverflow.com/questions/6095530/maximum-call-stack-size-exceeded-error

chrome_osx_sierra_aw_snap
chrome_windows
chrome_osx_sierra

aeschylus:
…pped jquery 1.9.x for 3.1.1. Fixes #1127, resolves #1126.


aeschylus:
@rsinghal The settings have been updated so each thing requires review. I never heard back from github support, but I will try to see if I can re-enable the required checks.

aeschylus:
Hm. I'm getting this failure now as well. I'm a bit confused why this test wasn't failing before, since the bug often occurred. Worth fixing. Thanks for the review.

aeschylus:
There seems to be a problem with me having used the IIIF/mirador as a remote, so I am deleting this branch and PR, and re-submitting.

PR #1144, opened 2016-10-06
Update README to trigger travis reset.

aeschylus:
We need to change this to trigger the new travis org connection.


stweil:
Did you rename the project from `IIIF/mirador` to `ProjectMirador/mirador`? That might be the reason why the `coverage/coveralls` CI test does not get the expected status.

aeschylus:
Yes, it is the reason the badge, like the travis badge, is/was wrong, but there are also settings reasons. See https://coveralls.io/repos/github/ProjectMirador/mirador/badge.svg

It's still returning an "unknown" value.

PR #1146, opened 2016-10-07
Fix some typos in comments (found by codespell)

stweil:
Signed-off-by: Stefan Weil <sw@weilnetz.de>


aeschylus:
Hi Stefan, I had another look at this and it seems some of the spelling corrections were made to our dependencies. It is perhaps concerning that programmers don't always spell correctly, but I don't want to edit our dependencies in any way that might bother the build and dependency management system down the line. 

Would you please reverse those other spelling mistakes that occur on objects in the lib folder? If not, we can always close this request and do this on another branch.

stweil:
@aeschylus, which files are you referring to? The PR does not include changes to files in a `lib` folder.

rsinghal:
I think he is referring to the files in `css/plugins`.  Those are third party files, so we don't want to mess with them.

stweil:
I removed the `css/plugins` related commit from the PR now.

Issue #1151, opened 2016-10-17
Save/Cancel after editing annotation not working


radpet:
Hello,

I am not sure if this is duplicate issue but when I create annotation the panel buttons work well, however when I try to edit existing one and I click one of the panel buttons it does not call the handler and only adds #action(save/cancel) to the url. This is happening on the http://projectmirador.org/demo/ and it is maybe  regression caused by the jquery update?

PR #1157, opened 2016-10-26
Migrate annotation loading fixes

radpet:
@aeschylus  Those are the fixes that re-renders the annotations when new annotations list has been loaded

@rsinghal  I have included the fix for https://github.com/ProjectMirador/mirador/issues/1151 inside since it is not much code

### DIFF
--- a/js/src/annotations/osd-region-draw-tool.js
+++ b/js/src/annotations/osd-region-draw-tool.js
@@ -76,37 +76,38 @@
       }
     },
 
-    render: function() {
+    render: function () {
+
+      if(this.parent.mode !== $.AnnotationsLayer.DISPLAY_ANNOTATIONS){
+        return ;
+      }
       this.svgOverlay.restoreEditedShapes();
       this.svgOverlay.paperScope.activate();
       this.svgOverlay.paperScope.project.clear();
       var _this = this;
       _this.annotationsToShapesMap = {};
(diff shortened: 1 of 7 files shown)


js/src/annotations/osd-region-draw-tool.js
```diff
@@ -121,6 +122,7 @@
         getAnnoFromRegion: _this.getAnnoFromRegion.bind(this)
       });
       this.svgOverlay.paperScope.view.draw();
+      _this.eventEmitter.publish('annotationsRendered.' + _this.windowId);
```
aeschylus: I like that we are no longer relying on deferreds here.

js/src/annotations/osd-region-draw-tool.js
```diff
@@ -76,37 +76,38 @@
...
       this.svgOverlay.restoreEditedShapes();
       this.svgOverlay.paperScope.activate();
       this.svgOverlay.paperScope.project.clear();
       var _this = this;
       _this.annotationsToShapesMap = {};
-      var deferreds = jQuery.map(this.list, function(annotation) {
-        var deferred = jQuery.Deferred(),
-        shapeArray;
+
+      for (var i = 0; i < this.list.length; i++) {
```
aeschylus: Why not this.list.forEach() ? It's much more readable, and matches the style of the rest of the codebase.

PR #1161, opened 2016-10-28
make the package public to enable bower registration. This will ease …

aeschylus:
…integration with rails asset pipeline.


aeschylus:
@rsinghal

PR #1163, opened 2016-10-29
Fix some slot and window memory leaks

radpet:
Add additional event triggered only when slot has been removed
Unsubscribe from all event handlers when slot has been destroyed in order to prevent memory leaks


js/src/workspaces/slot.js
```diff
@@ -236,7 +235,20 @@
...
+      var _this = this;
+      this.events.forEach(function(event){
+        _this.eventEmitter.unsubscribe(event.name,event.handler);
+      });
+    },
+
+    clearSlot: function () {
+      this.destroyEvents();
+
+      // @TODO This is not a task that slots should fulfil
```
radpet: @aeschylus Please check this, I am not really into what 

>  // This prevents the save controller
 +      // from attempting to re-save the window
 +      // after having already removed it.

it actually fixes

spec/workspaces/slot.test.js
```diff
@@ -2,31 +2,35 @@ describe('Slot', function () {
...
-    };
-
     this.slot = new Mirador.Slot({
       window: mockWindow,
       eventEmitter: this.eventEmitter
     });
   });
 
   describe('listenForActions', function () {
-    it('should respond to windowRemoved', function () {
```
radpet: window.js should handle this test case I suppose?

aeschylus: Yes, I think that should be added. Feel free to delete the test from here, but it should be added in Window.

If we can boost the test coverage a bit, I think this will be ready to go.

PR #1168, opened 2016-11-07
Remove bower from test suite

mejackreed:

PR #1169, opened 2016-11-07
Removes jquery-awesome-cursor from bower

mejackreed:
No need to have this in bower anymore.


mejackreed:
Note this could be unecessary if https://github.com/ProjectMirador/mirador/pull/1168 is merged.

PR #1171, opened 2016-11-07
The build seems to be incompatible with node v7.0.0.

mejackreed:
This commit codifies that in specifying the engine version compatibility.


aeschylus:
Yeah. I'm not sure what causes it, but I know node 7.0.0 is not stable yet. Thanks for the version freeze.

aeschylus:
Can you update this to point at 2.1.2?

mejackreed:
Done can you reopen? I don't have the rights.

Issue #1167, opened 2016-11-07
Tests fail locally after installing for the first time


mejackreed:
After cloning, installing, and setting up, 5 tests fail locally.
all failing tests are in `/mirador/spec/viewer/manifestListItem.test.js`

Is this to be expected or maybe an issue with installed packages/phantomjs?

csbailey5t:
@aeschylus and I started looking at these last week. If I remember correctly, they may be failing due to space issues in the midst of checking those long strings of HTML. I'm going to take a look at them  and see if we can rewrite to check the same thing without worrying about correct numbers of spaces.

rsinghal:
Yes - I wrote it to (naively) do a string compare, rather than a DOM compare.

PR #1175, opened 2016-11-07
Test using jasmine-jquery matchers rather than string matching.

csbailey5t:
This PR addresses #1167, where 5 tests regularly failed due to a single space mismatch. Instead of using string comparison, the tests now use jasmine-jquery matches on the DOM. This PR does maintain the general approach to checking for preservation of the manifest order as the previous tests.

### DIFF
--- a/spec/viewer/manifestListItem.test.js
+++ b/spec/viewer/manifestListItem.test.js
@@ -36,6 +36,12 @@ describe('ManifestListItem', function () {
     var html = jQuery.parseHTML(this.appendTo[0].outerHTML);
     var testImage = jQuery(html).find('li[data-index-number="0"] img');
 
+    var ids = [];
+    jQuery(html).find('li').each(function(i, el){
+      ids.push(jQuery(this).data('index-number'));
+    });
+
+    expect(ids).toEqual([-1, 0, 1, 5, 8, 13, 14, 16]);
     expect(testImage).toHaveData('image-id', 'http://www.example.org/iiif/book1/canvas/p1');
 
   });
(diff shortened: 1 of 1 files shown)


csbailey5t:
@rsinghal Perhaps I don't understand how your initial tests worked then. It seemed like you were moving around the book 1 manifest when creating the new ManifestListItem, then in the expectedHTML checking for whether the ```li``` with the correct index contains the expected markup based on the book 1 address.

csbailey5t:
Ah, I see it now. I think we just need to check the ```li``` structure now. If we're just testing index positioning, we don't actually need to check the book 1 uri at all.

rsinghal:
I did this a while ago, so my memory is rusty.

I had to move around book1 (which corresponds to the id in `dummyManifest.json`) in order to get ManifestListItem to insert it at various spots within the `appendTo` section, since ManifestListItem gets the index based on the `data` array. So, given an existing set of `li` tags, would ManifestListItem insert it at the correct position depending on the index value of the manifest I'm inserting?  It would have been easier to just give ManifestListItem the index where it should be inserted, but I probably wrote that code that made it difficult to give that value, so I fudged it a bit by using the `data` array.

rsinghal:
Yes, we just need to make sure all of the `data-index-number` values are correct after inserting the new manifest.

csbailey5t:
@rsinghal I updated the code to actually check the order of the ```li``` elements, still using Jasmine rather than string matching. It's perhaps a bit indirect, but seems to now do both things the original tests did.

Issue #1154, opened 2016-10-20
jQuery 3.0.0 breaks jQuery-UI dialog


gigamorph:
See https://github.com/jquery/jquery/issues/3157

Maybe we should add https://code.jquery.com/jquery-migrate-3.0.0.min.js to the package as recommended or at least document it?

PR #1183, opened 2016-11-08
Add in jquery migrate as a failsafe 

csbailey5t:
Adds jQuery Migrate as a failsafe for conflicts between jQuery 3 and jQuery UI Dialogs, addressing #1154.


Gruntfile.js
```diff
@@ -239,7 +240,7 @@ module.exports = function(grunt) {
 
     githooks: {
       all: {
-        'pre-commit': 'jshint cover'
+        'pre-commit': 'jshint'
```
mejackreed: Was this meant to be in the commit?

aeschylus: I think it was meant to, but shouldn't go in. I misunderstood @csbailey5t. The githook/test process needs some love, but in the meantime, if you need to commit something with failing test or fallen coverage to compare against travis, just `rm .git/hooks/pre-commit`. It will be re-added automatically on the next `npm install`.

mejackreed: I like to use `--no-verify` in this case too

aeschylus: Actually we do want this here, because there is no grunt cover task anymore. It will error out before the commit for that reason.

mejackreed: Sure.. my initial thought was, if you want to remove jQuery migrate at some point, you could just revert this commit. Now with this being in the same commit it would potentially have additional side effects.

csbailey5t: I think it does need to be removed since it's broken code, but in a separate commit/pr just for separation of concerns. I wasn't paying close attention when I included it in this commit. Are we wanting to recreate the test githook elsewhere, or simply remove it?

PR #1185, opened 2016-11-09
Don't wrap existing <a>s in metadata with another one (fixes #571)

### LINKED ISSUE #571: Links to URIs in HTML formatted metadata do not show properly
Some HTML markup is allowed in certain IIIF Presentation manifest metadata fields. But if they contain absolute URL links, those links do not show up properly because Mirador tries to convert them into links again.
A simple solution could be something like this: https://github.com/IIIF/mirador/commit/297f7eb34fd39392b976eeba464e224acf4f5b52

jbaiter:
Currently, when a metadata field contains HTML with an anchor tag that targets an absolute URL, Mirador's logic for wrapping HTTP-URLs will trigger and wrap an additional anchor-tag around it.

Example: `This is a <a href="http://example.com">link</a>` will become `This is <a href="<a href="http://example.com target="_blank">http://example.com</a>">link</a>`

This fix adds a check to `Mirador.MetadataView.addLinkstoUris` that prevents this behavior.

### DIFF
--- a/js/src/widgets/metadataView.js
+++ b/js/src/widgets/metadataView.js
@@ -205,7 +205,7 @@
           textWithLinks = text,
           matches;
 
-      if (typeof text === 'string') {
+      if (typeof text === 'string' && textWithLinks.indexOf('<a ') == -1) {
         matches = text.match(regexUrl);
 
         if (matches) {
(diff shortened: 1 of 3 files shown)


aeschylus:
This looks good to me. Thanks for applying to the development branch, and for the tests.

If you have any feedback on the CONTRIBUTING.md on this front, please let us know.

Thanks again for the fix.

jbaiter:
> If you have any feedback on the CONTRIBUTING.md on this front, please let us know.

Not CONTRIBUTING directly, but it would be great if you could make the branch that PRs should be based on the default one. Currently the information on which branch to base on is pretty hidden in the CONTRIBUTING file, which isn't really conductive to drive-by PRs like these :-)

PR #1189, opened 2016-11-11
Fix grunt watch server

aeschylus:


Gruntfile.js
```diff
@@ -43,12 +42,6 @@ module.exports = function(grunt) {
     'js/lib/modernizr.custom.js'
   ],
 
-  // libraries/plugins for running tests
-  specJs = [
```
mejackreed: Removal of the vestigal bower code is also in 7ce88c281cbad55a9492879ce5a026bbb56b6292

Issue #1110, opened 2016-09-09
Annotation figure can be selected when not in edit mode


ViktorRibchev:
Please see: 
[Mirador 2 select.zip](https://github.com/IIIF/mirador/files/463379/Mirador.2.select.zip)

A figure can be selected when not in edit mode when the figure is next to a border of the canvas.

camillevilla:
This ticket is being closed as part of the Mirador 3 issues review. We encourage all Mirador 2 implementers to try out Mirador 3 and leave feedback in new tickets.

In Mirador 3, annotation creation has been moved out of the core library and into [projectmirador/mirador-annotations](https://github.com/ProjectMirador/mirador-annotations).

PR #1190, opened 2016-11-11
Restricts an annotation to the bounds of the tileSource image.

### LINKED ISSUE #1111: Rectangle annotation tool logic breaks
Rectangle annotation tool logic breaks when two of the borders are touching the canvas borders. 
Please see: 
[Mirador 1 logic.zip](https://github.com/IIIF/mirador/files/463391/Mirador.1.logic.zip)

mejackreed:
Previously, an annotation could be forced out of the image canvas,
this change only allows the annotation size change (delta) to be
modified if the drag point is within the bounds of the image.

Fixes #1111 

This logic is a bit more simplified and does seem to change the annotation experience a little. If a user quickly drags an annotation past the canvas area, the annotation may not snap to the bounding area. This didn't seem too devastating to me, but would like input from more experienced Mirador users/devs.

I believe this also may address #1110 but would like external verification.

anno

### DIFF
--- a/js/src/annotations/osd-svg-overlay.js
+++ b/js/src/annotations/osd-svg-overlay.js
@@ -451,31 +451,56 @@
       return mousePosition;
     },
 
-    adjustDeltaForShape: function(lastPoint, currentPoint, delta, bounds) {
-      //first check along x axis
-      if (lastPoint.x < currentPoint.x) {
-        //moving to the right, delta should be based on the right most edge
-        if (bounds.x + bounds.width > this.viewer.tileSources.width) {
-          delta.x = this.viewer.tileSources.width - (bounds.x + bounds.width);
-        }
-      } else {
-        //moving to the left, prevent it from going past the left edge.  if it does, use the shapes x value as the delta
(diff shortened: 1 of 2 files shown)


csbailey5t:
@mejackreed I pulled your branch and am testing it out. Right now, I'm actually able to create a rectangle that I can then drag (not resize, but just drag) outside of the bounds of the canvas.

aeschylus:
@csbailey5t @mejackreed Yeah, I'm also getting the problem when initialising the rectangle from outside the image area.

Both of you guys have spiked on this one, so can you maybe pair on it today? I will get in the integration fixes so this test won't fail intermittently.

csbailey5t:
I'm also able, on the branch from @mejackreed, with a little bit of work, to keep resizing the box until I can actually flip it outside of the canvas (at last it looks to be out of the canvas). This may be different behavior than what we're fixing, but seems related. 

rect_logic_break_flip

mejackreed:
This doesn't actually work.. I was hopeful though

PR #1196, opened 2016-11-15
update rectangle bounds to fix annotation logic breaks

### LINKED ISSUE #1111: Rectangle annotation tool logic breaks

mejackreed:
Fixes #1111


mejackreed:
Well this doesn't work either.

Issue #1159, opened 2016-10-27
JS error in thumbnails slider (bottomPanel)


regisrob:
To reproduce the bug (with the latest master branch):
- open any object from the load window
- open any image in ImageView
- replace current objet with another one (any book item in the list)
- open any image in Image View and try to navigate to the next image by using the slider => the expected action is not performed, and the following error appears in the console:

`mirador.js:4 Uncaught Error: cannot call methods on slider prior to initialization; attempted to call method 'option'`

_Tested on MAC OS 10.11.6 / Chrome 54.0.2840.71_

aeschylus:
@rsinghal @sdellis @gigamorph 

This is a result of the window update function not being idempotent. There are still event observations bound on some parents of re-loaded elements, so this is causing a number of strange bugs due to events stacking on top of each other.

Awhile back we made the decision to try to "refresh" an existing window when a new object was added, rather than completely destroying the existing window and recreating a new one. I'm now questioning this choice, as it adds a great deal of complexity and is essentially an admission that we want to manage a bunch of mutable shared state between windows, rather than clearing the way and starting over.

@rsinghal and I are planning to pair today with a small group from Stanford to get them oriented, and this might be a good issue to start with, since it touches on Mirador's core and has to do with the event system, without being so complex that even @rsinghal and I don't fully understand offhand what might be going wrong.

rsinghal:
One motivation behind "refreshing" an existing window was to more easily preserve the user state of that window when a new object was loaded into it (e.g., if they had the Metadata panel open and opened a new object, open the Metadata panel for the new object).  We can move ahead with destroying and creating a new window as needed and save some of the state to the slot and pass it to the new window.  Or, we can do away with preserving the user state and just use the configured state on creation of a new window (which is what currently happens with a brand new window/slot)

beaudet:
ConservationSpace definitely has a use case where the slot configuration needs to be saved and later reconstructed. Perhaps if a null configuration is passed, then it uses defaults?

aeschylus:
The way we're moving forward with it at the moment is to destroy the window, but initialise the new window with similar settings to the original window. It's quite a challenge to test (we're starting with the tests this time), but we're making progress.

The current behaviour for null configurations is to use defaults via object extension.

aeschylus:
[...]

PR #1197, opened 2016-11-15
Replace old windows instead of reusing

aeschylus:


aeschylus:
I have solved the bug, but at the expense of writing new integration tests, which I have not been able to get to run more than two event hops down. Would be very open to some help with that early tomorrow morning @mejackreed, @versae, @csbailey5t

aeschylus:
This also does not completely solve the issue for the changeLayout function yet, but I am working on that now.

mejackreed:
Is there an issue to link to here?

csbailey5t:
This addresses #1159.

PR #1198, opened 2016-11-15
reset viewerEvents if the content of the tooltip is updated because o…

### LINKED ISSUE #1195: Edit button doesn't always work when dealing with overlapping annotations
When we have overlapping annotations, such as when two rectangles overlap, when in pointer mode, sometimes the edit button simply doesn't work. I can hover over the overlap, have the popup to go into edit or delete come up for both, but the edit button does nothing. If I hover over each box individually, it will.

rsinghal:
…f multiple, stacked annotations

closes #1195 

@csbailey5t Can you confirm this is working as well?

### DIFF
--- a/js/src/annotations/annotationTooltip.js
+++ b/js/src/annotations/annotationTooltip.js
@@ -161,11 +161,15 @@
             _this.removeAllEvents(api, params);
           },
           visible: function (event, api) {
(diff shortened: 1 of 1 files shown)


aeschylus:
@csbailey5t Looks good to me too.

Go ahead and merge.

PR #1207, opened 2016-11-23
If canvas.thumbnail is level 0, don't use

### LINKED ISSUE #1170: Mirador assumes an image service available via canvas.thumbnail supports arbitrary /w,/ requests
Related to #872

I've been experimenting with optimising thumbnails, and including enough info in the manifest for a client to request lots of fast thumbs without having to dereference an info.json.

Each canvas has a thumbnail property in the manifest like this:

```javascript
"thumbnail": {
  "@id": "https://dlcs.io/thumbs/wellcome/1/79fbc149-42b9-43d2-979d-c66c435073e4/full/64,100/0/default.jpg",
  "@type": "dctypes:Image",
  "service": {
    "@context": "http://iiif.io/api/image/2/context.json",
    "@id": "https://dlcs.io/thumbs/wellcome/1/79fbc149-42b9-43d2-979d-c66c435073e4",
    "protocol": "http://iiif.io/api/image",
    "height": 1024,
    "width": 651,
    "sizes": [
      {
        "width": 64, "height": 100
      },
      {
        "width": 127, "height": 200
      },
      {
        "width": 254, "height": 400
      },
      {
        "width": 651, "height": 1024
      }
    ],
    "profile": [
      "http://iiif.io/api/image/2/level0.json"
    ]
  }
}
```

from http://library-uat.wellcomelibrary.org/iiif/b28047345/manifest

This is a level 0 service with only these four sizes available, but enough of the image information inline in the manifest for a client to use straight away. This is telling a client "I *only* have these sizes available for thumbs, I really want you to use them!"

Mirador identifies the service - but doesn't recognise it as a level 0 service, so starts asking for thumbs from it that are not available. It can't just do this:

`thumbnailUrl = $.Iiif.makeUriWithWidth(service['@id'], width, version);`

(https://github.com/ProjectMirador/mirador/blob/master/js/src/utils/utils.js#L43)

...instead it needs to decide whether it can use one of the sizes available, and if so construct an Image API request to the thumb service for it, and if not, use the main image service. Identifying that it's level 0 and using the main image is likely to be the quickest solution if client-side scaling would cause problems.

I'd like to release that inline thumbnail service to the live environment for Wellcome, but I don't want to do that if it means a Wellcome manifest has broken thumbs when viewed in Mirador!

tomcrane:
Fixes #1170 very conservatively - `getThumbnailForCanvas` will carry on doing what it did before unless there is an explicit level0 profile assertion inline in a service belonging to canvas.thumbnail, in which case it will fall back to the thumbnail ['@id'].

A better approach would be to flip this round, and only attempt to make an ad-hoc thumbnail request if something like:

`$.Iiif.supports(profile, 'sizeByW')`

where `supports` would understand how Image API profiles work - both listed features and features implied by the levelN profile URI. Happy to have a go at that but it's a bigger piece of work.

There are two `getThumbnailForCanvas` functions - in manifest.js and utils.js. I changed them both.


azaroth42:
:+1:

And demonstrates the weirdness of two thumbnail functions.  Refactoring seems useful here!

js/src/manifests/manifest.js
```diff
@@ -93,12 +94,21 @@
...
+               compliance = $.Iiif.getComplianceLevelFromProfile(service.profile);    
+            }
+            if(compliance === 0){
+                // don't change existing behaviour unless compliance is explicitly 0            
+                thumbnailUrl = canvas.thumbnail['@id'];
+            } else {
+                // Get the IIIF Image API via the @context
+                if (service.hasOwnProperty('@context')) {
+                    version = $.Iiif.getVersionFromContext(service['@context']);
+                    console.log('version');
```
rsinghal: Can you remove the `console.log`?

js/src/utils/utils.js
```diff
@@ -34,13 +35,21 @@
...
+            compliance = $.Iiif.getComplianceLevelFromProfile(service.profile);    
+        }
+        if(compliance === 0){
+            // don't change existing behaviour unless compliance is explicitly 0            
+            thumbnailUrl = canvas.thumbnail['@id'];
+        } else {
+            // Get the IIIF Image API via the @context
+            if (service.hasOwnProperty('@context')) {
+                version = $.Iiif.getVersionFromContext(service['@context']);
+                console.log('version');
```
rsinghal: Can you remove the `console.log`?

tomcrane:
@rsinghal - logging removed, sorry for taking so long!

Issue #1204, opened 2016-11-18
Create annotation iframe bug 


radpet:
In CS space project we are using mirador inside iframe and we found the following issue:

Start drawing annotation
Drop the cursor out of the iframe
Then when come back to the iframe a new shape is created and the old one (the initial) is drawn on the canvas but it is not active and persisted after save

It seems the mouse up event is not fired properly. Any ideas for fix?

I recorded a video representing the issue :)

[mouseup_bug_iframe.ogv.zip](https://github.com/ProjectMirador/mirador/files/600578/mouseup_bug_iframe.ogv.zip)

camillevilla:
This ticket is being closed as part of the Mirador 3 issues review. We encourage all Mirador 2 implementers to try out Mirador 3 and leave feedback in new tickets.

In Mirador 3, annotation creation has been moved out of the core library and into [projectmirador/mirador-annotations](https://github.com/ProjectMirador/mirador-annotations).

PR #1214, opened 2016-12-07
Update paper js version

radpet:
Update paper js version
Add workaround for paper.js shape fill color because it does not detect mouse cursor on shapes with transparent fill color

Seems to fix several issues:

- https://github.com/ProjectMirador/mirador/issues/1212
- https://github.com/ProjectMirador/mirador/issues/1204


Gruntfile.js
```diff
@@ -36,7 +36,7 @@ module.exports = function(grunt) {
     'js/lib/URI.min.js',
     'js/lib/mousetrap.min.js',
     'js/lib/isfahan.js',
-    'js/lib/paper-full.min.js',
+    'js/lib/paper-core.min.js',
```
radpet: +1

PR #1215, opened 2016-12-13
fix arguments in the publish event 'ADD_MANIFEST_FROM_URL'

kbois:
This PR adds the location attribute "Added from url" back in the manifest list panel.


regisrob:
This should fix #1216

PR #1228, opened 2017-01-09
Fix in metadata manipulation with uppercase letters

morpheus-87:
Previously all uppercase letters in the metadata were prefixed by a space. This manipulation is not necessary, because it just looks strange.

 Additionally the first character of every token in the metadata was converted to uppercase. This manipulation is not necessary, because it's already done by a CSS rule.


js/src/widgets/metadataView.js
```diff
@@ -164,11 +163,7 @@
     var label = attr;
 
     label = label.replace(/^@/, '');
-    label = label.replace(/([A-Z])/g, ' $1');
     label = label.replace(/\s{2,}/g, ' ');
-    label = label.replace(/\w\S*/g, function(txt) {
```
rsinghal: Can you actually leave these lines in and remove the relevant line in the CSS https://github.com/ProjectMirador/mirador/blob/2.1.3/css/mirador.css#L1802?  This does a better job at capitalizing than the CSS does (see `Creator(S)`):

screen shot 2017-01-11 at 2 49 40 pm

jbaiter: The problem is that while the CSS can be modified by MIrador users, the capitalization via JavaScript is currently hard-coded. It would be good to offer this option, since the convention of title-casing every word is pretty much an English-only thing, or at least it is not commonly used in German. Maybe we could restrict this style to English labels and labels that do not declare a language?

jbaiter: Any update on this? The CSS should obviously removed, but in my opinion title-casing should be done by the entity creating the manifests and not by the client.

aeschylus: @jbaiter Agreed that this should be a user choice, and otherwise falls on the repository or publisher of the manifest. I will merge this as is, and then do another PR to remove the CSS capitalisation.

PR #1235, opened 2017-01-16
Display bottom panel when in fullscreen mode

jbaiter:
For some reason the bottom panel is disabled when Mirador is in fullscreen mode, making navigation pretty tedious. This PR removes the calls to disable the bottom panel in fullscreen mode.


aeschylus:
Let's take this up on the call this Thursday. I remember this was requested at one point, so perhaps it's worth making this a configuration option.

Additionally, it might do to improve the responsiveness of the bottom panel so that it isn't so tiny on large displays. The size also causes many more thumbnails to need to load than may really be necessary for the user to navigate right off the bat.

For now I'll mark this as a `won't fix`.

PR #1236, opened 2017-01-18
Configurable language

morpheus-87:
This PR makes the language of the UI configurable in the init step.

### DIFF
--- a/js/src/viewer.js
+++ b/js/src/viewer.js
@@ -43,15 +43,27 @@
     init: function() {
       var _this = this;
 
-      //initialize i18next
-      i18next.use(i18nextXHRBackend).use(i18nextBrowserLanguageDetector).init({
+      // i18next options
+      var i18nextOptions = {
         fallbackLng: 'en',
         load: 'unspecific',
         debug: false,
         backend: {
           loadPath: _this.state.getStateProperty('buildPath') + _this.state.getStateProperty('i18nPath')+'{{lng}}/{{ns}}.json'
         }
-      }, _this.setupViewer.bind(_this));
+      };
+
+      // set the language from configuration
+      var configuredLanguage = _this.state.getStateProperty('language');
+      if(configuredLanguage){
+        i18nextOptions.lng = configuredLanguage;
+      }
+
+      //initialize i18next
+      i18next.use(i18nextXHRBackend).use(i18nextBrowserLanguageDetector).init(
+        i18nextOptions,
+        _this.setupViewer.bind(_this)
+      );
       // because this is a callback, we need to bind "_this" to explicitly retain the calling context of this function (the viewer object instance));
     },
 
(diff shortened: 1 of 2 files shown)


aeschylus:
This has some conflicts with the improvements from one of your other PRs. Also, there is another way to implement configuration which is used in other areas of the app, so can you resubmit this as a new PR with that pattern? 

A good example can be found by looking at the [`saveController.js`](https://github.com/ProjectMirador/mirador/blob/2.1.4/js/src/utils/saveController.js) and the `DEFAULT_SETTINGS` in [`settings.js`](https://github.com/ProjectMirador/mirador/blob/2.1.4/js/src/settings.js)

Marking as wontfix for now, but open to a revised PR.

morpheus-87:
I checked your change request and saw, that there is no other way to get configuration parameters in `viewer.js` except reading it from the state.

It seems, that the behaviour you are talking about, was removed in commit d0f6310 by @rsinghal.

rsinghal:
That's the correct way to get information from the configuration @aeschylus.  Not sure what you are looking for?

aeschylus:
Sorry, was closing like 12 PRs simultaneously and don't think I grasped what this was trying to do. If you need to get the configuration dynamically (rather than on initialisation of the whole App using extend), this makes sense.

Can you settle the conflict in viewer.js and we can merge from there?

morpheus-87:
In the meantime, I fixed the conflict, @aeschylus.

aeschylus:
Looks like viewer.js is still a stub, but that would be the place I think. You would need to test that the language was one thing in the config before the runtime initialisation.

Thanks for fixing this up. Looking forward to merging.

morpheus-87:
Is is really necessary to have a separate testcase for setting the language, @aeschylus?

aeschylus:
@morpheus-87 Always a fair question. I think it could be useful since it's happening at runtime. Can one of us help set up the test case?

morpheus-87:
Would be great, if somebody could help me with setting up the test case.

morpheus-87:
Could somebody help me with setting up the test case?

rsinghal:
@morpheus-87 Hopefully this will help you set up the tests:

https://github.com/ProjectMirador/mirador/blob/detail_images_and_image_choice/spec/widgets/sidePanel.test.js

Specifically, this is to create a skeleton config: https://github.com/ProjectMirador/mirador/blob/detail_images_and_image_choice/spec/widgets/sidePanel.test.js#L17

And then this actually tests something that is in the config: https://github.com/ProjectMirador/mirador/blob/detail_images_and_image_choice/spec/widgets/sidePanel.test.js#L102

morpheus-87:
I just updated the branch and set up a testcase in `spec/mirador.test.js`.

aeschylus:
Thanks @morpheus-87. This looks good to me.

PR #1238, opened 2017-01-19
Fixed Travis build to use a C++11 compatible compiler

bitzl:
Fixed Travis build to use a C++11 compatible compiler as needed to install the node module iltorb.

Several Travis builds fail as node-gyp compiles brotli and the Node.js bindings to install iltorb. Brotli is written using C++11, which is currently only supported by g++ version 4.8. Travis images have a much older version (something like 3.x) as default.

This fix follows a bit hidden part of the Travis documentation: https://docs.travis-ci.com/user/languages/javascript-with-nodejs#Node.js-v4-(or-io.js-v3)-compiler-requirements


aeschylus:
@bitzl 

Thanks, this helps provide some context for this configuration change in some of the other PRs. I would like to merge this to enable those others, but can you describe in more detail what problems this is causing? We are seeing travis builds pass fairly consistently.

bitzl:
This PR fixed a problem with old GCC versions in the Travis build image that prevented builds over a month ago. Meanwhile Travis upgraded its build images, making the fix obsolete.

PR #1239, opened 2017-01-19
Localize rights and links information in metadata

morpheus-87:
This PR reintroduces some german locales, that were added in #1227, but disappeared later on.

My assumption is, that this was caused by clicking the button with the label `Update branch`, which is displayed in every pull request. The purpose of this button is to keep the source branch up to date with the target branch. For me, it looks like all files, that were changed in both the source and the target branch, will be overwritten in the source branch by clicking this button. When merging the target to the source branch on the commandline, I would expect merge conflicts, that are obviously ignored by the web interface.


.travis.yml
```diff
@@ -1,9 +1,17 @@
...
 before_install:
-- npm install
+  - npm install
 # before-script:
-# - grunt test cover
+#   - grunt test cover
 script:
-- npm run travis
+  - npm run travis
+env:
```
aeschylus: Happy to merge this one in if we can remove the environmental changes, or at least explain why they're useful. Is this a local configuration?

morpheus-87: I updated the corresponding branch, so that the changes to the Travis configuration are reverted.

aeschylus:
Coverage is not functionally reduced. We need to adjust the threshold.

PR #1244, opened 2017-01-31
Augmented debugging facility for EventEmitter

gigamorph:
In "debug" mode, EventEmitter used to print traces when subscribe, unsubscribe, or publish is called. Now it prints out call information also when an event handler registered via subscribe gets called. 

Since by default it prints out an overwhelming amount of traces to the console, the default function now used is `console.log` instead of `console.trace`. `console.trace` can be used by setting `this.trace` to `true`. Also you can opt not to include certain events in logging by overwriting `excludePatterns`.


js/src/utils/eventemitter.js
```diff
@@ -1,38 +1,109 @@
...
   $.EventEmitter = function(config) {
     jQuery.extend(true, this, {
-      'debug': $.EventEmitter.debug
+      'debug': $.EventEmitter.debug, // Log events to console if true
+      'trace': $.EventEmitter.trace,  // Use console.trace for logging if true, console.log if false
+      'debugExclude': $.EventEmitter.excludePatterns // substring patterns for event IDs to exclude
     }, config);
     this.emitterId = $.EventEmitter.nextId();
+    if (this.debug) {
+      this.logger = new $.EventEmitter.Logger({
```
aeschylus: Nice use of browser-independent logging.

spec/utils/eventemitter.test.js
```diff
@@ -54,22 +54,64 @@ describe('EventEmitter', function () {
...
+    var em = new Mirador.EventEmitter({debug: true});
     var eventName = "foo";
-    spyOn(console, 'trace');
+    spyOn(console, 'log');
     em.publish(eventName);
-    expect(console.trace).toHaveBeenCalled();    
+    expect(console.log).toHaveBeenCalled();
+  });
+
+  it('should recognize trace option when logging events', function() {
```
aeschylus: Thanks for tests!

PR #1267, opened 2017-02-16
Add directionality plugin to TinyMCE and configure for both RTL and LTR

### LINKED ISSUE #1251: Configure TinyMCE to allow annotation creation right-to-left and bi-directional

mejackreed:
Closes #1251 

[UPDATED] to support RTL viewing of anno


aeschylus:
Cool; that was straightforward enough. We should talk about how configurable this needs to be from the open source perspective. I think rtl is worth leaving in, but perhaps these plugins should be passable from the Mirador config.

mejackreed:
That seems like a great idea. What would you propose? The TinyMCE plugins and toolbar be configurable from the mirador config?

aeschylus:
Yeah, I guess we have two choices:

We allow the user to pass in "tinyMCE config" directly, and fold it in on top of whatever defaults we have. This would give them fine-grained control, but perhaps might be above the heads of some installers.

Alternatively we can create our own little set of "on/off" features and allow that to be passed in, and we would handle that ourselves explicitly as a set of profiles in the initialisation.

It's also worth noting that the entire annotation editing box is now implemented as a plugin, so it is designed to be swapped out as needed. Not sure if we should use that or promote that here, or if it is just a non-overlapping fact.

mejackreed:
@aeschylus the update pushed should allow for users to configure the TinyMCE plugins and toolbar using mirador settings. It also sets a default one (empty string) in case setting not provided.

Issue #418, opened 2015-02-19
Improve keyboard controls


snydman:

randystern:
I wish Mirador supported the up and down movement of the displayed image using the arrow keys.

When I try to really read a page, I get down to the bottom of the image window and I need to push the page up so that I can see the text I have not read yet. I would prefer to use the arrow keys -- or to have that as an option.

The two ways of moving the page-image up and down for viewing are, 1) drag with mouse or, 2) to mouse-click inside the very small navigation control within the interface (see attached screenshot).

When I drag the page with my mouse I seem to lose my place. Also, if you are reading more than a few pages, it would be nice to advance forward and backward in the page/image sequence using the keyboard.

Give it a try for yourself. This is a great letter: http://iiif.lib.harvard.edu/manifests/view/drs:23074260$1i

Thanks,

Bill Comstock, Harvard

aeschylus:
Verify this works when we integrate.

aeschylus:
I'm gathering together some of these tickets after seeing a recent pull request. This is worth discussing on some of the upcoming calls from a design perspective.

https://github.com/ProjectMirador/mirador/issues/418
https://github.com/ProjectMirador/mirador/issues/7
https://github.com/ProjectMirador/mirador/issues/572
https://github.com/ProjectMirador/mirador/issues/596
https://github.com/ProjectMirador/mirador/issues/1139

aeschylus:
It's also worth noting that the key design constraint here is how to manage focus, since we have multiple identical objects that might be acted upon with the same commands (windows and each of their components). You can see why this is difficult if you consider that Mirador already supports moving around the canvas, and zooming in and out, using WASD, arrow keys, and the - and + keys. It's difficult to use because the canvas element itself has to "have focus" (in the strict browser/accessibility sense). In @randystern's letter above, for instance, clicking on the canvas, and then using the WASD or arrow keys will let you move around as described, and you can zoom with the -/+ keys.

mejackreed:
This issue has some great historical context, but I believe may not be further actionable with the advent of Mirador 3.

PR #1268, opened 2017-02-17
Added keyboard navigation.

bitzl:
Adds keyboard nagivation features to Mirador. If there are multiple windows, the action are applied to all of them.

| Key                              | Action                                                |
| -------------------------------- | ----------------------------------------------------- |
| <kbd>←</kbd>                     | Go to previous page.                                  |
| <kbd>→</kbd> or <kbd>SPACE</kbd> | Go to next page.                                      |
| <kbd>i</kbd>                     | Show or hide information/metadata.                    |
| <kbd>↵</kbd>                     | Toggle fullscreen using the largest possible option. |


aeschylus:
Thanks for submitting this, and for the handy table. Nice use of mousetrap to handle these, but I have a few notes on the implementation and suggest that we migrate this to the keyboard issue for fuller discussion, since there are many design considerations surrounding this feature. 

Implementation notes: keyboard presses, like clicks or other DOM events, should not directly call functions on the objects on whose views they act, but rather should emit an "action" so the the state manager can propagate the appropriate effect. This ensures that the application state is always up-to-date in the save controller, and if there are other views that depend on the state change, they too can react to that change, instead of having one view re-render and another not.

These functions are also getting passed down three levels, and they affect all windows simultaneously, which I'm not sure makes the most sense for all users. It would be useful (for keyboard shortcuts) to develop and track the concept of the "focused window", have it be visually apparent, and have keyboard actions such as next/previous only affect that object. Then, there could perhaps be a modal control (holding down control or focusing the whole workspace) to enable simultaneous page-turning.

Here is one of the issues dealing with keyboard controls: https://github.com/ProjectMirador/mirador/issues/418

I'm wrangling together the many related tickets there.

aeschylus:
Closing this for now, but it will still be available here for reference.

Issue #1264, opened 2017-02-14
Add CSS compilation


gigamorph:
Use [Less](http://lesscss.org) to compile CSS. This will help modularize and maintain style definitions easier. Please assign this issue to me.

gigamorph:
Created a PR at #1270. 

I made a crude attempt to restructure the existing style definitions in Less way, namely multiple .less files with nested styles and parametrization.

I haven't tried to fix any semantic errors or duplicates/redundancies this time, but hopefully the new structure will make it easier to find and fix those.

One thing I did is parametrizing color definitions. All colors are ultimately maintained as constants in `theme/colors.less`. You can see the definitions there are rather messy, revealing need for re-theming the whole set of styles and maybe reduce the number of colors used in a consistent way. Then we can do a secondary parametrization, such as 
```
@off-white: #fafafa;
@editor-text-color: @off-white;
.annotation-editor textarea {
  color: @editor-text-color;
}
```

Other aspects can be parametrized and unified in the same way, e.g., fonts, paddings, etc., after which we could even create different themes of styles for the user to choose from.

Note: all styles are now defined under the selector `.mirador-container`. Which means the root element should have that class for the styles to be applied. For example, see the change in `index.html`:
```
<div id="viewer" class="mirador-container"></div>
```

aeschylus:
Hey @gigamorph. We're working on reviewing and merging this PR, but we're concerned that this might be a breaking change (requiring the labelling of the parent div). One solution is to generate (as part of the viewer template) a containing element with the correct class(es).

rsinghal:
Could you also check to see if the element given to Mirador has the class?  If not, add it.

gigamorph:
@aeschylus @rsinghal Can I just add the class to the viewer (`.mirador-viewer`)'s immediate parent, as @csbailey5t suggested in #1270?

rsinghal:
That's what I was suggesting.  I think it's the cleanest solution.

aeschylus:
Yeah, I don't see why not since we aren't changing any properties that might interfere with the parent's behaviour on the host page.

gigamorph:
PR is updated with the change now.

csbailey5t:
Closed in #1270

PR #1270, opened 2017-02-18
Set up Less to compile mirador.css

### LINKED ISSUE #104: Automate css namespacing so ours won't conflict with embed hosts.
It is annoying and error prone to write css while namespacing everything. A lot of our css is not namespaced because we never talked about conventions surrounding it in the beginning.

As long as we're compiling a larger stylesheet out of several anyway, we might as well pass it through a pre-compile process (or use LESS). This would also allow us to more transparently represent an actual style guide for the interface.

gigamorph:
See #1264 

Closes #104


aeschylus:
@gigamorph Thank you so much for this. This is going to be much more pleasant to work on now. It looks like there are some conflicts generated by the move to 2.2 as the development branch. Can you look into these? They should just have to do with github getting confused by the .gitignore and current css files.

gigamorph:
Interesting. I've only changed css, Gruntfile,and package.json and the coverage decreased.

gigamorph:
I have resolved the conflict. The file structure might be open to discussion and change. I deleted and ignored mirador.css because it isn't meant to be edited directly.

If you haven't seen it yet, I had some more comments about this PR at #1264.

csbailey5t:
@gigamorph The refactor looks great. Rather than requiring the user to change their markup, could you write a bit of JS to add the necessary class (`.mirador-container`) to the `#viewer` div? While it's a minor change, it would prevent an upgrade from immediately breaking installs of Mirador.

gigamorph:
@csbailey5t Ok, that sounds good.

PR #1286, opened 2017-02-24
update i18nNext path configuration to new api

### LINKED ISSUE #1285: i18nNext update broke the customizable path using resGetPath
i18nNext was updated but the path configuration was not maintained

mejackreed:
Closes #1285


mejackreed:
Additional items for discussion could be to pop the `locales` in a different directory so if this config broke in the future, we would know it.

PR #1290, opened 2017-02-25
Use npm packages instead of checked in external libraries

mejackreed:
This pull request shifts external dependencies from a checked in version, to an external source via npm.

Not all packages were able to be delivered this way, but most were.

Hopefully this gives us a few things:

 - less code to maintain (even though its external we are still maintaining it)
 - the ability move / change build later on with modern tools if needed 
 - the ability to easily update dependencies


package.json
```diff
@@ -46,6 +46,28 @@
     "test": "./node_modules/.bin/karma start ./karma.conf.js",
     "travis": "./node_modules/grunt-cli/bin/grunt ci --verbose --force & npm run test"
   },
-  "dependencies": {},
+  "dependencies": {
```
radpet: I think we should use the fixed version of the libs and update manually after having the compatibility testing

aeschylus: I agree that we should be using fixed version numbers so people's local builds don't break.

mejackreed: All of these are pinned with a `^` and from what I understand this keeps them within a major version release. Source: https://docs.npmjs.com/misc/semver#caret-ranges-123-025-004

Is within a major version release good enough for this? Do most of these not adhere to semver?

aeschylus: They do in theory, but npm has has had issues before. That said, I think we can wait to see if anything breaks, as it really shouldn't, and we do want patch updates if any arise.

PR #1296, opened 2017-02-28
Adds a mirror tool to the manipulation controls

### LINKED ISSUE #1254: Add mirror control to image manipulation tool bar
Desire is to be able to show mirror image via an image manipulation tool

mejackreed:
Closes #1254 (but some potential new issues)

However, there are some issues with this implementation. That is the
reason for setting the default to false, so this feature is only
"experimental".

1. The annotation hover listeners do not mirror.
2. All zoom/pan controls are x-reversed when mirrored. This feels
potentially intractable in Mirador and looks like OSD may need
to implement something. See discussion:
https://github.com/openseadragon/openseadragon/issues/916

Hopefully this gets the ball rolling on how we should tackle the other issues. I look forward to hearing other's expertise on this.


aeschylus:
I'll post to the OSD Gitter channel about intercepting events.

spec/locales/en/translation.json
```diff
@@ -50,6 +50,7 @@
     "load": "Load",
     "logo": "Logo",
     "metadataTooltip": "View information/metadata about this object",
+    "mirrorTooltip": "Mirror image",
```
rsinghal: This has nothing to do with your PR, but is there a way to reuse the translation files in spec, rather than having to edit two files every time there is a change?

mejackreed: Yeah totally! I asked that in the Slack channel and didn't get a response. Here I'm just following the pattern that is existing.

rsinghal: Whoops, didn't see that in slack.  I don't even see the files being used in the tests anywhere.  Am I missing any?

mejackreed: No i thought the same thing :)

rsinghal: Looks like @aeschylus added them last November with https://github.com/ProjectMirador/mirador/commit/dd412143df6c530b4f575c160d05b8255b7f1caa. I think we can remove them in a separate PR.

mejackreed: Linking this discussion to https://github.com/ProjectMirador/mirador/pull/1302

aeschylus:
An update from @iangilman can be found here: https://github.com/openseadragon/openseadragon/issues/916#issuecomment-283737611

Looks like it is not too onerous to override certain events temporarily. I still think we should try the element flipping first though (see comment in review). Presumably CSS should do it for us.

mejackreed:
@aeschylus I looked at that a bit (OSD notes) but couldn't figure out how to implement the delegation in Mirador. I know how to connect things via events, but how does that tap into OSD? Maybe someone else wants to take that up?

versae:
mirror-control_low

js/src/annotations/osd-region-draw-tool.js
```diff
@@ -162,6 +164,9 @@
       };
       var hoverColor = this.state.getStateProperty('drawingToolsSettings').hoverColor;
       var annotations = [];
+      if (this.horizontallyFlipped) {
+        location.x = this.svgOverlay.viewer.tileSources.width - location.x;
```
aeschylus: Here and one other place, we will need to go back after merging image choice to use the canvas coordinate system instead of the tilesources.

js/src/annotations/osd-svg-overlay.js
```diff
@@ -474,6 +487,9 @@
       if (mousePosition.y > this.viewer.tileSources.height) {
         mousePosition.y = this.viewer.tileSources.height;
       }
+      if (this.horizontallyFlipped) {
+        mousePosition.x = this.viewer.tileSources.width - mousePosition.x;
```
aeschylus: See above comment re: canvas width in image choice/detail images

PR #1300, opened 2017-03-01
Fix in thumbnail display

morpheus-87:
This PR fixes a displaying bug in Internet Explorer:

screenshot-ie

As you can see, all of the thumbnail labels except the currently active one get cut off.

Besides, the CSS rules for the bottom panel are reorganized, before some of them were accidentally in the wrong LESS file.


css/mirador.less/panels/bottom-panel.less
```diff
@@ -38,6 +38,12 @@
         padding: 0px 10px 0px 10px;
         /*transition: all 0.03s ease-out;*/
         display: inline-block;
+        margin-bottom: 11px !important;
+
+        &.highlight {
+          padding-left: 8px !important;
+          padding-right: 8px !important;
```
aeschylus: Can we find a way to do this without `!important`.

Some opinions:
https://css-tricks.com/when-using-important-is-the-right-choice/
http://stackoverflow.com/a/3706876/564684

In this case, the specificity is not narrow enough for the change to take hold. If this is how we want to fix the bump up, we should find out why it's happening and reorganise accordingly. Our strategy for attaching the label seems to be what's causing the problem here. The behaviour we probably intended was: "make all thumbnails line up at the top as equal height, and display the label underneath. If the label is too long, truncate it. If there is no label, the thumbnails should still line up at the top, and there should simply be a blank space beneath those without labels.

morpheus-87: These rules containing `!important` came in with the commit 9f23360 by @rsinghal. Is it really necessary, to keep the `!important`? I just tested it without and nothing changed.

rsinghal: 2 1/2 years ago!  I have no idea why I did that :)

aeschylus: Cool. Thanks for the update. The diff didn't make clear that those were pre-existing.

css/mirador.less/panels/bottom-panel.less
```diff
@@ -47,6 +53,21 @@
           display: block;
           box-sizing:border-box;
 
+          &.thumbnail-image {
+            background-color: gray;
+            box-shadow: 0px 0px 5px @black-a60;
+            float:left;
+            margin-right:15px;
+            /*height: 80px;*/
```
mejackreed: Is this commented out code needed?

morpheus-87: I don't know, I did just copy-paste. Shall I remove it?

rsinghal: Probably legacy

mejackreed: Yeah I'd say kill it if its still hanging around 👍

morpheus-87: Ok, then I will remove these lines of CSS.

Issue #1160, opened 2016-10-28
Border/Fill color drop-downs go out of canvas when working in lower resolution


ViktorRibchev:
Border/Fill color drop-downs go out of canvas when working in lower resolution. 
![mirador1](https://cloud.githubusercontent.com/assets/17718464/19795886/8952c640-9ce8-11e6-8f99-340c6d9bbd96.png)
![mirador2](https://cloud.githubusercontent.com/assets/17718464/19795887/8958ce50-9ce8-11e6-8f73-4e75ec60d0f2.png)

Browser resolution: 1280x1024 and lower.

mejackreed:
Would a right justify satisfy this?
![screen shot 2016-11-07 at 10 42 41 am](https://cloud.githubusercontent.com/assets/1656824/20063948/f2e7a77e-a4d6-11e6-99bd-e8cab2e81c80.png)

ViktorRibchev:
@mejackreed Right justify will solve the problem

mejackreed:
Spectrum.js seems to apply inline styles to position this element. Are there patterns that mirador uses to override out of the box plugin functionality? 

Or is `!important` ok to use in this context?

versae:
Is it OK to fix the `left` property, or do we want the widget to be shown aligned to the lower-right corner of the control buttons?
```css
.mirador-osd-color-picker .sp-container {
    left: 133px !important;
```

PR #1301, opened 2017-03-01
Use css left to position the color pickers.

### LINKED ISSUE #1209: Annotation color pickers don't open in second slot
Those issues are also regressions.

Scenario to reproduce.

1. Open [](http://projectmirador.org/demo/)
2. Activate annotations inside slot 2
3. Choose a shape tool inside slot 2
4. Click border/fill color

Result: Nothing happens
Expected result: Color picker

jkeck:
It was previously using jQuery offset which styles the element relative to the document and not the canvas/window element.

This PR also uses outerWidth instead of width to account for element borders.

Closes #1209 

## 2nd Window
2nd-anno-style

## 1st Window
1st-anno-style

_To ensure that we don't break #1160, which is what this code was initially written to fix._


js/src/widgets/contextControls.js
```diff
@@ -200,10 +200,7 @@
       pickerOffset=pickerContainer.offset(),
       windowWidth = this.state.windowsElements[this.windowId].width();
       if (pickerContainer.width() + pickerOffset.left > windowWidth) {
-        pickerContainer.offset({
-          top: pickerOffset.top,
-          left: windowWidth - (pickerContainer.width())
-        });
+        pickerContainer.css('left', windowWidth - (pickerContainer.outerWidth()));
```
jkeck: I'm not sure this is ultimately the best way to position this, but I believe this is doing what the code was initially intending to do when written.

mejackreed: Yep agree here. It's tough because positioning this is difficult. 👍

PR #1304, opened 2017-03-01
[WIP] Fixes #855. Now annotation popups hide if mouse leaves the canva…

### LINKED ISSUE #855: Annotations Viewer remains open after leaving Mirador element
When Mirador is both iframed in and when it's used as an element within a larger place, if you hover over an annotation close to the edge and leave the element/iframe, then the annotation remains opened instead of closing. 

This is an issue when the annotation viewer covers up buttons on any of the corners since the qtip window appears over the controls.

versae:
…s. Shape highlight still remains in some cases.


versae:
Added a new commit. Still not solved completely.
fix-annotations-embedded

js/src/annotations/osd-region-draw-tool.js
```diff
@@ -133,6 +133,31 @@
       });
       this.svgOverlay.paperScope.view.draw();
       _this.eventEmitter.publish('annotationsRendered.' + _this.windowId);
+
+      _this.eventEmitter.subscribe('slotLeave.' + _this.windowId, function(event, eventData) {
```
radpet: The callback seems to be leaking :), please unsubscribe in the destroy function

versae: Sure thing.

PR #1305, opened 2017-03-01
remove zoom factor for rectangle calculation

### LINKED ISSUE #1245: osd-svg-rectangle is not correctly parsing xywh fragment selectors to paperjs segments
@radpet @beaudet

http://dms-data.stanford.edu/data/manifests/Stanford/ege1/manifest.json has a series of annotation lists that are in `xywh` format.  However, in [`_createSegments`](https://github.com/ProjectMirador/mirador/blob/2.1.3/js/src/annotations/osd-svg-rectangle.js#L41) in `osd-svg-rectangle`, the original x and y points are not the resulting points used when creating the segments.  

The first screen shot is what the rectangles look like with a version of Mirador before paperjs (the first canvas in the manifest).
screen shot 2017-02-02 at 2 27 24 pm

The second screen shot is what the rectangles look like in the 2.1.3 branch of Mirador using paperjs.  All of their start points are shifted up and to the left and are now overlapping, which they shouldn't be doing.
screen shot 2017-02-02 at 2 28 50 pm

mejackreed:
Closes #1245 

screen shot 2017-03-01 at 3 24 08 pm

Thanks @jkeck and @radpet !


js/src/annotations/osd-svg-rectangle.js
```diff
@@ -38,7 +37,7 @@
       return shape;
     },
 
-    _createSegments:function(initialPoint,pixel, overlay){
+    _createSegments:function(initialPoint, overlay){
```
mejackreed: I assume changing the arguments for this "private" method is acceptable as a non-api breaking change?

aeschylus: Yes.

PR #1307, opened 2017-03-01
Fixes #1021. Add a 10% of pannable background beyond the image bounda…

### LINKED ISSUE #1021: Adjust location of trash and rotate points when they are near the edge of the canvas
@radpet  If a shape is drawn near the bottom of the image, the trash can icon is not visible.  Likewise, if the shape is drawn near the top of the image, the rotate point is not visible.

screen shot 2016-07-19 at 4 04 41 pm
screen shot 2016-07-19 at 3 54 10 pm

Lastly, the rotate point is not adjusting in size as the user zooms in and out.

versae:
…ries


rsinghal:
Closing because this doesn't fix the issue.  See #1021

PR #1309, opened 2017-03-01
Activate the correct paperScope when a tool is selected

### LINKED ISSUE #1208: Cannot annotate when more than one slot is inside annotation mode
This issue is not presented in 2.1.1 so it seems to be a regression. 

The steps to reproduce are:

Open a new slot (or test inside http://projectmirador.org/demo/ {by default more than 1 slot} )
Open annotation menu on slot 1
Open annotation menu on slot 2

Click a shape tool on slot 1
Click a shape tool on slot 2
Start drawing on slot 1

Result:
The shape seems to be enlarged and the annotation popup is inside slot 2
After save the shape disappears

If you need any screens tell me :)

### LINKED ISSUE #1272: annotation tools not working in latest version of Chrome on OS X
Hi, We've (@mrwho ) been testing out an updated version Mirador and we are having issues with it working in Chrome Version 56.0.2924.87 (64-bit) in OS X. Seems to work fine with Firefox. I have tested our own site as well as the Mirador demo.

http://digitalcollections.frick.org/digico/compare.html#3fdd6967-640e-4b69-88d5-a80feb3b5318

http://projectmirador.org/demo/

### LINKED ISSUE #1298: Bug: Draw annotation in one window/slot to create an annotation in another
After opening two windows and entering into annotation mode in both, you can select the rectangle tool from the left window, draw on the canvas in the right window, and it creates an annotation on the canvas in the left window. 
anno_bug

csbailey5t:
Fixes #1272 fixes #1298 fixes #1208


PpyMohan:
Hi,
Can you tell how you got two window annotations on openseadragon paperjsOverlays

PR #1319, opened 2017-03-03
add thick/thicker line style options

### LINKED ISSUE #1299: Add additional line weights/thicknesses to Line Style drop-down on annotation creation
Desire is to have access to thicker lines for annotation bounding boxes in addition to line, dashes, and dots-and-dashes

drh-stanford:
This PR fixes #1299. It changes how the annotation shapes are rendered by managing a new `shape.data.strokeWidth` variable.

1. It changes the signature of `onHover` to supply the desired (scaled) stroke width.
1. It removes the "default / edit / current" notions of a stroke width.
1. It renames `shape.data.nonHoverStroke` to `shape.data.nonHoverStrokeColor`.
1. It deprecates the `data.newlyCreatedStrokeFactor` support -- I wasn't sure this was useful.
1. I chose solid = 1, thick = 3, and thickest = 6 for the stroke width options.


js/src/annotations/osd-svg-overlay.js
```diff
@@ -339,8 +352,7 @@
         //should remove the styles added for newly created annotation
         for(var i=0;i<_this.draftPaths.length;i++){
           if(_this.draftPaths[i].data && _this.draftPaths[i].data.newlyCreated){
-            _this.draftPaths[i].strokeWidth /= _this.draftPaths[i].data.newlyCreatedStrokeFactor;
-            _this.draftPaths[i].data.currentStrokeValue /= _this.draftPaths[i].data.newlyCreatedStrokeFactor;
```
aeschylus: How is it that in a number of places you are able to remove the strokeValue? Is it just not used in these instances?

js/src/annotations/osd-svg-pin.js
```diff
@@ -82,18 +80,19 @@
         }
       }
     },
-    
-    onHover:function(activate,shape,hoverColor){
+
+    onHover:function(activate,shape,hoverWidth,hoverColor){
+      shape.strokeWidth = hoverWidth;
```
aeschylus: Not a reflection on your solution, but I wonder why onHover needs any of these arguments at all. Presumably these are inherent configuration properties that could be passed down to the object on instantiation. ¯\_(ツ)_/¯

radpet: The osd svg pin is rather service that operates with shapes it is not the shape itself

Issue #1325, opened 2017-03-06
2.3.0 branch has problem deleting annotations.


aeschylus:
![mirador delete anno bug](https://cloud.githubusercontent.com/assets/145874/23618244/9b1007aa-0244-11e7-8277-e9fbe30022cc.gif)
Haven't done a `git bisect` yet, but at some point in 2.3.0 we created an issue where clicking delete on an annotation popup causes the modal background of the delete confirmation box to be placed on top of the modal itself, making it impossible to click and locking the user in that state.

drh-stanford:
the display of the DialogBuilder is working but it goes inactive immediately afterward, so the callback is never triggered -- https://github.com/ProjectMirador/mirador/blob/2.3.0/js/src/annotations/annotationTooltip.js#L209-L218

drh-stanford:
verified that deleting annotations works in v2.1.3, 2.2.0, and v2.2.0.

drh-stanford:
looks like the bug was introduced in 61213d1

drh-stanford:
i need to pair with someone on this to make further progress...

i don't know the provenance of `css/bootstrap-modal.css` which is where the opacity is being set and the z-index looks like a workaround to bootstrap -- see https://github.com/ProjectMirador/mirador/blob/2.3.0/js/src/utils/dialog-builder.js#L19-L26

csbailey5t:
I'm not sure of a final cause or fix yet, but this seems to be happening because the `shown.bs.modal` event in https://github.com/ProjectMirador/mirador/blob/2.3.0/js/src/utils/dialog-builder.js#L19 is not being triggered. If you go through the delete process and get to where the fade in blocks access to the modal footer, where the "yes" and "no" buttons are, you can trigger the event as written in the code through the chrome dev console and get the correct behavior. (`$('div.bootbox.modal.fade.in').trigger("shown.bs.modal");`)

PR #1331, opened 2017-03-10
Change shown event to use bootbox init because shown.bs.modal is not fired

radpet:
@aeschylus @csbailey5t @drh-stanford
The shown.bs.modal event is not fired because the transition.js required by modal.js to work properly is missing from the gruntfile (and from the build), however using the booxbot init function it works properly.Anyway I added the transition.js to the build and make use of the init hook.

Fixes https://github.com/ProjectMirador/mirador/issues/1325


mejackreed:
Nice catch!

PR #1334, opened 2017-03-13
Search within

versae:
Integrates @jeffreycwitt's [fork](https://github.com/jeffreycwitt/mirador/tree/feature/2.1-search-within) (not preserving commit history) and it's branched off @aeschylus' recent work on manifest layouts.

Behavior can be tested with these two manifests:
 - ~~http://scta.info/iiif/pl-zbsSII72/manifest~~ (not IIIF compliant)
 - http://wellcomelibrary.org/iiif/b18035978/manifest
 - http://scta.info/iiif/plaoulcommentary/svict/manifest
 - http://scta.info/iiif/plaoulcommentary/sorb/manifest
 - http://scta.info/iiif/plaoulcommentary/reims/manifest

It's still incomplete as the code contains some non working parts left as TODO.


versae:
It doesn't highlight the annotation but it navigates to it in the correct canvas.

search-within-take-to-anno

versae:
There is no way to link each match to its annotation according to the [specification](http://iiif.io/api/search/1.0/#multi-annotations-hits), so when there is more than one annotation linked to a search result, links to the annotations are shown. This links do not change the canvas but rather only the bounds to focus.

screen shot 2017-03-22 at 3 37 12 pm

spec/widgets/searchWithin.test.js
```diff
@@ -122,5 +122,55 @@ describe('Search tab', function() {
...
+        "canvasID": canvasID,
+        "bounds": bounds
+      });
+    });
+
+    it('should not change canvas after clicking on a second annotation linked to a search result', function() {
+      this.sandbox.find('#search-within-form input.js-query').val('found');
+      this.sandbox.find('#search-within-form').trigger('submit');
+      var anno1 = this.sandbox.find('.search-annotation').first(),
+          anno2 = this.sandbox.find('.search-annotation:nth-child(2)').first(),
```
radpet: https://github.com/airbnb/javascript#variables just sharing :)

versae: Yep, once we merge #1337 and adopt Airbnb coding style it's going to be fun times :)

jeffreycwitt:
I think it would be really useful to add a boundary box around annotated target of a search result. The zoom in is great, but it still leaves a person looking for the search result. 

Something like this: https://www.youtube.com/watch?v=8V9QIMwFEJs (sorry about the music in the background of this video)

versae:
I agree. But it would be important not to confuse the bounding box of a search result with an annotation, they don't always match. Maybe a fadein-fadeout effect would be nice. We can add an issue for this to discuss further.

jeffreycwitt:
Here are some good scta manifest for testing:

http://scta.info/iiif/plaoulcommentary/svict/manifest
http://scta.info/iiif/plaoulcommentary/sorb/manifest
http://scta.info/iiif/plaoulcommentary/reims/manifest

The search service for these manifests returns annotations (not HITS), and coordinate regions for the paragraph containing the search result. Yellow highlighting comes from my search service not mirador. 

Things seems to work pretty well. I encountered a few bugs, as I clicked through. But sometimes that could be due to bad data.

jeffreycwitt:
I'm noticing a small bug if an image fails to load, an error occurs and freezes the entire system. 

In this case, after selecting an search result (canvas 14r), Mirador navigates. The request for the image was failing (a problem on my end, which I fixed now). But this was triggering the error below and freezing up the system. 
 
GET http://loris2.scta.info/reims/R14v.jpg/full/58,/0/default.jpg 500 (INTERNAL SERVER ERROR)
mirador.js:42143 image failed to load: http://loris2.scta.info/reims/R14v.jpg/full/58,/0/default.jpg
mirador.js:33821 Uncaught TypeError: Cannot read property 'source' of undefined
    at $.Overlay.resize (mirador.js:33821)
    at Object._thisRotate [as handler] (mirador.js:33306)
    at mirador.js:7094
    at $.Viewer.raiseEvent (mirador.js:7116)
    at $.Viewport.setRotation (mirador.js:22158)
    at resetImageManipulationControls (mirador.js:38526)
    at Object.<anonymous> (mirador.js:38677)
    at Object.dispatch (mirador.js:5)
    at Object.q.handle (mirador.js:5)
    at Object.trigger (mirador.js:6)

The error was triggered by an error on my image server. But it would be nice for the system not to freeze up.

versae:
That's interesting. I have talked to @aeschylus about how lenient Mirador needs to be. This kind of checks belong to that ongoing conversation. I would say that Mirador lacks a proper user feedback mechanism and without a standardized way to inform the user any decision we make here might be arbitrary. That being said, my personal opinion is to avoid the app from crashing/not working because an error on the image server. Could you please provide a failing manifest and search term so I can quickly reproduce?

Issue #1318, opened 2017-03-02
Proposal: Format/Write all code by given convention


radpet:
I suggest discussing the format rules and committing them inside mirador using http://editorconfig.org/ and following https://github.com/airbnb/javascript as base model of writing code

mejackreed:
👍  I've enjoyed working with Airbnb style guides. I would maybe say could we use a tool that works with a given Airbnb config? https://github.com/airbnb/javascript#resources

Perhaps jshint or eslint? Mirador is already configured with jshint?

Also I'm assuming the es5 configuration? https://github.com/airbnb/javascript/blob/master/linters/.jshintrc

aeschylus:
https://github.com/ProjectMirador/mirador/issues/514#issuecomment-92476971

At the time JSCS looked like a good tool, since it would actually force the code into a certain style (rather than producing style warnings during a PR). At one point we had jshint preventing pushes using a githook, but that ended up being problematic and was disabled.

Very eager to revisit this now that we have more people being harmed by style issues. 👍

sdellis:
I don't know of any tool that will just flag new additions/edits. I personally really like [StandardJS](http://standardjs.com/) for maintaining consistency in style. No configuration (so you have to be ok with the rules) and you catch the style errors before they are submitted as PRs. 

I haven't touched the Mirador code in a while, so it's up to the people who work on it the most, but just throwing it out as something you should check out that would be pretty easy to set up and maintain.

mejackreed:
We have used HoundCi to do only changed line style checking, but it can be a bit noisy on GitHub comments.

csbailey5t:
I think eslint is a great tool combined with the Airbnb plugin. I'm also not averse to strictly enforced style automatically done by https://github.com/prettier/prettier.

@mejackreed I think keeping some sense of the historic character of the code is great, but that will always exist in the git history. In theory, I'm inclined to think that enforced standards for the whole codebase may be worth it. In practice, using eslint with the airbnb-es5 rules (by way of https://www.npmjs.com/package/eslint-config-airbnb-es5) there are `6972 problems (5753 errors, 1219 warnings)` in `js/src`. Eslint can autofix some of these, resulting in: `2234 problems (1023 errors, 1211 warnings)`. 

That likely means that conforming the whole codebase to those guidelines is out of reach. You can use the `--cache` option to only check changed files, though, which would allow the community to gradually conform the codebase over time. 

[...]

PR #1337, opened 2017-03-16
Adds in Airbnb style via .eslintrc

mejackreed:
This PR starts to move towards Airbnb styleguide conformance without changing anything too drastic. Since the Airbnb styleguide is maintained via an eslint package, this PR adds that in.

It also updates the CONTRIBUTING.md to discuss what is going on currently and hopefully how we can get there.

Starts to address #1318 

Thoughts? @radpet @csbailey5t @sdellis @aeschylus @rsinghal ?

Our previous .jshintrc didn't really conform to AirBnb style, and the canonical expression of that style is now in eslint.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -5,5 +5,13 @@
     "jQuery": true
   },
   "rules": {
+    "no-param-reassign": [
+      2, {
+        "props": true,
+        "ignorePropertyModificationsFor": [
+          "$"
+        ]
+      }
+    ]
   }
 }
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,9 @@
+{
+  "extends": "airbnb-base/legacy",
+  "globals": {
+    "Mirador": true,
+    "jQuery": true
+  },
+  "rules": {
+  }
+}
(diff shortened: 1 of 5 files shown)


sdellis:
Awesome, @mejackreed ! Should there be an `.editorconfig` file here as well for editor integration, or is that something we add ourselves?

radpet:
@sdellis I think most ide look for .eslintrc automatically

mejackreed:
@sdellis /shrug.. I'm not sure. `.editorconfig` looks like it handles only really spacing issues https://github.com/airbnb/javascript/blob/master/.editorconfig . That could be done, but I think thats more of an add-on or additional thing.

csbailey5t:
@sdellis If you've installed eslint into your editor (typically by way of a plugin), it should pick up on the .eslintrc file in the project. But, you do need the linter package. For instance, in Atom, you would need the `linter` package and the `linter-eslint` package.

csbailey5t:
@mejackreed `grunt-contrib-eslint` seems to be deprecated in favor of `gruntify-eslint`, which is written by the same developer and seems to just be an updated version. 

`npm WARN deprecated grunt-contrib-eslint@0.0.5: Please use gruntify-eslint`

Perhaps we should switch to that in this PR?

mejackreed:
Great catch @csbailey5t . I've updated the PR to reflect the correct version.

PR #1352, opened 2017-03-21
adds a script that updates the demo site

mejackreed:
@aeschylus @rsinghal the idea behind this was to make the release process a bit easier. 

We can even automate the build uploading to the github api, but that will take some more work.


mejackreed:
Also now adds example.html used in build release and script to automate creating releases. The grunt task doesn't fully do this.

PR #1367, opened 2017-04-04
Fix rendering of within fields with an array value

jbaiter:
This PR fixes a bug I introduced in #1234 : When the `within` attribute of the Manifest contained an array of strings, the `getWithin` function would recursively call itself to render the individual entries.
However, I forgot to add the `this` parameter to the `map` call, resulting in an error when such a Manifest would be loaded.

An example of such a manifest can be found here: http://digi.ub.uni-heidelberg.de/diglit/iiif/cpg23/manifest.json

### DIFF
--- a/js/src/widgets/metadataView.js
+++ b/js/src/widgets/metadataView.js
@@ -169,7 +169,7 @@
       var collectionLabel = within.label || collectionUrl;
       return '<a href="' + collectionUrl + '" target="_blank">' + collectionLabel + '</a>';
      } else if (within instanceof Array) {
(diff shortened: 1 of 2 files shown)


aeschylus:
Thanks @jbaiter. I do plan to help set up a test for the dynamic language updating feature soon, but trying to finish up another feature at the moment. 

Thanks for the bug fix.

morpheus-87:
Thanks for your comment about the dynamic language, @aeschylus. But it would be better placed at #1236, isn't it? ;-)

Issue #1374, opened 2017-04-06
Fullscreen support is incomplete for IE 11


c-alers:
You can use the arrow icon to open the Mirador viewer into "fullscreen", but the arrows do not work take the viewer out of "fullscreen" (you have to use the ESC key). Also, the Mirador.fullscreenElement function does not properly return the element in IE 11. This turns out to be a problem between IE 11 vs edge. In IE11, the exit function is called "msExitFullscreen" and the fullscreen element is called "msFullscreenElement". Updating utils.js resolves this issue.

camillevilla:
This ticket is being closed as part of the Mirador 3 issues review. We encourage all Mirador 2 implementers to try out Mirador 3 and leave feedback in new tickets.

PR #1375, opened 2017-04-06
Fullscreen support for IE 11

c-alers:
The exitFullscreen and fullscreenElement functions were not working for
IE browsers not set to "edge" condition. The methods have been updated
to support non "edge" IE 11 browsers.  #1374


aeschylus:
@c-alers Can we convince you to add a test for this feature? For example, a test that uses isFullScreen/webkitIsFullScreen/msIsFullscreen to see if the operation was carried off successfully?

for context, here are the MDN docs with various prefixes for the property to confirm: 
https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API#Prefixing

```` javascript
var isInFullScreen = document.fullScreenElement || document.mozFullScreen || document.webkitIsFullScreen || document.msIsFullScreen;
````

I think the right place would be in [`mirador.test.js`](https://github.com/ProjectMirador/mirador/blob/develop/spec/mirador.test.js) (as an integration test), or alternately, in [`utils.test.js`](https://github.com/ProjectMirador/mirador/blob/develop/spec/utils.test.js).

c-alers:
Np. I'd be happy to add a test. I haven't worked with Karma before so I'm reading up on the documentation. I should be able to add the test by middle of next week (day job permitting).

aeschylus:

PR #1388, opened 2017-04-13
change dev process description

aeschylus:
Change release process to reference develop instead of pre-named release branches.


csbailey5t:
This replicates PR #1387

PR #1393, opened 2017-04-19
Detail images and image choice

aeschylus:


aeschylus:
Checklist for reference of the test that are being worked on now:
### window
- [ ] eventedCanvases are generated on initialisation
- [ ] changing the canvasID changes the focused canvasModel
- [ ] changing aspects of the eventedCanvas cause the interior eventEmitter supplied to each eventedCanvas to trigger an event in the global Mirador eventEmitter
### sidePanel
- [ ] layersTab is initialised or not depending on the configuration given
### layersTab
- [ ] tab shows/hides when it is navigated to/from
- [ ] list items render from fixtures
- [ ] events are picked up and rendering changes occur when a mock eventedCanvas is changed
- [ ] loading states have correct stylings applied
### imageView
- [ ] "item" is present in openseadragon world coordinates
- [ ] "item" changes when the page is turned
- [ ] there are multiple items when there are multiple details or choices turned on
- [ ] event handlers are triggered when events are fired

js/src/workspaces/window.js
```diff
@@ -762,22 +822,28 @@
...
+
+      _this.canvasID = canvasID;
+
+      // TODO: Completely remove the 'removeTooltips' event. They are not actions and
+      // they do not represent any state change. They are essentially
+      // function calls, which should be handled as responses to the
+      // publication of the state update event currentCanvasIDUpdated
+      // below.
       _this.eventEmitter.publish('removeTooltips.' + _this.id);
-      _this.eventEmitter.unsubscribe(('annotationListLoaded.' + _this.id));
```
aeschylus: Because the annotationsLayer is only being initialised once for the window, it will not re-subscribe itself, so this is no longer required. In the initialisation sequence, this event is actually called _after_ the subscription is bound for the first time, so this was causing the event to never be called.

js/src/settings.js
```diff
@@ -54,8 +54,8 @@
         "toc" : true,
         "annotations" : false,
         "tocTabAvailable": true,
-        // "layersTabAvailable": true,
-        "searchTabAvailable": false,
+        "layersTabAvailable": true,
```
rsinghal: This should be set to `false` by default

js/src/widgets/layersTab.js
```diff
@@ -62,6 +64,32 @@
       this.localState(localState);
```
rsinghal: Also, before merging, the layers tab should not be visible to the user for non-ImageView modules.

js/src/widgets/imageView.js
```diff
@@ -528,117 +634,104 @@
...
-          'tileSources':  infoJson,
-          'uniqueID' : uniqueID
-        });
+      _this.osd = OpenSeadragon({
+        id: osdID,
+        uniqueID: osdID,
+        preserveViewport: true,
+        blendTime: 0.1,
+        alwaysBlend: false,
+        showNavigationControl: false
```
aeschylus: This is using the OpenSeadragon object directly, since the way it is used differs somewhat from the base version. These can be synchronised and we bring the other views into the same logic.

rsinghal: Is there a reason _not_ to use the Mirador OSD module, since all it does is return the osd object anyway?  I would rather we keep that part consistent so if someone needs to come in and add a configuration option that gets passed to OSD, there is a singular place for it, instead of (now) two places.

index.html
```diff
@@ -68,10 +68,19 @@
            { "manifestUri": "http://dzkimgs.l.u-tokyo.ac.jp/iiif/zuzoubu/12b02/manifest.json", "location": "University of Tokyo"},
            { "manifestUri": "http://www2.dhii.jp/nijl/NIJL0018/099-0014/manifest_tags.json", "location": "NIJL"},
            { "manifestUri": "http://digi.vatlib.it/iiif/MSS_Vat.lat.3225/manifest.json", "location": "Vatican Library"},
+           { "manifestUri": "http://media.nga.gov/public/manifests/nga_highlights.json", "location": "National Gallery of Art"},
```
rsinghal: Duplicate manifest - delete

js/src/widgets/imageView.js
aeschylus: @rsinghal Addressed in latest commit.

aeschylus:
Clicking "merge" now, given review status and permission/availability issues.

PR #1396, opened 2017-04-20
Hide layout button when window is in fullscreen mode

### LINKED ISSUE #751: Hide buttons during Window fullscreen mode
One of the scenarios for buttons from the slots menu (add slot left, right ...) is inappropriate.
Scenario 1: When we use fullscreen mode from the main menu we call `document.fullscreenElement` and the whole document is visible is fullscreen mode. In this case, inserion of new slot is within the visible (visible in fullscreen mode) area.
Scenario 2: When we use the OSD fullscreen mode we call `OpenSeadragon.requestFullScreen(_this.parent.parent.element[0])` and only current OSD within the focused slot is visible. Buttons from the slots menu add new slot but it is outside the visible element and it is visible only if we exit the fullscreen mode.

regisrob:
Fixes #751. Only applies to OSD fullscreen mode


aeschylus:
@regisrob This looks like a good call. Can you add a quick test for this at some point that checks if the button is hidden in fullscreen mode?

regisrob:
@aeschylus ok, I will try to do that tomorrow ;)

regisrob:
We tried to write a test for that a while ago but the test was still failing:
```
describe('Window FullScreen Button', function () {
      var utils = Mirador;

      it('should hide mirador window menu buttons', function () {
        utils.enterFullscreen(this.window.element[0]);
        expect(this.appendTo.find('.mirador-icon-window-menu').css('display')).toBe('none');
      })
      
      it('should show mirador window menu buttons', function () {
        utils.exitFullscreen();
        expect(this.appendTo.find('.mirador-icon-window-menu').css('display')).toBe('block');
      })
    })
```

The error was:
```
PhantomJS 2.1.1 (Mac OS X 0.0.0) Window Basic Operation Window FullScreen Button should hide mirador window menu buttons FAILED
    Expected '' to be 'none'.
    spec/workspaces/window.test.js:137:84
PhantomJS 2.1.1 (Mac OS X 0.0.0) Window Basic Operation Window FullScreen Button should show mirador window menu buttons FAILED
    Expected '' to be 'block'.
    spec/workspaces/window.test.js:142:84`
```

aeschylus:

PR #1402, opened 2017-05-02
Use json datatype when getting annotation lists

ojlyytinen:
Use jQuery.getJSON instead of jQuery.get to retrieve annotation lists. This forces the response to be parsed as JSON. Otherwise getAnnotations won't work if the http response headers don't specify application/json content-type. Ideally the content-type should of course be set properly but it doesn't hurt to make Mirador a bit more tolerant. And it appears that getJSON is used everywhere else already.


ojlyytinen:
Yes, the main thing it does is it forces jQuery to interpret the response as JSON and turn it into a JSON object rather than guessing what to do with it based on the content-type header. The way it was before, the response would get turned into a JSON object only if the response headers specify content-type application/json, otherwise `list` would just be an unparsed String. The rest of the getAnnotations function assumes that the response will be a JSON object so if the server doesn't add the right content-type header then `list.resources` won't work.

We publish our IIIF as just a bunch of static files on a web server. Right now Mirador loads everything else fine, except the annotations. That's because the rest of Mirador code already seems to be using getJSON rather than get.

PR #1404, opened 2017-05-04
Test contributions to Mirador

dicksonlaw583:
Here is a number of additions to the unit test suite, covering the following areas:

- Utilities
- Tabs
- Some OSD elements
- Some windows elements
- Annotation endpoints (except CATCH)

According to my local runs, this should raise coverage by at least 10%.

Please edit any tests that you find inappropriate for upcoming Mirador developments. Individual categories of tests can also be merged from our repositories if you see fit. They start with `development_test_contrib_`.


mejackreed:
I started to look through these with @dicksonlaw583 this morning. Overall looks great, but would like a second set of eyes on it. I'm not sure about the flux with layers tab and annotations tab, and "CATCH".

aeschylus:
I don't think there should be too much rework visa-vis layers, etc., and it will be useful to have the tests there for that purpose anyway.

Discussing the catch endpoint on the dev standup this morning, we would like to move that out of core into a plugin (since it already is a plugin). Catch is Harvard's annotation server, and is still in the codebase here for historical reasons, as our first implementation of an asynchronous endpoint.

PR #1408, opened 2017-05-08
Add identifiers to the rights and links metadata items

morpheus-87:
This PR adds static identifiers to the rights and links metadata items, that are used as classname in the rendering step. This can be useful for styling these sections in the metadata sidepanel via easy CSS rules:

```css
.metadata-item.within {
  color: red;
}
.metadata-item.license > .metadata-value {
  font-weight: bold;
}
```

### DIFF
--- a/js/src/widgets/metadataView.js
+++ b/js/src/widgets/metadataView.js
@@ -38,6 +38,7 @@
 
           if (typeof itm.value === 'string' && itm.value !== '') {
             tplData[metadataKey].push({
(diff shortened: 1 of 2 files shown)


aeschylus:
Thanks @morpheus-87. Looks good to me. In light of recent discussions regarding Mirador's API and feature additions, I've added a link to this PR to #1412, so we can talk about how to best support this through the API or a plugin paradigm. This is generic enough that it should probably have been there originally, so I'll merge this.

PR #1361, opened 2017-03-28
Conservation space changes contributed back

radpet:
Properly add and destroy annotation tooltip events and try avoid memory leaks

@rsinghal @aeschylus @beaudet

This is a fix for https://github.com/ProjectMirador/mirador/issues/1320 plus the start of aligning ConservationSpace fork and official mirador repo.

Issue #1320, opened 2017-03-03
Horizontal grey lines showing up periodically.


aeschylus:
To reproduce: create an annotation, but cancel it without saving. Move the mouse without going outside the window or hovering over any exiting annotation regions. You will see a horizontal gray line. Inspecting this in the developer tools reveals it to be the tooltip for the canceled annotation.

sdellis:
I just ran into this today on the demo on the ProjectMirador.org site. This is definitely tooltip "residue", but I got to it by different means. 

1. Go to the demo. 
2. Close one of the slots.
3. Click the annotation icon in the remaining slot.
4. Move over the image and you will see the gray line appear.
5. It prevents me from then drawing a target with any of the shape tools.

gigamorph:
Appears every time I enable the annotation overlay and the mouse cursor enters the canvas. Disappears when I hover over an existing annotation. I added the screenshots in the wrong ticket: https://github.com/ProjectMirador/mirador/issues/1328

radpet:
Thats qtip element.. Check it out in the dom tree. I think the ConsSpace implementation does not have this problem because it has few lines that are different from the mirador codebase, I can contribute them if you want :)

camillevilla:
This ticket is being closed as part of the Mirador 3 issues review. We encourage all Mirador 2 implementers to try out Mirador 3 and leave feedback in new tickets.

PR #1411, opened 2017-05-09
Band-aid for hiding the gray line in #1320 from view

dicksonlaw583:
Since PR #1361 is now stagnant due to scope issues, I'm pushing up a small patch that hides the unwanted gray line in #1320 until the underlying cause is addressed. Some use cases such as manuscript studies are sensitive to visual noise and would benefit from the patch in the short term.

### DIFF
--- a/css/mirador.less/overrides/qtip.less
+++ b/css/mirador.less/overrides/qtip.less
@@ -1,5 +1,6 @@
 .mirador-container {
   .qtip {
+      border: none;
   }
 
   .qtip-bootstrap {


aeschylus:
Thanks @dicksonlaw583. This is a good gesture, but I think we should address the underlying cause. What change was introduced in #1361 that resolved this issue?

dicksonlaw583:
This commit resolves it: 

https://github.com/ProjectMirador/mirador/pull/1361/commits/63f0761e2f2178ae10b3be8e2fffa5ac030a99e4

However, it is a squashed commit containing the ruler in addition to the fix, so we need to pick apart the commit manually. How would you prefer to start?

PR #1413, opened 2017-05-16
Fix the gray bar in #1320

dicksonlaw583:
This is a follow-up to PR #1411 that addresses #1320 from the source. It contains the specific fragment in PR #1361 responsible for removing the gray bar and better handling of events.
 
Credit: Conservation Space, @radpet


aeschylus:
Thanks for working to extract this Dixon. This is great. I know @gigamorph and all others annoyed by the grey lines will be happy. And thanks @radpet for the former PR.

PR #1417, opened 2017-05-19
Show error modal when manifest can't be loaded

jbaiter:
Previously Mirador would just not show the manifest without notifying the user about what was wrong. Also includes a special message for the frustrating mixedContent cases, where manifests served via HTTP cannot be loaded into Mirador instances served via HTTPS.

A demo is hosted via HTTPS at https://jbaiter.de/mirador
If you try loading a Manifest from e.g [e-codices](http://e-codices.ch/metadata/iiif/zbs-SII-0043/manifest.json) you can see the error message for mixed content.


jbaiter:
Any updates on this? I think it would be a very worthwhile addition to Mirador, especially considering the current behaviour of leaving the user in the dark about what went wrong.

rsinghal:
I think this makes a lot of sense.  Not having some tests is a bit problematic (but I also see that `viewer.test.js` is not populated at all, at the moment).  @aeschylus Would you be able to help stub out a test or two before we approve and merge?

mejackreed:

aeschylus:

PR #1421, opened 2017-05-24
Show a manifest button in every window

morpheus-87:
This PR adds the possibility to show a manifest button, that opens the manifest url in a new tab. Configuration is possible either in `windowObjects` or `windowSettings` with the key `showManifestButton`, default is false.

### DIFF
--- a/css/less/global.less
+++ b/css/less/global.less
@@ -113,7 +113,8 @@
 
   .mirador-viewer a.mirador-icon-view-type,
   .mirador-viewer a.mirador-icon-metadata-view,
-  .mirador-viewer a.mirador-osd-fullscreen  {
+  .mirador-viewer a.mirador-osd-fullscreen,
+  .mirador-viewer a.mirador-icon-manifest-link {
     background: none;
     color: @gray-91;
     box-sizing: border-box;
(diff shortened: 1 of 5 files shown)


aeschylus:
We should bring this up on a call. A lot of people have asked for this feature, but implementing it as a button clashes somewhat with the attempts to avoid what previous design reviews have termed "button proliferation". There may be another location to expose this information, or it may make sense to implement this as a plugin that uses the existing "userButton" configuration options that allow the developer to add any button they like at runtime.

morpheus-87:
Ok, what do you think would be another location for this link, @aeschylus? Somewhere in the metadata?

Anyway, I implemented the idea as [plugin](https://github.com/dbmdz/mirador-plugins/tree/master/ManifestButton).

regisrob:
@aeschylus @morpheus-87 I think it would be a good idea to bring this up on a future call. It might be part of a larger discussion on the need or usefulness to better expose IIIF resource URIs in Mirador (Manifest in the first place, and also Canvas/Image URIs...).

See this issue in UV: https://github.com/UniversalViewer/universalviewer/issues/439
based on this IIIF story: https://github.com/IIIF/iiif-stories/issues/74 (where the idea of recommended UI/UX patterns for this kind of thing has surfaced)

morpheus-87:
With #1461 the display of the manifest link was implemented in the metadata view, so I close this PR. Who ever prefers a manifest button, can still use [this plugin](https://github.com/dbmdz/mirador-plugins/tree/master/ManifestButton).

PR #1449, opened 2017-06-19
Fix recursion errors

aeschylus:
This uses the solution @Wedjaa provided here, with his tests, https://github.com/ProjectMirador/mirador/pull/1445, but replaces Set with array. At some point we need to upgrade our toolchain to let contributions with es6 come in.

Travis was failing because Phantom doesn't handle Set by default (nor do most browsers).

### DIFF
--- a/js/src/utils/saveController.js
+++ b/js/src/utils/saveController.js
@@ -72,6 +72,7 @@
     this.init(newConfig);
   };
 
+
   $.SaveController.prototype = {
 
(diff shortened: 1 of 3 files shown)


aeschylus:
@rsinghal or others, can I get a quick review when you get a chance (and Travis is finished)?

aeschylus:
Merging because the coverage decrease is not significant and it has been approved/reviewed.

PR #1456, opened 2017-06-29
adding padding around new button

jeffreycwitt:
The button needs a bit more padding on the right and left :)


aeschylus:
_Danke_ @jeffreycwitt

PR #1460, opened 2017-07-18
Minor UI updates

regisrob:
This PR is a first pass at enhancing a few things in the Mirador UI, it is not comprehensive at all and of course it does not stand in for a complete design review. Basically it just fullfills the immediate needs that I have for a Mirador instance I am currently working on. It includes:
- small changes to improve readability/clearness of the metadata panel (css only)
- various style improvements to the side panel, especially: 
  - tabs
  - ToC (adjust font sizes, get rid of bold on hover...)
  - search form and results (still need some work)
- new translated strings in the side panel, and updates to fr/en locales


regisrob:
Ok no prob @rsinghal, I will look at it on Monday ;)

regisrob:
@rsinghal I took a look at searchWithin.test.js and I have no idea why the test is failing :confused: It seems to me that the small changes I made in searchWithinResults.js have nothing to do with this test... but I may be wrong since I am not familiar with JS tests at all. The rest of the changes is only on css files.

js/src/widgets/searchWithinResults.js
```diff
@@ -437,14 +437,12 @@ $.SearchWithinResults.prototype = {
             '{{canvaslabel}}',
           '</a>',
           '{{#if annotations}}',
-            '<div>',
```
rsinghal: Removing this `div` is breaking the test. So either it needs to remain, or the test needs to be fixed.  I didn't write any of this code, so I can't be certain why it needs this element, but my guess is the test rely on a certain structure.  @aeschylus may have more insight

regisrob: Ok, looks quite weird for me, I have put the 'div' back in the searchWithinResults.js template block, to pass the test.

PR #1464, opened 2017-08-10
[WIP] sidebar annotations panel to display annotations

jeffreycwitt:
I turned on primitive use of sidebar annotations panel to display annotations with and without coordinate; slide bug in tag selection event, but basic functionality seems to work; still needs tests

This was based off the 2.4 branch. 

While this doesn't work perfectly yet, it works well enough. I'm not able to test it with annotations coming from and storage endpoint, so I don't know yet if it works with these kinds of annotations.

Nevertheless, it is an improvement on what we have, and it touches on almost no other code. The only real change is the annotationsTab. Likewise, because it has to be turned on in the configuration settings, no one who doesn't want it will even know it is there.

Being able to turn it on might alert people to its existence and inspire them to help make it better as well

mirador-annotation-side-panel

### DIFF
--- a/css/mirador.less/panels/side-panel.less
+++ b/css/mirador.less/panels/side-panel.less
@@ -43,6 +43,13 @@
       margin: 0 -2px;
     }
 
+    .annotationListItemAlt {
+      list-style-type: none;
+      border-left:0;
+      border-right:0;
+      padding:10px 10px 10px 10px;
+    }
+
     .annotationListItem {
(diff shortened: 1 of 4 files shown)


jeffreycwitt:
There are some spaces accidentally removed in sidePanel.js. But there are no actual changes here. So all changes here can be completely ignored.

jeffreycwitt:
This is replaced by the newer PR #1475

PR #1486, opened 2017-10-12
Improvement for the annotation deletion confirm dialog

morpheus-87:
This PR adds the confirm dialog, when deleting an annotation, to the document body instead of each window.


aeschylus:
I'm closing this for now because we want to avoid adding objects to the document. Calling `document` can interfere with a containing application, and since the annotation cancellation pertains to the annotating state of a single specific window. I'm happy to re-open if more context can be provided, but I need to clear out the queue here while I have a minute (lots of long-standing requests!).

morpheus-87:
The intention of this PR was, to avoid issues with the `z-index` - I made a screencast to show you what I mean.

out

PR #1498, opened 2017-10-26
Removes final bower dependency in favor of NPM

mejackreed:
Also removes some libraries that were checked in but not being used.


aeschylus:
Tested locally and seems to work without any noticeable change, despite upgrade from unknown version of qTip2 to known, pretty recent version.

aeschylus:
Ignoring coveralls, since changing package manifests should not have any bearing on code reporting.

PR #1500, opened 2017-10-27
Update README.md

alanyee:
Add links

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,11 +12,11 @@
 Mirador uses [Node.js](https://nodejs.org/) and a build system to assemble, test, and manage the development resources. If you have never used these tools before, you may need to install them.
 
  1. Install [Node.js](https://nodejs.org/)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -14,9 +14,9 @@ Mirador uses [Node.js](https://nodejs.org/) and a build system to assemble, test
  1. Install [Node.js](https://nodejs.org/)
  2. Install the Grunt command line runner i.e. `npm install -g grunt-cli`
  3. Install the Bower command line utility i.e. `npm install -g bower`
- 1. Clone the Mirador repository
- 1. Change into the Mirador directory
- 1. Install all dependencies with `npm install`. Run `npm start`.
+ 4. Clone the Mirador repository
```
mejackreed: Can we leave these all as 1's please? That way if one above gets updated they all don't need to be reordered?

alanyee:
@mejackreed the coverage went down a bit. Is that okay with you?

mejackreed:

PR #1505, opened 2017-11-15
fixes an issue where SAS#update was referencing the incorrect URI

mejackreed:
I would like a review from @glenrobson but locally this is working for me. Previously I was getting a 404 from a deployed version of SAS, but unsure of the version.

cc @eefahy

### DIFF
--- a/js/src/annotations/simpleASEndpoint.js
+++ b/js/src/annotations/simpleASEndpoint.js
@@ -151,7 +151,7 @@ $.SimpleASEndpoint = function (options) {
       delete annotation.fullId;
       delete annotation.endpoint;
       jQuery.ajax({
-        url: _this.url + '/update/' + encodeURIComponent(annotationID) + '?APIKey=' + _this.APIKey, // this.prefix+
+        url: _this.url + '/update?APIKey=' + _this.APIKey, // this.prefix+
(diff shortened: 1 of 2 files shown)


glenrobson:
Hi Jack,

I think this could be related to:

https://stackoverflow.com/questions/36220740/why-is-this-url-returning-error-400-from-tomcat

as the current simpleASendpoint.js passes in the annotation uri at the end of `/update/`. I think by changing it to a `uri` param it avoids the 400 error but as the update code of SAS doesn't use the uri parameter it would be safer just to remove:

```encodeURIComponent(annotationID)```

from the update URL. Does that make sense?

Cheers

Glen

glenrobson:
I should just clarify I would remove the id from the update uri so:

```url: _this.url + '/update?uri=' + encodeURIComponent(annotationID) + '?APIKey=' + _this.APIKey, ```

becomes:

```url: _this.url + '/update?APIKey=' + _this.APIKey, ```

It may need a slash at the end of update: ```/update/?APIKey=' + _this.APIKey,```. I can test later if you don't manage to test it earlier.

mejackreed:
Thanks for the feedback @glenrobson . I've updated the PR and tested it successfully with your suggestions.

Issue #1506, opened 2017-11-18
Supporting 1-level collection on initial load vs collection tree view


jeffreycwitt:
I wanted to change a behavior change in Mirador after the merge of @dicksonlaw583 collection tree PR #1405

Any collection now loaded in the config data property with "collectionUri" is now getting passed into the collectionManifestPanel instead of the default manifestPanel.

The relevant changed block of code is here:

https://github.com/ProjectMirador/mirador/pull/1405/files#diff-78dd9471e983fbeee0c86bfdeb827128L282

I wouldn't exactly call this is bug, but it is a significant behavior change. If anyone has relied on passing URL of collections to Mirador, Mirador will no longer load these collections as a list of Manifests. 

Instead the collectionManifestPanel must be enabled (BTW, I still can't figure out how to enable this in the configuration.), and once enabled the collection will show up in the collection explorer.

Would it be possible to load collections as they have been loaded in the past when the collectionManifestPanel is not enabled? And load them into the collection explorer only when the collectionManifestPanel is enabled?

Is this conditional behavior desirable? I'm not sure.

My only point right now is that if someone has relied on the previous behavior but wants to pull in the newest release of Mirador, they will experience this update as a breaking change.

camillevilla:
This ticket is being closed as part of the Mirador 3 issues review. We encourage all Mirador 2 implementers to try out Mirador 3 and leave feedback in new tickets.

PR #1507, opened 2017-11-20
Add back manifest support for collectionUri on the classic manifests panel

dicksonlaw583:
Allow the classic manifests panel to accept first-depth manifests from preloaded `collectionUri` entries by hooking into the `onCollectionReceived` event.

Addresses: #1506

### DIFF
--- a/js/src/viewer/manifestsPanel.js
+++ b/js/src/viewer/manifestsPanel.js
@@ -55,6 +55,10 @@
           _this.eventEmitter.subscribe('manifestReceived', function(event, newManifest) {
             _this.onManifestReceived(event, newManifest);
           });
+          
+          _this.eventEmitter.subscribe('collectionReceived', function(event, newCollection) {
+            _this.onCollectionReceived(event, newCollection);
+          });
         },
 
         bindEvents: function() {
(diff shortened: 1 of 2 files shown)


jeffreycwitt:
Thanks @dicksonlaw583 

One question for you and the mirador community: What happens in cases here, where the CollectionPanel is turned on and a 1-level deep collection is included in the data.collectionUri? 

Will the manifests for that collection load and appear in the collection explorer? 

I could see different implementors desiring differing behaviors. 

One implement might want the manifest in the collection to load and the collection to appear in the collection explorer. 

Another person might have a single collection of 1,000 manifests and not want the manifest to load, but only to appear in the collectionExplorer. 

Would it be possible to add a property to the collection called "loadManifests": true so that the data property might look like: 

`{collectionUri: myCollectionUrl, loadManifests: true}`

In the case of true, the manifests of the collection will be loaded on Mirador initiation. If the collectionExplorer config is set to true, the collection will also appear in the collectionExplorer.

In the case of false, the manifests  will not be loaded, but the collection will appear in the collectionExplorer if the collectionExplorer config is set to true. If the collectionExplorer config is set to false neither the collection nor its manifests will appear.

What do you think?

dicksonlaw583:
The PR's behaviour is exactly like the way it used to be, it loads in the manifests in sequence without bringing up the collection tree.

I'm not sure about how widespread the `loadManifests` use case is, but if there is demand for it I can look into it in my own time.

aeschylus:
Thanks for this @dicksonlaw583, and for a thorough test.

PR #1510, opened 2017-12-13
Make i18nPath independent from buildPath

morpheus-87:
This MR makes the i18nPath independent from the buildPath, so that one can host the locales anywhere.


aeschylus:

Issue #1515, opened 2018-01-31
Uncaught Unsupported image format for LegacyTileSource error in http://projectmirador.org/demo/


stephanpauls:
Hi,

When we add the manifest file : http://services.libis.be/m2/manifest/IE9002686manifest.json
which refers to jp2 images, the manifest is visible in the manifest list, but when we try to open it, it does not appear in the viewer and the error 'Uncaught Unsupported image format for LegacyTileSource.' is logged in the console. 

It does work in another demo viewer like : http://demos.biblissima-condorcet.fr/mirador

regisrob:
Hi,
This manifest does not work for me on the latest develop branch... but not for the same reason! The bug I am facing is this one #1495. 
It seems to me that the "LegacyTileSource" bug has already been fixed by @aeschylus in the "iiif-evented-canvas" library dependency (see this release in November: https://github.com/sul-dlss/iiif-evented-canvas/releases/tag/0.0.3-beta). So it may just be a matter of updating your dependencies in Mirador (by setting "iiif-evented-canvas" to "0.0.3-beta" in package.json and then running "npm install")
Hope that helps

stephanpauls:
Hi,

Thanks for your help, but http://projectmirador.org/demo/ is not or own viewer. It is the demo viewer of the mirador developers group. So it may be a task for  @aeschylus  or another team member to update this site.

stephanpauls:
Hi,

Sorry , the issue is still open.

Thanks for your help, but http://projectmirador.org/demo/ is not or own viewer. It is the demo viewer of the mirador developers group. So it may be a task for @aeschylus or another team member to update this site.

sdm7g:
I'm also getting this error with current develop branch on manifests that work on v2.4.0. 
Updating package.json to `"iiif-evented-canvas": "0.0.1-alpha-5",` as suggested above @regisrob seems to fix it.

aeschylus:
I have this [open PR](https://github.com/ProjectMirador/mirador/pull/1522) that will fix this issue for some resources, however, I think more fixtures are needed. The link to a fixture manifest above no longer resolves. Are there any remaining, failing manifests that can be offered up for testing purposes?

Lucidiot:
Here is a failing manifest we have : https://data.teklia.com/manifests/IIIF-2018/HORAE/BNF/text-4.json

Lucidiot:
[...]

PR #1522, opened 2018-03-02
update evented-canvas

aeschylus:
This is intended to fix this issue: https://github.com/ProjectMirador/mirador/issues/1515

However, it doesn't seem to do so. Updating is still necessary, however.


mejackreed:
If the update doesn't address #1515, what does the update address?

mejackreed:
:shipit: thanks for the nudge @regisrob

regisrob:
😉 thanks!

PR #1523, opened 2018-03-06
Added facing-pages viewingHint and support for per canvas viewingHints

atiro:
Added support for per canvas viewingHints and handle the facing-pages hint. I'm not sure if there is any other logic that needs to be handled with page indexing if a volume alternates between paged/facing-pages. It seems to work so far...

### DIFF
--- a/js/src/widgets/bookView.js
+++ b/js/src/widgets/bookView.js
@@ -414,10 +414,17 @@
 
       this.focusImages = [];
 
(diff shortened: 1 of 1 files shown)


atiro:
AH, I can see a problem already when you click previous on the following canvas, it will skip over as it is back to 2 pages. Will modify.

aeschylus:

PR #1532, opened 2018-06-01
Regex assumes encoded URI from manifest

superbland:
I had some links which contained unescaped Arabic characters in part of the URL. As this didn't match the regular expression, they were not recognised as links by the `addLinksToUris` method.

I updated the code not to assume the URI would already be encoded, by first always decoding (to avoid double-encoding), then encoding, before using the regex. In this way manifests with unencoded URI's will be processed, the link can have an encoded href attribute, and a more meaningful, decoded display.


aeschylus:

PR #1540, opened 2018-08-08
CSS directory cleanup

markmatney:
1. The standalone jQuery UI stylesheet has been removed in favor of the one from npm. Now the version numbers of the JS and CSS match.
2. The "materialize-icons" styles have been moved to the `less` directory, because (a) it seems that that is the appropriate place for styles that should be put into `mirador.css`, and (b) the `css` directory is being used both as a temp directory for CSS-related Grunt tasks and as place for third-party stylesheets. The styles have also been scoped to `.mirador-container`.
3. The standalone Normalize.css has been removed in favor of the one from npm. Unfortunately, the current version (2.1.2) is not available on npm, so I used the closest later version that was available (3.0.1).

    In doing this, I first reverted the changes that were made to the current Normalize.css. One of those changes (on `a:focus`) ended up going [into the upstream repo](https://github.com/necolas/normalize.css/commit/4b94eb7d4d290251d43684a087d0564ef4eb3685) for 3.0.0, and the other one (on `hr`) I moved to the "globals" less file (which in turn scopes it to `.mirador-container`).

    I am fairly confident that the rest of the [changes](https://github.com/necolas/normalize.css/blob/master/CHANGELOG.md) between the two versions don't break anything, but it would be nice if someone else could review.


aeschylus:

Issue #1561, opened 2018-12-05
Multiple images showing when navigating quickly


knabar:
This is an issue I ran into in my own project, but was able to reproduce as of today on http://projectmirador.org/demo/ on Chrome 70.0.3538.102

- Load http://projectmirador.org/demo/ with browser cache emptied, so no images are cached
- In the right-hand viewer, quickly click the right arrow to advance to the next image about 5-10 times to advanced faster than the images can load
- Then navigate backwards with the left arrow one image at a time - multiple images will show up when the current image is smaller than the one preceding it.

![screen shot 2018-12-05 at 21 35 14](https://user-images.githubusercontent.com/197099/49542377-bc8ed580-f8d5-11e8-95f5-cc49734b1746.png)

I'm assuming there is an issue with images not yet having loaded not being removed from display and then showing up once they are loaded, even when they are no longer the active image.

This is a major issue in my project, since I have images that load rather slowly, so it's easy to trigger.   Does anybody have an idea on where in the code base the cause for this could be?  I've tried a number of simple hacks, but have not been able to improve the situation.

camillevilla:
This ticket is being closed as part of the Mirador 3 issues review. We encourage all Mirador 2 implementers to try out Mirador 3 and leave feedback in new tickets.

PR #1564, opened 2018-12-23
Fix opacity of images shown or hidden before they finish drawing

knabar:
Attempt to fix issue https://github.com/ProjectMirador/mirador/issues/1561

When showing or hiding an image that has not yet been drawn, instead of doing nothing remember the desired opacity setting and then apply it when drawing image the first time.


knabar:
It appears that these changes are not yet sufficient to resolve the issue in all cases. Still investigating.

aeschylus:

PR #1565, opened 2019-01-06
Clear out window events on re-initialization

knabar:
On window re-initialization, e.g. on layout change, unsubscribe any existing events to avoid duplicate event handling.  Not unsubscribing the `TOGGLE_BOTTOM_PANEL_VISIBILITY` event in particular causes the bottom panel toggling to stop working, since every click will toggle it twice, setting it back to its original state.

### DIFF
--- a/js/src/workspaces/window.js
+++ b/js/src/workspaces/window.js
@@ -69,6 +69,11 @@
           focusState = _this.viewType,
           templateData = {};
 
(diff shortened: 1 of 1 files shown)


aeschylus:

This seems useful from a memory leak perspective. If we can get some tests going on this, I'd love to see it reappear on the release-2.7 branch.

PR #1573, opened 2019-01-11
open reducers and actions to modification by plugins

aeschylus:


src/reducers/index.js
```diff
@@ -3,19 +3,18 @@ import workspaceReducer from './workspace';
 import windowsReducer from './windows';
 import manifestsReducer from './manifests';
 import configReducer from './config';
-import infoResponsesReducer from './infoResponses';
```
mejackreed: I think this maybe snuck in with the repo transfer. This removes the infoResponses stuff I added.

__tests__/integration/mirador/plugins.html
```diff
@@ -12,89 +12,103 @@
     <script src="../../../node_modules/react-dom/umd/react-dom.development.js"></script>
     <script>document.write("<script type='text/javascript' src='../../../dist/mirador.min.js?v=" + Date.now() + "'><\/script>");</script>
     <script type="text/javascript">
-      class MiradorShareButton extends React.Component {
```
mejackreed: This is probably just an artifact of the repo transfer, but do you mind updating the whitespace changes here so that this isn't entirely rewritten?

src/reducers/index.js
```diff
@@ -3,19 +3,18 @@ import workspaceReducer from './workspace';
...
-  infoResponses: infoResponsesReducer,
-});
-
-export default rootReducer;
+export default function createRootReducer(pluginReducers) {
+  return combineReducers({
+    workspace: workspaceReducer,
+    windows: windowsReducer,
+    manifests: manifestsReducer,
+    config: configReducer,
```
mejackreed: I think the infoReducer needs to be in here also

PR #1577, opened 2019-01-11
Gets the script from CDN if not available locally

mejackreed:
Enables netlify / deploy to be able to easily display this page


mejackreed:
Working now here: https://deploy-preview-1577--mirador-dev.netlify.com/__tests__/integration/mirador/plugins.html

🎊

PR #1603, opened 2019-01-15
Add prettier to project, closes #1600

### LINKED ISSUE #1600: Add .prettierrc

bodo22:
closes #1600


mejackreed:
Could we review this approach together? I'd like to be able to implement something that didn't have such wide ranging changes if possible.

mejackreed:
It seems to me that we are introducing additional style guides where eslint now takes on prettier/recommended. Is there a way to send our current eslint config to prettier instead?

I'm a little concerned about the cognitive overhead of figuring out which config should take precedence for which setting. Especially since this is quite a large departure from AirBnb starting point.

bodo22:
Agreed! Maybe it makes more sense to open a PR managing some `// eslint-disable-line`, closing this for now.

Issue #1611, opened 2019-01-16
Export unconnected components...


jkeck:
... in addition to existing default connected component.  This could help us w/ writing tests for these components as well as allowing other react apps to use parts of Mirador.

See https://github.com/reduxjs/redux/blob/master/docs/recipes/WritingTests.md#connected-components for more info

### Components:
 - [X] App.js #1634
 - [X] ~~Display.js~~ (not connected)
 - [X] ManifestForm.js #1634
 - [X] ManifestListItem.js #1634
 - [X] ~~ManifestMetadata.js~~ (not connected)
 - [X] OpenSeadragonViewer.js #1634
 - [x] ViewerNavigation.js #1614 
 - [ ] Window.js
 - [X] WindowTopBarButtons.js #1634 
 - [ ] WindowViewer.js
 - [ ] Workspace.js
 - [X] ~~WindowBackground.js~~ (not connected)
 - [X] WindowTopBar.js #1634 
### Component Tests:
 - [X] ~~Display.test.js~~ (not connected)
 - [X] ManifestListItem.test.js #1634
 - [x] ~~ManifestMetadata.test.js~~ (not connected)
 - [x] ViewerNavigation.test.js #1614
 - [ ] Window.test.js
 - [X] ~~WindowBackground.test.js~~ (not connected)
 - [X] WindowTopBar.test.js #1634 
 - [ ] WindowViewer.test.js
 - [ ] Workspace.test.js

PR #1614, opened 2019-01-16
Export unconnected component to better test Mirador components

mejackreed:
Based off of group discussion in #1611 

Seems like a really useful approach for us to test components easier and also provide unconnected for reuse in plugins/ custom apps.


mejackreed:
Yes ideally we should do this for everything. Maybe we can tick them off one at a time to start getting them in sooner? I've updated the comment here to track this: https://github.com/ProjectMirador/mirador/issues/1611#issue-399602015

Issue #1632, opened 2019-01-16
Look at existing windowing system libraries to more fully-define windowing behaviours


aeschylus:
https://deploy-preview-1594--mirador-dev.netlify.com/__tests__/integration/mirador/
https://strml.github.io/react-grid-layout/examples/0-showcase.html
https://github.com/zzarcon/react-cristal
https://palantir.github.io/react-mosaic/
https://golden-layout.com/

Have a look through these and record here some of the drawbacks and advantages. Possibly prioritise windowing behaviours based on use-case need and difficulty of implementation.

sdellis:
Look into the dependencies, styles, extensibility, etc. of the grid systems to help inform a decision.

mejackreed:
I've started a sheet here to add our data: https://docs.google.com/spreadsheets/d/16loH81rUWaff2oD8ppsRviar3MajJUjnxUJl6QKx4U4/edit#gid=0

PR #1635, opened 2019-01-16
Spike to convert the window top bar to the material framework; part of #1632

cbeer:
🤷‍♂️


sdellis:
This bundle is pretty big. The [Minimizing Bundle Size](https://material-ui.com/guides/minimizing-bundle-size/) doc recommends two options for reducing bundle size as a temporary fix to tree shaking. Unfortunately, [the ability to tree shake material-ui is still being worked out](https://github.com/mui-org/material-ui/pull/13391). 

However, it looks like this branch follows the Option 1 recommendation, so I'm wondering why material-ui is making the bundle so heavy?

What does the `withStyles` module do? I wonder if this destructured import requires [babel-plugin-import](https://github.com/ant-design/babel-plugin-import) to help reduce the size?

src/components/WorkspaceControlPanel.js
```diff
@@ -34,34 +44,73 @@ class WorkspaceControlPanel extends Component {
...
-      <div className={ns('workspace-control-panel')}>
-        <ConnectedManifestForm setLastRequested={this.setLastRequested} />
-        <ul>{manifestList}</ul>
+      <Drawer
+        className={classNames(classes.drawer, ns('workspace-control-panel'))}
+        variant="permanent"
+        classes={{ paper: classNames(classes.drawer) }}
+        open
+      >
+        <Menu id="add-form" anchorEl={anchorEl} open={Boolean(anchorEl)} onClose={this.handleClose}>
```
sdellis: We should use semantic elements and wrap this menu in a <nav> tag, e.g. `<nav id="mirador-workspace-control-panel-menu">`.

camillevilla:
@sdellis thanks for looking into the bundling issues. It would be interesting to put heads together to talk about Material UI’s viability. [withStyles](https://material-ui.com/customization/css-in-js/#withstyles-styles-options-higher-order-component) returns an HOC with a `classes` property. 

`babel-plugin-import` sounds useful. Should we add a ticket to implement it outside of this branch?

sdellis:
@camillevilla 	yes, let's tackle the bundle size in a separate ticket/branch. This PR looks good to me and none of my comments should necessarily stop it from merge I don't think.

mejackreed:
Do we know if `babel-plugin-import` will reduce the bundle size yet? I'm just asking the question here as a I'd like to move this forward today, but it might be useful to know if there is a way forward/backward on taking on a large dependency required in our bundle?

src/components/WorkspaceControlPanel.js
cbeer: I'm not sure it's so much of a nav menu right now; it's just wrapping a single form that behaves modal-y; the good news is, the whole thing is temporary until we build out the actual page.

sdellis:
@mejackreed we are somehow including the entirety of Material UI in this bundle, and it seems like there are ways to prevent this. We should be able to figure out the treeshaking or destructuring piece in another ticket/branch, so I see that as the way forward. I think it will be more difficult to back out of any such framework, but Material UI in documentation and theory seems to do the things we want. I'm approving this PR for merge and creating a separate ticket for treeshaking now.

cbeer:
I'll rebase it.

jkeck:
:shipit:

eps1lon:
Hey, material-ui contributor here. Just stumbled over this since this was referenced from one of my PRs and I wanted to investigate what issues you encountered with tree shaking.

>This bundle is pretty big

@sdellis
Do you have some diagnostics about the bundle? Something like a tree map for the bundled files before and after this PR?

You don't actually have to use `withStyles` if you use material-ui components. It's just the internal styling solution. Since you're already using css you could simply use classnames from your stylesheets and use those in the material-ui components.

sdellis:
@eps1lon Our treemap only shows the mirador library because we are specifying `umd` for `LibraryTarget` in our webpack. My understanding is that we need an `es6` build to be able to treeshake. I noticed that `umd-module` is also an option for `LibraryTarget`. It wraps es6 in umd. Not sure if that will work, but can test it out.

mejackreed:
I'm seeing this in our current build:

screen shot 2019-01-31 at 10 54 17 am

https://gist.github.com/mejackreed/ae13d154ded8a836891acff3f776cc24#file-stats-json

PR #1640, opened 2019-01-17
adds json static server with minimal configuration

### LINKED ISSUE #1636: Add fake (local) json api endpoint to serve test fixtures
integration / e2e tests should depend on fixtures served within the context of the test framework.  This can easily be done with a node static server (e.g. https://github.com/typicode/json-server).  The server should be able to provide appropriate http response codes for valid and invalid requests.

christopher-johnson:
closes #1636 
note that custom response statuses are configurable, but not included in this PR.


__tests__/logs/json-server.log
```diff
@@ -0,0 +1 @@
+
```
mejackreed: When running this locally I'm seeing this file change (as expected). What is unexpected is that this is checked in. Should we figure out a way to ignore this log?

__tests__/integration/mirador/basic.test.js
```diff
@@ -7,15 +7,15 @@ describe('Basic end to end Mirador', () => {
...
     await page.waitFor(1000);
-    await expect(page).toMatchElement('li', { text: 'https://purl.stanford.edu/sn904cj3429/iiif/manifest' });
+    await expect(page).toMatchElement('li', { text: 'http://localhost:5000/sn904cj3439.json' });
     await expect(page).toMatchElement(
       'h3',
       "Peter's San Francisco Locator. The Birds-Eye-View Map of the Exposition City. Published by Locator Publishing Co",
     );
     await expect(page).toMatchElement('div', /Color/);
-  });
+  }, 30000);
```
mejackreed: One quick question about this, I see this added in this test, but not in the others. Is this needed here?

package.json
```diff
@@ -6,12 +6,15 @@
   "scripts": {
     "lint": "node_modules/.bin/eslint ./",
     "server": "node_modules/.bin/http-server",
+    "pretest": "npm run server:json",
+    "posttest": "kill $(lsof -t -i:5000)",
```
bitzl: Regarding a great developer experience - should we support development on windows? Many libraries run Windows desktops for their staff.

__tests__/logs/json-server.log
christopher-johnson: this file should not be committed.  added to .gitignore and removed.

christopher-johnson:
> Is there a way that you could add a small, single test proof of concept test here that would utilize the custom response status aspect of this added dependency?  

I hesitate to do this with this PR.  I think that the integration tests would need to be enhanced to allow per test (or per category) API server configuration to do this.  This is entirely possible, but not with the integration test framework as it currently exists.  I have added a basic [server.js](https://github.com/ProjectMirador/mirador/blob/259352d2a106d22ea6160236a730c06844aa1db4/scripts/json-server/server.js) script that shows the pattern for a server config that can be used when developing a real integration test framework discussion takes place.

christopher-johnson:
As requested, added an integration test that breaks Mirador by serving a non-conforming json from the fixture server. 
 
Note that an invalid response is not yet handled by ManifestMetadata (that assumes all API responses are valid manifests).

Also, accessing React errors from the jest-puppeteer page object is problematic.  Not clear to me how or if this can be done.  I suggest reconsidering jest-puppeteer for e2e tests.

aeschylus:
Having a look at this now locally in its most recent state. Everything reads well, but let me finish testing this before merge.

Issue #1587, opened 2019-01-13
"Add Manifest/Album" button


mejackreed:
When clicked, this button should open a basic "album selection interface" (which could just be a list of text for right now). When clicking on an album/manifest it should open a new `Window` with the selected manifest.

<img width="780" alt="screen shot 2019-01-13 at 9 16 23 am" src="https://user-images.githubusercontent.com/1656824/51087713-f42e4000-1713-11e9-8168-251124e7bc9c.png">

List of manifests and existing manifest form field should be contained here

@jvine did we come to a consensus on the naming here for the "Album?"

camillevilla:
We now have an add manifest button, but it's a simple text form. 
Moving this to the backlog, as it seems like there are more pre-requisites to this view.

camillevilla:
Adding in a backend detail that this work should address:
The manifest list should also be added to the config. This represents a list of manifests that CAN be added to the windows. Note that this list should be separate from the list of manifests currently displayed by windows. 

I'm including the GIFs below to clarify the concept:

current M3 implementation
![manifest-list](https://user-images.githubusercontent.com/5402927/52296438-3b82a780-2933-11e9-9794-edc96af79f59.gif)

example from M2
![m2-manifest-list](https://user-images.githubusercontent.com/5402927/52296679-d8dddb80-2933-11e9-8c24-b009bb9d1436.gif)

ggeisler:
Below are several tickets that collectively represent a load window plugin that I think could meet the goal of this ticket:

- #1833 - Design for load window plugin (story ticket)
- #1834 - Implement 'List view' of load window plugin
- #1835 - Add 'Layout' option to load window tools menu
- #1836 - Implement 'Card view' of load window plugin

camillevilla:
I've added these to the project board and am going to close this in favor of #1833.

Issue #1586, opened 2019-01-13
Update Workspace interface to reflect sidebar design


mejackreed:
## Current "Workspace"
<img width="1156" alt="screen shot 2019-01-13 at 9 13 01 am" src="https://user-images.githubusercontent.com/1656824/51087685-ab768700-1713-11e9-90b8-b12e4afca1c7.png">
## Target "Workspace"
<img width="808" alt="screen shot 2019-01-13 at 9 13 44 am" src="https://user-images.githubusercontent.com/1656824/51087688-b7fadf80-1713-11e9-8abe-ea296c075c77.png">

camillevilla:
Seems like this first pass has mostly been addressed in #1682, with some refactoring in #1725.

Full-screen button functionality addressed by #1702 / #1713

I'm opening another ticket for the menu bar functionality:
![workspace toolbar menu](https://user-images.githubusercontent.com/5402927/51930206-e08b0680-23ae-11e9-9c06-5ebccba6abca.png)

PR #1643, opened 2019-01-17
Refactor control-panel div from App into a WorkspaceControlPanel component.

jkeck:
Rationale being that issues like #1586 and #1587 will be able to build on top of this and not have to all work in the top-level App component.

🤷‍♂️

### DIFF
--- a/src/components/WorkspaceControlPanel.js
+++ b/src/components/WorkspaceControlPanel.js
@@ -1,0 +1,77 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import Display from './Display';
+import ConnectedManifestForm from './ManifestForm';
+import ConnectedManifestListItem from './ManifestListItem';
+import ns from '../config/css-ns';
+
+/**
+ * Provides the panel responsible for controlling the entire workspace
+ */
+class WorkspaceControlPanel extends Component {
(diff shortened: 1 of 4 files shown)


jkeck:
Pushed up fix to the conflict.  Also, I had added a test last night which allowed the build to pass using the existing CC config (w/ 84% diff coverage), just as a heads up.

src/components/WorkspaceControlPanel.js
```diff
@@ -0,0 +1,77 @@
...
+    super(props);
+    this.state = {
+      lastRequested: '',
+    };
+
+    this.setLastRequested = this.setLastRequested.bind(this);
+  }
+
+  /**
+   * setLastRequested - Sets the state lastRequested
```
jbaiter: I think that docstring could be a bit more descriptive, what exactly is `lastRequested`?

aeschylus: Originally this was a flag to allow us to display only the information for the most recently requested manifest. We can probably just delete this at this point, as the side panel designs don't really require this behaviour.

jkeck: I agree @jbaiter.  I was simply moving this comment w/ its accompanying code from another place to here in my component extraction, so not really sure I'm the best person to describe what the intent of the original author was.

PR #1646, opened 2019-01-17
Add css linting w/ sass-lint

### LINKED ISSUE #1639: Add scss linting
Linting our scss could help bring some consistency to the styles around the app.

jkeck:
Closes #1639

### DIFF
--- a/src/styles/index.scss
+++ b/src/styles/index.scss
@@ -1,42 +1,44 @@
+@import 'variables';
+
 body {
-  background: white;
+  background: $white;
   height: 100%;
 }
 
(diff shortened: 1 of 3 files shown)


src/styles/index.scss
```diff
@@ -1,42 +1,44 @@
+@import 'variables';
```
jkeck: Not sure if this is really needed w/ the single css file (or even the right pattern), but thought I would pull it out into a separate file as opposed to throwing a bunch of variables at the top of the file. 🤷‍♂️

PR #1654, opened 2019-01-17
Fixes app breaking when trying to render WindowViewer with failed fetch

bodo22:


christopher-johnson:
this should be testable and include a test

PR #1656, opened 2019-01-18
refactor MiradorViewer logic out of init function

mejackreed:
The intent here is to create a more testable, extensible interface for default initialization

Increases coverage stats 10.5% 🕺

### DIFF
--- a/src/init.js
+++ b/src/init.js
@@ -1,66 +1,9 @@
-import React from 'react';
-import ReactDOM from 'react-dom';
-import { Provider } from 'react-redux';
-import deepmerge from 'deepmerge';
-import ConnectedApp from './components/App';
-import createRootReducer from './reducers/index';
-import { store, actions } from './store';
-import settings from './config/settings';
+import MiradorViewer from './lib/MiradorViewer';
 import './styles/index.scss';
-
-/**
- * Process Plugins
- */
-const processPlugins = (plugins = []) => {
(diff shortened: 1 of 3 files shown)


src/init.js
```diff
@@ -1,66 +1,9 @@
...
-
-  ReactDOM.render(
-    <Provider store={store}>
-      <ConnectedApp config={config} />
-    </Provider>,
-    document.getElementById(config.id),
-  );
-
-  return viewer;
+  return new MiradorViewer(config);
```
aeschylus: Would like to avoid using a class here, but can be factored out later. https://medium.com/javascript-scene/javascript-factory-functions-vs-constructor-functions-vs-classes-2f22ceddf33e

src/lib/MiradorViewer.js
```diff
@@ -0,0 +1,73 @@
...
+    store.dispatch(action);
+
+    ReactDOM.render(
+      <Provider store={store}>
+        <ConnectedApp config={config} />
+      </Provider>,
+      document.getElementById(config.id),
+    );
+
+    return viewer;
```
aeschylus: Classes inherently return themselves. This class will return something different depending on whether it is called with new or not.

mejackreed: Gotcha. Right now we aren't exposing this class publicly so we can control who/where it is instantiated, right?

Issue #1668, opened 2019-01-21
Normalize state


charbugs:
Redux docs suggest to keep state in a normalized fashion: https://redux.js.org/faq/organizing-state#how-do-i-organize-nested-or-duplicate-data-in-my-state

Therefor each state item should be treated like a database table with rows and ID's (see `window` and `manifests` in the example). A row should only contain primitives and arrays. Related objects should be stored as their own items and referenced by their ID's. Having ID's twice, as object keys and as property values of the respective object, can be convenient in iterations.

This may be an unnecessary overhead if a certain state item is meant to only store a single object/row. So I suggest to treat it in a normal way (see `config` in the example)

```javascript
{
  windows: {
    'window-1': {
      id: 'window-1',
      manifestId: 'http://foo',
      canvasIndex: 0
    },
    'window-2': {
      id: 'window-2',
      manifestId: 'http://bar',
      canvasIndex: 0
    }
  },
  manifests: {
    'http://foo': {
      id: 'http://foo',
      manifestation: '...'
    },
    'http://bar': {
      id: 'http://bar',
      manifestation: '...'
    }
  },
  config: {
    'foo': 'bar'
  }
}
```

charbugs:
This might be also a design decision that should be documented in the wiki.

christopher-johnson:
:+1: 
this is a good idea.  in the absence of GraphQL, it may also be useful to normalize the manifestation as well.

jbaiter:
I think this is the way to go for the store state.

>  this is a good idea. in the absence of GraphQL, it may also be useful to normalize the manifestation as well.

This might also be worth a shot, though the level of normalization should probably be driven by actual use cases in the code. I can image that having top-level `canvas`, `annotations` and `range` keys could be useful.

aeschylus:
I agree with this approach. We've seen that listening to arrays with connect causes frequent re-renders. @jbaiter suggests to incrementally convert the state, and I agree with that approach. I am experimenting with removing the manifestation from the store, as it complicates state serialisation.

mejackreed:
👍 to the incremental approach also.

bitzl:
:+1: to normalize the state and the incremental approach

PR #1669, opened 2019-01-22
1668 normalize state

charbugs:
Main reason for this PR is to normalize the window reducer as described in #1668. But it takes changes in 13 files to get all tests pass. For me, this shows the strong coupling of the existing code base and stresses the importance of a clear concept for the state layout.

### DIFF
--- a/__tests__/src/actions/window.test.js
+++ b/__tests__/src/actions/window.test.js
@@ -11,7 +11,7 @@ describe('window actions', () => {
 
       const expectedAction = {
         type: ActionTypes.ADD_WINDOW,
-        payload: {
+        window: {
           id: 'helloworld',
           canvasIndex: 1,
           collectionIndex: 0,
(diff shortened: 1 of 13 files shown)


charbugs:
The PR doesn't pass the coverage test. But it's not clear to me why. If i check the details it complains about a file that I didn't touch. Could anyone ease the restrictions? @mejackreed perhaps?

mejackreed:
I'm 👍 to easing the restrictions here and can merge.

@charbugs can you rebase on master to resolve the conflicts?

jkeck:
I've noticed that code coverage can be a bit tricky in cases like this.  It might even be that you've reduced code (:tada:), some of which was tested, and that caused a total test coverage reduction.

In these cases I think we can be pragmatic about the coverage failures when the intent/testing of the code meets the community expectations/guidelines.

mejackreed:
For clarity, we are going to force push a version of this branch up including only the needed commits by @charbugs.

PR #1675, opened 2019-01-22
Update json-server startup/shutdown to prevent rogue processes

mejackreed:
When running the tests after merging #1640 I started to get some rogue processes that I was having to kill manually.

The intent of this PR is to use the built in functionality provided by our test harness to manage this process, rather than having to do this process management on our own.

I _think_ this should be supported cross platform.

@bitzl , @christopher-johnson what do you think?


christopher-johnson:
there are several problems with using `posttest `to kill the e2e fixture server.  If _any_ test fails, then `posttest `will not execute and the server process will not be killed.  This is by design in npm.  

To make this happen [npm-run-all](https://github.com/mysticatea/npm-run-all/blob/master/docs/npm-run-all.md) can be added to the test script.  

I think that the e2e tests should execute independently from the unit tests (and start from a different script) and probably should be in their own package.  I am not happy with using the` jest-puppeteer` config globally, as this affects all tests.  If you remove the preset from the` jest config`, you will see that the unit tests are also affected.  This should not happen.

mejackreed:
Thanks for the comment. My understanding is that you agree using `posttest` is probably undesirable. This PR solves that problem, but you also would prefer a different approach to the entire test setup (better separation of unit tests and integration tests).

Is that worth a discussion elsewhere? This PR doesn't aim to resolve that problem at the moment. However it does solve the issue that macOS and Windows users will have a rogue running process https://github.com/ProjectMirador/mirador/blob/master/scripts/json-server/kill-server.js#L4-L8 . This is something I ran into a lot over the past day and was having to manually kill of processes.

christopher-johnson:
this does solve the multiplatform rogue process issue. 

I did a quick benchmark to see if there is any performance benefit in not using test hooks.
init and kill fixture server from jest-puppeteer.config
```
test: 5.674ms
``` 
init and kill fixture server with npm test hooks
```
test: 6.419ms 
```
so I guess I am :+1:  on this as a `quick fix`.

PR #1676, opened 2019-01-22
Zoom OSD canvas based off of the first tilesource passed in

### LINKED ISSUE #1674: When navigating canvases, the next/previous canvas is not always the same size, OSDViewer should handle that
Viewer should be centered on canvas

canvassize

mejackreed:
Fixes #1674 

Useful fixture manifest with canvases that vary in size: https://media.nga.gov/public/manifests/nga_highlights.json

## Before
canvassize
## After
canvassizefix

### DIFF
--- a/src/components/OpenSeadragonViewer.js
+++ b/src/components/OpenSeadragonViewer.js
@@ -44,7 +44,15 @@ export class OpenSeadragonViewer extends Component {
     const { tileSources } = this.props;
     if (!this.tileSourcesMatch(prevProps.tileSources)) {
       this.viewer.close();
-      tileSources.forEach(tileSource => this.addTileSource(tileSource));
+      Promise.all(
+        tileSources.map(tileSource => this.addTileSource(tileSource)),
+      ).then(() => {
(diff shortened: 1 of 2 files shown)


src/components/OpenSeadragonViewer.js
```diff
@@ -44,7 +44,14 @@ export class OpenSeadragonViewer extends Component {
     const { tileSources } = this.props;
     if (!this.tileSourcesMatch(prevProps.tileSources)) {
       this.viewer.close();
-      tileSources.forEach(tileSource => this.addTileSource(tileSource));
+      Promise.all(
+        tileSources.map(tileSource => this.addTileSource(tileSource)),
+      ).then(() => {
+        // An incredibly naive way to evaluate this for now.
```
jbaiter: Does this warrant a `FIXME` or `TODO` prefix?

mejackreed: Sure! Thanks for the suggestion. I've pushed up something adding the TODO and more context.

PR #1679, opened 2019-01-23
Refactor workspace component. closes #1678

### LINKED ISSUE #1678: Refactor <Workspace /> and related tests

charbugs:
- component function -> class
- shallow rendering in component test
- use matchesElement() in component test

### DIFF
--- a/src/components/Workspace.js
+++ b/src/components/Workspace.js
@@ -10,19 +10,26 @@ import ns from '../config/css-ns';
  * @memberof Workspace
  * @private
  */
-const Workspace = ({ windows }) => (
-  <div className={ns('workspace')}>
-    {
-      Object.values(windows).map(window => (
-        <ConnectedWindow
-          id={window.id}
-          key={window.id}
-          window={window}
-        />
-      ))
-    }
-  </div>
-);
+export class Workspace extends React.Component {
+  /**
+   * guess what
+   */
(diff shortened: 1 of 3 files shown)


src/components/Workspace.js
```diff
@@ -10,19 +10,26 @@ import ns from '../config/css-ns';
...
-          key={window.id}
-          window={window}
-        />
-      ))
-    }
-  </div>
-);
+export class Workspace extends React.Component {
+  /**
+   * guess what
```
jbaiter: Can we leave out that docstring? I think we should only add docstrings that add real value, I don't think this would qualify :-)
Probably required some quorom on the eslint settings, maybe an exception for commong stuff like `render` and the React lifecycle hooks would be good?

mejackreed: 👍

charbugs: Sorry, just wanted to tease the useless docstrings like /** Renders things */

mejackreed: I think a PR updating the config here: https://github.com/ProjectMirador/mirador/blob/master/.eslintrc#L15-L23 would be most welcome, especially based on our now shared experience.

charbugs:
Sry, I screwed up. Will create a new PR tonight/today.

Issue #1670, opened 2019-01-22
Understand/Fix codeclimate coverage preferences


aeschylus:

christopher-johnson:
It might be worthwhile to look at other code coverage tools (e.g. codecov.io).  [example here](https://codecov.io/gh/christopher-johnson/mirador/branch/codeclimate)

I find the codeclimate test coverage PR report to be lacking.

mejackreed:
Great idea @christopher-johnson . I've added CodeCov integration and updated Travis with a `CODECOV_TOKEN` env var. We should be all setup for a PR to switch out codeclimate for CodeCov if someone wants to take that on.

PR #1683, opened 2019-01-23
adds codecov.io / replaces code climate

christopher-johnson:
Note that codecov.io is configurable using a [.yml file](https://docs.codecov.io/docs/codecov-yaml).  

It is also possible to push notifications to Slack.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,12 @@
 language: node_js
 node_js:
   - 'node'
-before_install:
-  - curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > ./cc-test-reporter
-  - chmod +x ./cc-test-reporter
+install:
+  - npm install -g codecov
 
 before_script:
   - npm install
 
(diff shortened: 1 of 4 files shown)


cbeer:
Could you say more about the reasons to replace code climate? I don't have a preference either way, but it'd be helpful to me if we also note /why/ a change is being made.

christopher-johnson:
@cbeer 
I think that codecov is a better product.  (I omit a detailed enumeration of the differences, but if you test it out, then it should clear enough) Detailed in line code coverage reporting is very useful for reviewers.  In fact, I used it today to assist with analyzing and fixing test coverage from the normalize state PR in #1681 .  

The code climate code navigation interface lacks detail and is IMO annoying for this purpose.  codecov also has more configuration options if required, including disabling the bold PR coverage report (should that be an agreed upon preference).

aeschylus:
Could not agree more with @christopher-johnson on this count. There is some more context for the discussion here: https://github.com/ProjectMirador/mirador/issues/1670#issuecomment-456850517

PR #1691, opened 2019-01-24
Refactor <Workspace /> closes #1678

### LINKED ISSUE #1678: Refactor <Workspace /> and related tests

charbugs:
See #1678
Closes #1678

### DIFF
--- a/src/components/Workspace.js
+++ b/src/components/Workspace.js
@@ -10,19 +10,26 @@ import ns from '../config/css-ns';
  * @memberof Workspace
  * @private
  */
-const Workspace = ({ windows }) => (
-  <div className={ns('workspace')}>
-    {
-      Object.values(windows).map(window => (
-        <ConnectedWindow
-          id={window.id}
-          key={window.id}
-          window={window}
-        />
-      ))
-    }
-  </div>
-);
+export class Workspace extends React.Component {
+  /**
+   * render
+   */
+  render() {
+    const { windows } = this.props;
+    return (
+      <div className={ns('workspace')}>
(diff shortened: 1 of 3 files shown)


aeschylus:
While I understand the intent of separating connected and unconnected components, I'm wary of including implementation details in the names of our components. To me it has a "code smell", and hints at an inconsistent architecture that we haven't figured out yet. Seems like the kind of thing that can be addressed later, but we should keep an eye on it.

mejackreed:
@aeschylus can you say more about this? Are you referring to `ConnectedWorkspace`?

src/components/Workspace.js
```diff
@@ -10,19 +10,26 @@ import ns from '../config/css-ns';
  * @memberof Workspace
  * @private
  */
-const Workspace = ({ windows }) => (
-  <div className={ns('workspace')}>
-    {
-      Object.values(windows).map(window => (
-        <ConnectedWindow
-          id={window.id}
```
camillevilla: The only thing that confuses me about this PR is the removal of `id` from `Window`. Can anyone provide insight into that?

jbaiter: It's because `Window` gets the `id` from the actual `window` state. I think this was a redundant attribute, even before the PR. See here: https://github.com/ProjectMirador/mirador/blob/master/src/components/Window.js#L48

camillevilla: Ah! Many thanks for the context, @jbaiter. That resolves my concern.

cbeer:
works for me 👍

PR #1708, opened 2019-01-28
add size task useful for providing size of current distribution

mejackreed:
I think this might be a useful tool, so we can be consistently evaluate our distribution size. Note, that the size shown here is gzipped. 

```sh
user$ npm run size

> mirador@3.0.0-alpha.0 size /Users/user/dev/mirador
> size-limit


  Package size: 204.85 KB
  Size limit:   300 KB
  With all dependencies, minified and gzipped
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "scripts": {
     "lint": "node_modules/.bin/eslint ./ && node_modules/.bin/sass-lint -v ./src/styles/**/*",
     "server": "node_modules/.bin/http-server",
-    "test": "npm run build && npm run lint && jest -c jest.json",
+    "test": "npm run build && npm run lint && npm run size && jest -c jest.json",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -11,6 +11,7 @@
     "build": "webpack --mode=production",
     "build:dev": "webpack --mode=development",
     "build:watch": "webpack --watch --mode=development",
+    "size": "size-limit",
```
camillevilla: If you run this task just after running the dev build (i.e. you just ran the dev server), you'll get a very different number. I'm wondering if we should:
- include an explicit call to `npm run build` to this script OR 
- add something to the README so that potential adopters can run this command upfront

mejackreed: What about making it part of the `test` ci build?

mejackreed: Great, I've pushed this change up.

PR #1713, opened 2019-01-28
Build out a WorkspaceFullScreenButton component

### LINKED ISSUE #1702: Add behavior for workspace full-screen icon
The workspace menu bar has an icon for full-screen:

window-workspace-behavior-v_1

The intention is that when the user clicks the full-screen icon, the Mirador workspace goes into full-screen mode. This is probably mostly straightforward, but here are some details about the expected behavior:

- Windows in the workspace are not maximized (there is a separate window maximize icon, and a forthcoming ticket describing the expected behavior for that) but are left at the same size and workspace position. Here's an example where the workspace contains a single window (click the workspace full-screen icon to see the behavior):

https://xd.adobe.com/view/a470e834-eaef-4168-60db-aa90f130c384-d241/screen/628c9429-b3fc-430a-a241-0b4d58316a92/Desktop-workspace-with-adjacent-browser-window

- Although windows remain the same size, because the workspace viewport is larger in full-screen mode, the user might see more window content than in non-full-screen mode, depending on how their workspace looks in non-full-screen mode. For example, see this mockup with multiple windows, where some windows are not visible in the initial workspace viewport but are visible when the workspace is maximized:

https://xd.adobe.com/view/a470e834-eaef-4168-60db-aa90f130c384-d241/screen/bdb2e5a4-479f-4bdf-a53a-6666ae348ab4/Desktop-workspace-with-adjacent-browser-window-mul

cbeer:
Fixes #1702
After:

screen shot 2019-01-28 at 16 27 01
 =>

screen shot 2019-01-28 at 16 27 32


src/actions/workspace.js
```diff
@@ -0,0 +1,17 @@
...
+ * @param  {String} windowId
+ * @memberof ActionCreators
+ */
+export function fullscreenWorkspace(fullscreen) {
+  return { type: ActionTypes.FULLSCREEN_WORKSPACE, fullscreen };
+}
+
+/**
+ */
+export function stopEslintFromComplaining() {
```
aeschylus: You can stop eslint from complaining by using the ‘default’ keyword on the function above.

cbeer: If I remember correctly, if I use default then I have to do something different in the index.js file.. I just need to figure out how to disable the eslint check here.

cbeer:
Sure, I can pull out the menu stuff into a separate commit at least.

PR #1716, opened 2019-01-29
Add a toggle button for the WindowSideBar component

### LINKED ISSUE #1582: Toggle window sidebar with button in WindowTopBar
Enable toggling of WindowSidebar with a button on the lefthand side of WindowTopBar.
screen shot 2019-01-13 at 8 59 47 am

jkeck:
Closes #1582 

window-sidebar


src/actions/window.js
```diff
@@ -40,3 +40,13 @@ export function addWindow(options) {
...
 }
+
+/**
+ * toggleWindowSideBar - action creator
+ *
+ * @param  {String} windowId
+ * @memberof ActionCreators
+ */
+export function toggleWindowSideBar(windowId) {
+  return { type: ActionTypes.TOGGLE_WINDOW_SIDE_BAR, windowId };
```
jkeck: I wrestled w/ how to handle this in the state.  This is the simplest implementation that I could think of, but wasn't sure if I should go deeper w/o knowing how we'll want to model this in the state down the road (e.g. I could set this as its own object w/i the window for future extension).

Happy to go down other routes w/ how to handle this in the state now or we can kick that down the road.

jkeck:
It appears that this is now failing coverage on https://github.com/ProjectMirador/mirador/pull/1716/files#diff-09f74b447a8923de952c2fc0057b3d8bR14 (the `mapDispatchToProps` now moved to the `WindowTopBar` container component).

What's our strategy for testing these now that they're in containers (because it appears that they've been left untested in the container refactor).

PR #1722, opened 2019-01-29
refactors state related classes

### LINKED ISSUE #1721: split actions / actionCreators from store / revise directory structure
actions do not care about store and store does not care about actions.  
they can exist separately in a state container.
Propose this directory structure:

```bash
src 
  - state
     - createStore.js
     - actions/
        - actionTypes.js
        ...
     - reducers/
     - selectors/

christopher-johnson:
closes #1721
- defers store initialization to the consumer implementation. 
- uses wildcard export for actions
- separates actions from store
- creates a container for state related classes 
- moves action-types into actions container
- adds code coverage exclusion for index files


mejackreed:
Can you please comment about the reasons for this PR. I'm not sure I grasp all of it yet from #1721. Is it mostly just a project/file reorganization approach? I would like to capture the "why" behind it for the future.

christopher-johnson:
this follows the "separation of concerns" design pattern.  store and actions are functionally different and mutually independent. 

 instantiation of store should also probably not occur before export.  What you want to do is export a store constructor and not a singleton.  This allows an implementation to instantiate store with parameters (though the createStore function does not currently do that).  From what I understand, the singleton is cached, but there is no guarantee that the instance will not be duplicated in the tree, and this is extra overhead.  

I view this PR as a prerequisite to further refinement and abstraction of actions.  This includes components only importing actions that they use.  

Also, the rationale behind separating the redux-related classes from the react classes in the directory structure seems to me very clear.   Futhermore, entry points should almost always be indices.    Using store.js as the entry point to the core distribution is a bit goofy.

mejackreed:
Thanks for that context @christopher-johnson

Issue #1662, opened 2019-01-18
Spike on rolling our own windowing system


aeschylus:
There's a demo of this in the [relevant PR](https://github.com/ProjectMirador/mirador/pull/1724), sort of. There are a few other elements of the proposed custom windowing behaviour to implement there, namely the large ("infinite"), pan-able workspace area. There are a few ways to implement this, and @sdellis and I are going to try to document these considerations here.

We need something like:
![overlapping windows - workspace shift](https://user-images.githubusercontent.com/145874/51943985-b0a02b00-23cf-11e9-8aa4-6e7a3cf6ed13.gif)
where the orange box is the browser viewport (or mirador topmost viewer container), the blue things are windows, and the outer dark red box is the workspace element, which is translated far to the left and top of the browser viewport to provide space for the windows to occupy. This margin value would be applied to all the windows' 3D transform style attribute (as children of the workspace element).

Another diagram showing this setup: (Note that the workspace "artboard" is shifted up and to the left so there is enough space for its windows, which are child elements).
<img width="970" alt="screenshot 2019-01-29 at 2 14 39 pm" src="https://user-images.githubusercontent.com/145874/51944222-59e72100-23d0-11e9-88f2-da270400ee9b.png">
Here's a gif screencapture of this being done in the browser console with the element style panel. A css scale transform is applied to the workspace element based on the bounding box of all open windows within the workspace "artboard" element, accomplishing the "exposé" feature:
![overlapping windows - expose](https://user-images.githubusercontent.com/145874/51943076-7766bb80-23cd-11e9-9c7b-4edb122232f8.gif)

aeschylus:
@jvine @ggeisler 
Does this seem right? Do you imagine the panning behaviour apply to any of the other workspace types (if we go that direction)?

jvine:
My concept of the "infinite" workspace has been that the workspace starts out being the size of your Mirador viewport, until you push something outside of it, then it expands as much as it needs to - so the workspace is only as large as the layout you currently have in it, and panning never extends beyond the actual open windows. Exposé in that context is only available when the content expands beyond the viewport, and it shrinks the content to fit the viewport. 

That means the expose would look less like this:
<img width="668" alt="screen shot 2019-01-29 at 5 18 40 pm 2" src="https://user-images.githubusercontent.com/6945691/51951670-582a5700-23ea-11e9-883d-e42d07f330d5.png">

and more like this:
[...]

PR #1724, opened 2019-01-29
Overlapping windows

aeschylus:
This is a first spike on the overlapping window system. It implements only the drag behaviour, positioning the windows in the workspace using a CSS 3D transform. The drag behaviour is attached to the top bar's jsx properties, and should probably be refactored, or some library used to cover potential browser compatibility issues. Perhaps a better way to implement the interaction would be mousedown + touch events.


aeschylus:
https://github.com/ProjectMirador/mirador/issues/1662

PR #1729, opened 2019-01-29
Start building out the workspace sidebar menu

### LINKED ISSUE #1728: Workspace sidebar menu -> "List all open windows"
Split out from #1726
- Take a first pass at implementing this Worskapce sidebar menu pop-out
- Display a list of all open windows in separate divs

Story: When I click on the "List all open windows" option, I should see a list of all open windows' manifest labels. 

No need to make the divs interactive; we'll punt on the click-and-center behavior. 
workspace-menu

list of windows

cbeer:
Fixes #1728

### DIFF
--- a/src/components/WorkspaceMenu.js
+++ b/src/components/WorkspaceMenu.js
@@ -24,72 +24,37 @@ export class WorkspaceMenu extends Component {
   constructor(props) {
     super(props);
     this.state = {
-      windowListAnchorEl: null,
-      settingsAnchorEl: null,
-      exportAnchorEl: null,
+      windowList: {},
+      settings: {},
+      exportWorkspace: {},
     };
-    this.handleWindowListClick = this.handleWindowListClick.bind(this);
-    this.handleWindowListClose = this.handleWindowListClose.bind(this);
-    this.handleSettingsClick = this.handleSettingsClick.bind(this);
-    this.handleSettingsClose = this.handleSettingsClose.bind(this);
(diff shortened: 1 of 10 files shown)


src/components/WorkspaceMenu.js
```diff
@@ -0,0 +1,107 @@
...
+  render() {
+    const { handleClose, anchorEl } = this.props;
+    const { windowListAnchorEl } = this.state;
+
+    return (
+      <>
+        <Menu id="workspace-menu" anchorEl={anchorEl} open={Boolean(anchorEl)} onClose={handleClose}>
+          <MenuItem
+            aria-haspopup="true"
+            onClick={(e) => { this.handleWindowListClick(e); handleClose(e); }}
```
cbeer: @jvine @ggeisler: I'm not sure what the intended interaction pattern for working with these submenus is; this PR just closes the parent menu and leaves the child menu, and I'm hoping we can get away with that for the near future (or, I suspect you ultimately want some controls from the child menu to close both, and some controls to only close the child menu... it's probably doable, but ugly, and would be nice to tackle after we have a better established pattern..)

jvine: Not sure if I'm understanding this correctly, but I think the pattern should be that the parent menu is closed when a child menu/dialog/whatever is opened. Is that what you mean by "closes the parent menu and leaves the child menu"?

cbeer: Perfect; I was worried there was some statefulness implied by the close button.

PR #1730, opened 2019-01-29
return null from getDerivedStateFromProps per React docs h/t @jkeck

mejackreed:
https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops

> It should return an object to update the state, or null to update nothing.

### DIFF
--- a/src/components/ThumbnailNavigation.js
+++ b/src/components/ThumbnailNavigation.js
@@ -36,7 +36,7 @@ export class ThumbnailNavigation extends Component {
         manifest: props.manifest,
       };
     }
-    return [];
+    return null;
   }
 
   /**


cbeer:
I've split that question out into #1734; as long as we're using derived state, it seems like this is a good change.

mejackreed:
Sure happy to add some more information about this. `manifest`'s state as it stands in this component (and all other `Window` derived components) can be in transition while rendering is happening. This is to be expected as we are not blocking rendering of the `Window` based on whether or not the `manifest` fetching is done. So we have a couple options.. 
 - do some checks in our render method based on presence of manifesto object and conditionally render
 - refactor to a container/presentational approach
 - do this get derived state from props approach
This seemed to be the most straightforward approach to get things working, since our state (canvases) are derived from how our `manifest` props are changing.

If we do end up removing manifesto from the state, this could be simplified for sure.

PR #1740, opened 2019-01-30
Divide <WindowTopBar/> into component and container.

charbugs:


__tests__/src/components/WindowTopBar.test.js
```diff
@@ -20,22 +21,10 @@ describe('WindowTopBar', () => {
...
-      .find('WithStyles(Typography)')
-      .dive()
-      .dive()
-      .text()).toBe('Fixture Label');
-    expect(topBar.find('WithStyles(Button).mirador-window-close'));
-  });
-
-  it('calls the removeWindow prop when the close button is clicked', () => {
-    topBar.find('WithStyles(Button)').simulate('click');
-    expect(mockRemoveWindow).toHaveBeenCalledTimes(1);
```
mejackreed: Do we need to maintain this test expectation somewhere?

charbugs: I assume that it is already tested in the material ui library. The <WindowTopBar> test should only make sure that the right function is provided to the <Button>. I set out to refactor the component tests after I finished the component-container-stuff.

charbugs: updated the test

PR #1746, opened 2019-01-30
Fixes #1743 by explicitly setting config mode

### LINKED ISSUE #1743: Create valid webpack configurations for dev and prod
the current webpack configuration does not validate.  (`mode` property is required)

Recommend separate configuration files for dev and prod. Relates to #1602

mejackreed:


cbeer:
Should we have some part of the build validate the webpack configuration, or is it obvious when it is broken?

mejackreed:
Webpack will raise failures in the console, and also raise warnings. Before #1746 a warning was raised about this issue specifically in some cases.

I think validation is an internal webpack thing (v2 -> ) and maybe we can just rely on the tool?

mejackreed:
I'm happy to make changes here to address the #1743 (and hopefully unblock others) who need a valid webpack config. I don't want to veer to far off the original intent of #1743 though.

Can we have a larger webpack / project setup discussion to move other pieces of your comment forward to additional tickets, or set some time to talk to face to face about that? I do hear you and recognize that a lot of projects use webpack-dev-server and that the project setup doesn't resemble other React projects. I'm confident we can figure out a way forward.

jkeck:
What is the status of this change?  Are all concerns expressed in other issues, allowing this to be shipped?  (just trying to clear the backup of PRs and want to ensure I don't cause any context loss)

christopher-johnson:
the PR does not fix the issue.  Webpack configuration validation in tools like `IntelliJ` and documentation generators like `styleguidist` continue to report invalid configuration (even though technically, the exported module configuration may be valid).

mejackreed:
@christopher-johnson can you post some of those invalid configuration reports here? I'm not seeing a problem with the default webpack validator.

christopher-johnson:
from intellij:
`09:32	Can't analyse webpack.config.js: coding assistance will ignore module resolution rules in this file.
			Possible reasons: this file is not a valid webpack configuration file or its format is not currently supported by the IDE.
			Error details:  Cannot read property 'mode' of undefined
`

you can try running [styleguidist](https://react-styleguidist.js.org/docs/getting-started.html) on your own.  it will only read a configuration object, not an array.

Like I said, the export configuration is not actually "invalid", it is just oddball.

mejackreed:
I'm going to close this to reduce our list of open PR's.

PR #1759, opened 2019-01-31
[(inactive) Spike] Overlapping windowing system using react-Rnd

aeschylus:
overlapping windows

[Netlify Demo](https://deploy-preview-1759--mirador-dev.netlify.com/__tests__/integration/mirador/)

This PR implements an overlapping windowing system using react-Rnd, which includes some of the same dependencies as react-grid-layout. It allows dragging and resizing of windows, and should allow us to implement grid-like behaviour on top of it in the future. It also registers the position and size of windows in the redux store as they update, allowing us to query the workspace for windows intersecting with the workspace viewport boundary, in response to which we would then pan (or "expand") the workspace.

Because it uses a more complete library, this PR doesn't have the bugs of the previous overlapping spike PR I put in, so there shouldn't be any jitter while dragging; the window should stick to the initial mouse position.

Still needs to implement zIndex updating for the focused window.


aeschylus:
WIP PR that integrates it with master/other window systems: https://github.com/ProjectMirador/mirador/pull/1815

camillevilla:
Closing this in favor of the implementation PR, #1815.

PR #1766, opened 2019-01-31
adds webpack dev and devServer configs

### LINKED ISSUE #1602: use webpack devServer / webpackHotDevClient in development mode
webpack has a devServer that works well for live reloading and hot module replacement. The dev server will always run on a different port than the e2e server so that testing can be done simultaneously with the dev environment. Optimally, a distinct webpack development configuration would accompany the implementation. The 'react-dev-utils/webpackHotDevClient' can be used for refreshing and hot updates.

### LINKED ISSUE #1743: Create valid webpack configurations for dev and prod

### LINKED ISSUE #1764: Remove unused config property from <App/>
config is set with the `setConfig` action, so this property can be removed

christopher-johnson:
this adds:
- valid webpack production config
- webpack devServer configuration with HMR and auto-browser refresh
- webpack development build
- a basic dev library development initialization file (sans plugin processing).
- local developer app configuration and default developer app configuration as json files
- .env support
 
this changes:
- build artifact names (removes "." and versioning semantics)
- refactors initialization utility functions from `MiradorViewer` class

closes #1743
closes #1602
closes #1764


mejackreed:
When running `npm i` and then `npm start` I'm getting an error:

```
Error: Cannot find module 'webpack-dev-server'
```

Should this be referencing the installed package via npm?

mejackreed:
Gotcha thanks. I'm going to drop the "Needs Quorum" label on this so we can get consensus and buy in from everyone.

__tests__/integration/mirador/index.html
```diff
@@ -8,7 +8,7 @@
   </head>
   <body>
     <div id="mirador"></div>
-    <script>document.write("<script type='text/javascript' src='../../../dist/mirador.min.js?v=" + Date.now() + "'><\/script>");</script>
+    <script>document.write("<script type='text/javascript' src='../../../dist/mirador3.js?v=" + Date.now() + "'><\/script>");</script>
```
aeschylus: We need to keep the name of the package free from version information in order to do releases. "3" just happens to be the semver landmark we've reached, but henceforth we want to be able to easily update to 4, 5, 6, without necessarily making these a part of the name.

https://semver.org/

package.json
```diff
@@ -58,7 +56,9 @@
     "babel-eslint": "9.0.0",
     "babel-jest": "^23.6.0",
     "babel-loader": "^8.0.4",
+    "babel-preset-react-app": "^7.0.0",
```
aeschylus: What does this do, and do we need it?

config/webpack.config.dev.js
```diff
@@ -0,0 +1,333 @@
...
+          {
+            test: [/\.bmp$/, /\.gif$/, /\.jpe?g$/, /\.png$/],
+            loader: require.resolve('url-loader'),
+            options: {
+              limit: 10000,
+              name: 'static/media/[name].[hash:8].[ext]',
+            },
+          },
+          // Process application JS with Babel.
+          // The preset includes JSX, Flow, and some ESnext features.
```
aeschylus: I'm concerned that the ES6 language or syntax features supported here won't match the features supported by our production transpilation.

config/webpack.config.dev.js
```diff
@@ -0,0 +1,333 @@
...
+          },
+        ],
+      },
+      // ** STOP ** Are you adding a new loader?
+      // Make sure to add the new loader(s) before the "file" loader.
+    ],
+  },
+  plugins: [
+    // Generates an `index.html` file with the <script> injected.
+    new HtmlWebpackPlugin({
```
aeschylus: Not a blocker, but can we extend this in the future to provide a development environment for our embedded usecases?

src/index-dev.js
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { Provider } from 'react-redux';
+import App from './containers/App';
+import createStore from './state/createStore';
+import * as actions from './state/actions';
+import './styles/index.scss';
+
+const store = createStore();
+const devConfig = {
```
aeschylus: Maybe not a blocker, but it might be good to allow this to be overridden with a local file on the dev's machine so people aren't updating this config all the time in PRs

webpack.config.js
```diff
@@ -42,61 +42,37 @@ const babelLoaderConfig = {
...
-    module: {
-      rules: [
-        eslintLoaderConfig,
-        babelLoaderConfig,
-      ],
-    },
+module.exports = {
+  mode: 'production',
+  entry: {
+    m3core: './src/index-core.js',
```
aeschylus: _c.f._: comments above regarding bundle naming. Ultimately these dist bundles will the the published package for mirador as a library, so we want their names to be stable as we move through the versions. See https://www.npmjs.com/package/mirador

aeschylus: It's also important that we be able to distinguish between umd builds and environment-specific builds, as well as minified and unminified builds.

package.json
christopher-johnson: https://github.com/ProjectMirador/mirador/blob/9d2a5f45a5add1b43475652b7bb1c73e72c7a176/config/webpack.config.dev.js#L165-L167
It provides some babel magic:
https://github.com/facebook/create-react-app/blob/master/packages/babel-preset-react-app/webpack-overrides.js

Should be optional for the current build.

src/index-dev.js
christopher-johnson: agreed.  I think that the devConfig should use ` .env`

config/webpack.config.dev.js
christopher-johnson: I do not follow you here.  the "embedded use case" is an implementation detail that should be in the scope of integration testing.  The build should not care how the exactly the library is deployed.  The dev environment should only be concerned with producing a universally deployed artifact.  Of course, the injection of that artifact into an html template makes some assumptions about containment (in that there is none).  Again, these containment edge cases can be covered with integration tests.

config/webpack.config.dev.js
christopher-johnson: babel-loader configuration is identical for production and dev.  No additional features are provided in dev.  https://github.com/ProjectMirador/mirador/blob/e469d5063f6c77b807a6af3f2b2e468bee5c6a1b/webpack.config.js#L38-L60

__tests__/integration/mirador/index.html
christopher-johnson: https://github.com/ProjectMirador/mirador/blob/e469d5063f6c77b807a6af3f2b2e468bee5c6a1b/webpack.config.js#L8-L9
OK?

src/index-dev.js
christopher-johnson: implemented `.env`.  If the custom config key (REACT_APP_LOCAL_MIRADOR_CONFIG) does not exist, falls back to `defaultAppConfig`.  These config files are `json`.

Here is the method.
https://github.com/ProjectMirador/mirador/blob/7a739167694684e2b94180a34ebffbca05584ba2/src/index-dev.js#L11-L19

mejackreed:
Thanks for the work on improving developer experience. I like the option of being able to hot reload. A few questions and concerns that I have get raised here:
 - As a development community, do we have the ability / resources to maintain to separate webpack configs?  Adding additional complexity here feels like a risk. Especially that the configs vary so significantly. Is there a way to mitigate this risk? A `+786 −162` diff to get "hot reloading" is feeling like a large project overhead. I realize that there are other pieces in play here to get that diff, I'm concerned about long term maintenance of these setup and config pieces.
 - Is there a way to implement hot reloading without refactoring `MiradorViewer.js`
 - Is there a way forward to consolidate json-server, http-server, webpack-dev-server to a single express config? Maybe using `webpack-dev-middleware` and `webpack-hot-middleware`

I realize these are a lot of questions. Happy to jump on a call early or late to discuss some more.

christopher-johnson:
comments inline:

> As a development community, do we have the ability / resources to maintain to separate webpack configs? 

I think a reasonable compromise approach here is to allow external devs and implementors to extend or override your default webpack build and e2e testing setup.  I think that this is as simple as referencing `start.js`, `build.js` and `test.js` scripts in the package.json.  These scripts could then reference config filename variables, etc. in local `.env` configurations.  Furthermore, there should be a MiradorAPI parent class that MiradorViewer (and other implementations) could extend.  (the app constructor and initializer utilities would then be subclassable)

I strongly advocate decoupling the e2e (i.e. jest-puppeteer) tests from the library `src` as well.   the current development environment makes assumptions about non-chunked artifact names and a legacy deployment scenario. This is an annoyance for external implementors who will not deploy that way.  Preferential configurations are also hard coded there for example. 

> Adding additional complexity here feels like a risk. Especially that the configs vary so significantly.  Is there a way to mitigate this risk? 

I have worked with a similar (and more complex) dev and prod webpack configurations for awhile, so I see no risk or complexity.  Separate configuration files work extremely well for those who will also deploy production artifacts with webpack instead of static html.   In that case, prod looks a lot more like dev.   

> Is there a way to implement hot reloading without refactoring MiradorViewer.js

yes.  but there would duplicate initialization routines to maintain until the e2e tests are decoupled from the developer environment and a MiradorAPI parent class is developed.  Those `processPlugins()` and `processConfig()` are implementation utilities.  I would argue that you should use an inherited API for initializers in `<MiradorViewer/>`.  `<MiradorViewer/>` is an **implementation** and does not work as an API .  

> Is there a way forward to consolidate json-server, http-server, webpack-dev-server to a single express config? Maybe using webpack-dev-middleware and webpack-hot-middleware

not recommended.  again, imo, the webpack dev environment should have no binding dependencies on the e2e testing framework or production artifacts.  that is a circular dependency.  the e2e tests should only depend on a local or remotely npm published snapshot or production library artifact. 

In summary, I hope you can see that there is a lot more intended with this PR than HMR.   The main idea is to not depend on a production build and e2e test assumptions for local dev server deployment.  

I would split this PR into these sub PRs like this:

1. add .env support and env bootstrapping.
2. change production `build` task to use script (would include all proposed fixes to current webpack configuration)
3. add start script to bootstrap a development environment.  (status quo would remain the same, except that there would be `.env` overrides).  overrides could potentially be sourced from node_modules.
4. add test script.

aeschylus:
Closing this for now because of how much the codebase has moved, and I think a lot of this has been broken down into other tickets/PRs

PR #1779, opened 2019-02-01
Add OSD-internal state information to the workspace state; fixes #1644

### LINKED ISSUE #1644: Add zoom level of OpenSeaDragonViewer to redux state for each window

cbeer:


cbeer:
Hm. Can you take another look at https://deploy-preview-1779--mirador-dev.netlify.com/__tests__/integration/mirador/? I think this only started happening when I started comparing props to prevProps instead of going directly to the OSD state. If it doesn't create the same problem, I'll go in and fix up the failing test.

mejackreed:
@cbeer  That seems to have fixed things for me.

PR #1793, opened 2019-02-02
Disable ESlint rule: react/forbid-prop-types; fixes #1781

### LINKED ISSUE #1781: Disable react/forbid-prop-types in eslint config
It seems like we have a lot of `eslint-disable-line react/forbid-prop-types` piling up. Many of these are for objects like `workspace`, `window`, and `manifest`, which we use ... a LOT! 

Disabling the rule in our `.eslintrc` might be nice for devs. 

## Resources
- [eslint-plugin-react docs](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/forbid-prop-types.md) for `forbid-prop-types`
- [React docs](https://reactjs.org/docs/typechecking-with-proptypes.html#proptypes): see the `ObjectOf` and `shape` examples. Perhaps we'd introduce `shape` once these objects are a bit more stable?

camillevilla:
Closes #1781 - see resources and discussion in issue

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -12,6 +12,7 @@
   "rules": {
     "no-console": "off",
     "react/jsx-filename-extension": [1, { "extensions": [".js", ".jsx"] }],
+    "react/forbid-prop-types": off,
     "require-jsdoc": ["error", {
       "require": {
         "FunctionDeclaration": true,
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -33,7 +33,7 @@ class App extends Component {
 }
 
 App.propTypes = {
-  workspace: PropTypes.object, // eslint-disable-line react/forbid-prop-types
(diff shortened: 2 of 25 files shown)


cbeer:
I was reading about the eslint rule, and they suggest some alternative approaches:

> By default this rule prevents vague prop types with more specific alternatives available (any, array, object), but any prop type can be disabled if desired. The defaults are chosen because they have obvious replacements. any should be replaced with, well, anything. array and object can be replaced with arrayOf and shape, respectively. [1]

I'm happy to take a look at what following their recommendations looks like.


[1] https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/forbid-prop-types.md

camillevilla:
Closing this in favor of #1800 or ... doing nothing.

PR #1795, opened 2019-02-02
Remove unused Display component.

jkeck:


christopher-johnson:
try merging master.  It is comparing your outdated commit tip against
master which was before #1743 was merged. hth

Issue #1585, opened 2019-01-13
Setup i18n internationalization for strings in project


mejackreed:
There are several ways of implementing this. Ideally we would like an approach that could have the following traits:

 - [ ] Configured language by the Mirador adopter in the configuration
 - [ ] All locale files shouldn't need to be loaded by every instance (only configured language(s))

Possible SPIKEs to be investigated: 
- [ ] https://github.com/alibaba/react-intl-universal
- [x] https://github.com/i18next/react-i18next (#1796)
- [ ] Roll our own (see globocom/megadraft#214 for possible concerns)

bodo22:
Maybe it's worth considering a more react-independent library for i18n, e.g. https://www.i18next.com/ .
My 2 cents would be that adding, editing and updating i18n should be as easy as possible, e.g. by using an industry standard such as gettext and .po files, or plain old json

jkeck:
We've taken the `react-i18next` approach here (as implemented initially in #1796).

PR #1796, opened 2019-02-02
Use i18next / react-i18next for internationalizing strings

jkeck:
Part of #1585 

This is a [SPIKE] for analysis on the i18n library to use.

I have only translated one component for demonstration, but if we decide to go down this route I'll i18n the rest of the strings around the app so far.

## Open Questions/TODOs
- [x] Should the `withNamespaces` happen in the container or in the presentation component?  (It may make testing difficult in our presentational components)
- [x] Fix tests
- [x] Should the `t` prop function be required &  have a default set to simply return the key?
  * ~~If we do that, then if somebody doesn't setup i18n correctly in their component, they'll simply get the key returned (which might be confusing with if the key doesn't exist, maybe that's okay).~~
  * ~~If we don't do that, then we end up passing the prop in our tests, so our tests would end up passing where our component could fail to render if we break i18n integration (which seems like a bad thing).~~
  * Going w/ defining a defaultProp for now.  Can be refactored later as needed.
- [x] Complete i18n migration (if we decide to use this library)


cbeer:
It looks reasonable to me; if nothing else, it sets us up to be disciplined about keeping strings out of the code.

src/containers/WindowTopBar.js
```diff
@@ -2,7 +2,7 @@ import { compose } from 'redux';
 import { connect } from 'react-redux';
 import * as actions from '../state/actions';
 import miradorWithPlugins from '../lib/miradorWithPlugins';
-import WindowTopBar from '../components/WindowTopBar';
+import TranslatedWindowTopBar from '../components/WindowTopBar';
```
charbugs: Note that we're going to reintroduce the \<ConnectedFooBar/\> pattern here. I see that the problem comes with the `withNamespace()` HOC. At the end of [this wiki article](https://github.com/ProjectMirador/mirador/wiki/M3-Keep-store-connected-components-in-separate-files#problems) I outlined the problem of additional HOCs. In this case I would tend to place the `withNamsepaces()` code in the container rather than the component.

jkeck: I struggled with if i18n was presentational or functional (I can see arguments for both).  Happy to call it functional and put it in the container.

cbeer: @charbugs - is it possible to just change the import statement back to `WindowTopBar` and call it good? I'm a little worried that pushing i18n into the container is going to make it challenging to test if we have to stub out i18n in component tests.

charbugs: You're right, there is no clear distiction. As I said, I'am not sure how to handle this additional-hoc-cases. Perhaps we should generally put HOCs in the containers...

charbugs:
@jkeck 
> If we don't do that, then we end up passing the prop in our tests, so our tests would end up passing where our component could fail to render if we break i18n integration (which seems like a bad thing).

I would say, the same is true for every other prop in our **unit tests** because we test the components in isolation and mock out each prop. In the unit tests we go like: *Suppose the right props get passed in -- does the component behave as it should?* To check if serveral components interact correctly and pass the right props to each other would be a matter of integration tests. So I suggest to require the `t` prop without specifying defaults and mock it out in tests. 

Is it intended to test i18n?

cbeer:
It seems like a coin-flip to me: in the fully-connected case, the default property won't matter because we'll get a real i18n object; in testing the unconnected component, if we don't provide a default implementation, we'll end up making up some default implementation in all the tests instead.

I'd say 👍 to this implementation, and we can always incrementally replace it with a required function if/when it becomes an issue.

cbeer:
I'm going to ship this PR as-is, hopefully without too much controversy. Once we have consensus on how to manage those properties, it shouldn't be too disruptive to change it in a future PR.

charbugs:
@cbeer 

I still favor the require-without-default approach. As you said in unit tests we need to mock this prop. But when to application is running this approach has an important advantage: React will show an warning when something tries to render that component without providing the prop. This is not the case when we define a default prop because a default prop makes `isRequired` obsolete. (I consider this as an design failure of the proptypes framework but at least eslint complains if you try to define a default for a required prop.) I think this require check of React is a useful kind of integration test and we should not lever it out.

PR #1800, opened 2019-02-04
Address eslint forbidden prop types shapes

### LINKED ISSUE #1781: Disable react/forbid-prop-types in eslint config

cbeer:
Fixes #1781 (by giving shapes to forbidden prop types).

I don't know if this is an idiomatic way to do this, but I also agree with @camillevilla that we're probably not a point where we can strictly define our types yet.

### DIFF
--- a/__tests__/src/components/ViewerNavigation.test.js
+++ b/__tests__/src/components/ViewerNavigation.test.js
@@ -11,7 +11,7 @@ describe('ViewerNavigation', () => {
     previousCanvas = jest.fn();
     wrapper = shallow(
       <ViewerNavigation
-        canvases={[1, 2]}
+        canvases={[{ id: '1' }, { id: '2' }]}
         nextCanvas={nextCanvas}
         previousCanvas={previousCanvas}
         window={{ canvasIndex: 0 }}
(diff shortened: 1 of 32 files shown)


aeschylus:
This seems like a reasonable solution to me compared to more drastic ways of addressing this such as using typescript. Ultimately having types or shapes is likely to make generating useful documentation quite a bit easier.

I'd love to hear other opinions on the problem. I confess the heresy that, although I understand the theoretical appeal of defined types, "do nothing about this" also seems fine to me, since I haven't personally experienced much of a practical issue with passing untyped objects, since it's typically apparent based on the uses and implementation of the functions that process them what ought to be given/returned. Maintainability-wise it's probably a good idea though to address it some way or other.

PR #1807, opened 2019-02-05
Add withMiradorAPI HoC

### LINKED ISSUE #1802: Reimplement init utils / move MiradorViewer constructor into a HoC (withMiradorAPI)
first PR related to this: https://github.com/ProjectMirador/mirador/pull/1766#issuecomment-460274751
```javascript
/**
* instance method signature
*/
@withMiradorAPI(params)
class WrappedComponent extends React.Component{}
```
or

```javascript
const MiradorViewer = withMiradorAPI(store, config)(WrappedComponent)
```
`withMiradorAPI` HoC will include a constructor and `processConfig()` and `processPlugins()` related functions.   
`store` and `config` are parameters.

christopher-johnson:
refactors init methods from default `MiradorViewer` implementation 
simplifies initialization and allows for multiple implementations.
closes #1802

### DIFF
--- a/src/lib/withMiradorAPI.js
+++ b/src/lib/withMiradorAPI.js
@@ -1,0 +1,134 @@
+import React from 'react';
+
+import * as actions from '../state/actions';
+import createRootReducer from '../state/reducers';
+
+/**
+ * HoC that initializes a Mirador Application
+ * @param store
+ * @param config
+ * @returns {function(*): {contextType?: React.Context<any>,
+ * new(props: Readonly<P>): HOC, new(props: P, context?: any): HOC, prototype: HOC}}
+ */
+export default function withMiradorAPI(store, config) {
+  // TODO validate config schema before processing / handle invalid or missing configuration
+  // validateConfig(config);
+
+  doInitialization();
+
+  /**
+   * doInitialization
+   */
+  function doInitialization() {
+    processPlugins();
+    dispatchSetConfig(config);
+    fetchManifests();
+    addWindows(config);
+  }
+
+  /**
+   * dispatchSetConfig
+   */
+  function dispatchSetConfig() {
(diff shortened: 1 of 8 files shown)


cbeer:
What other implementations are expected? Knowing that would help me evaluate whether this refactor gets us close to supporting them.

src/lib/withMiradorAPI.js
```diff
@@ -0,0 +1,134 @@
...
+  function dispatchFetchManifest(uri) {
+    store.dispatch(actions.fetchManifest(uri));
+  }
+
+  /**
+   * addWindows
+   */
+  function addWindows() {
+    const thumbnailPositions = getThumbnailNavigationPositions(config);
+    thumbnailPositions.forEach((thumbnailNavigationPosition, index) => {
```
cbeer: One thing that was nice about the previous implementation is that we were clearly looping over a list of window configurations to turn them into windows in the workspace; I feel like this refactor has lost some of that clarity.

christopher-johnson: https://github.com/ProjectMirador/mirador/blob/2eee9d41d427d60c3d130f3ee5d1b6ffd1c68666/src/lib/withMiradorAPI.js#L56 will go away when the config validation occurs.  thumbnailNavigation.defaultPosition could be a default value.  convention over configuration is a better pattern.   I suggest fixing https://github.com/ProjectMirador/mirador/blob/2eee9d41d427d60c3d130f3ee5d1b6ffd1c68666/src/lib/withMiradorAPI.js#L77-L87 could add clarity to the whole thing.

there is no error handling or validation in the previous "implementation".  the strategy here is to handle errata and set configuration defaults with schema validation at the API layer.   

FYI: the implementation is where `ReactDOM.render(..)` is.

PR #1813, opened 2019-02-05
Use Contexts to pass a workspace reference to popovers that need them

### LINKED ISSUE #1785: WindowTopBar Buttons don't work in fullscreen
Clicking the drop-down icon in the top bar does nothing in fullscreen.

jkeck:
Closes #1785 

See example below which is in fullscreen.
screen shot 2019-02-05 at 2 21 01 pm

### DIFF
--- a/__tests__/src/components/Workspace.test.js
+++ b/__tests__/src/components/Workspace.test.js
@@ -3,6 +3,7 @@ import { shallow } from 'enzyme';
 import WorkspaceMosaic from '../../../src/containers/WorkspaceMosaic';
 import Window from '../../../src/containers/Window';
 import Workspace from '../../../src/components/Workspace';
+import { WorkspaceContext } from '../../../src/contexts';
 
 const windows = { 1: { id: 1 }, 2: { id: 2 } };
 
@@ -15,7 +16,9 @@ describe('Workspace', () => {
 
       expect(wrapper.matchesElement(
         <div className="mirador-workspace">
-          <WorkspaceMosaic windows={windows} />
+          <WorkspaceContext.Provider>
+            <WorkspaceMosaic windows={windows} />
+          </WorkspaceContext.Provider>
         </div>,
       )).toBe(true);
     });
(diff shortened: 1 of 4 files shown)


aeschylus:
It makes me nervous to introduce contexts at all, since we're already using redux. I know there's a bug with fullscreen, but at least in the browser's native API, this may be a problem with MUI's fullscreen component. Have we tried using the browser's native fullscreen that reuses the existing element?

jkeck:
I'm not sure if the .09 drop in coverage is legit.  Happy to work with somebody if this reduction seems problematic.

cbeer:
As a point of clarity, the fullscreen stuff isn't from MUI, but a separate react component library. The issue we're seeing is the MUI popovers, by default, are written into `document.body`, but the fullscreen implementation is making just the workspace fullscreen. 

I think @jkeck and I are both a little uncomfortable trying to push HTML DOM elements references through redux; it felt a little suspect and different from the other application state we're storing there. 

Maybe we can chat after standup?

aeschylus:
Knowing that MUI adds its popover to the DOM, and having heard the alternative approaches, **I'm for merging this now**. I'm not sure where else to ticket this yet, but I imagine it will come up in integration tests: it is a concern to have our dependencies appending things to the parent document body. If we were implementing a dropdown menu from scratch, it would not be necessary, so it seems possible that we would be able to find a library that didn't do this.

For trivia's sake, here is a breakdown of some concerns with the contexts API when also using other libraries that use it: https://medium.com/dailyjs/reacts-%EF%B8%8F-new-context-api-70c9fe01596b

cbeer:
I've rebased this, but unfortunately I think it's stalled out until enzyme fixes their API around React 16 Contexts..

jkeck:
@mejackreed thoughts?

mejackreed:
I'm 👍 to move this forward and ship.. but do we now need to wait on Enzyme to catch up?

PR #1815, opened 2019-02-05
Add overlapping windowing system

aeschylus:


src/components/WorkspaceFreeform.js
```diff
@@ -0,0 +1,64 @@
...
+        }}
+        onDragStop={(e, d) => {
+          setWorkspaceViewportPosition({ x: d.x, y: d.y });
+        }}
+        cancel={`.${ns('window')}`}
+        className={ns('workspace')}
+      >
+        {
+          Object.values(windows).map(window => (
+            <Window
```
mejackreed: Could this be a seam to render the `draggableWorkspaceWindow` (perhaps as a component) with a child of `Window`? Could that remove the conditional from `Window`?

aeschylus: Yeah, that's probably nicer. There was a lot of reshuffling. I'll make this update.

package.json
```diff
@@ -44,13 +45,14 @@
     "node-sass": "^4.9.2",
     "openseadragon": "^2.4.0",
     "prop-types": "^15.6.2",
-    "react": "^16.8.3",
-    "react-dom": "^16.8.3",
+    "react": "^16.7.0",
```
mejackreed: Do we need to specifically use 16.7 ? or is this just maybe rebase cruft?

package.json
```diff
@@ -31,6 +31,7 @@
     "@material-ui/core": "^3.9.1",
     "@material-ui/icons": "^3.0.2",
     "@researchgate/react-intersection-observer": "^0.7.4",
+    "acorn": "6.1.1",
```
mejackreed: Where is acorn being used?

package.json
```diff
@@ -85,6 +87,7 @@
     "jest": "^24.1.0",
     "jest-fetch-mock": "^2.1.1",
     "jest-puppeteer": "^4.0.0",
+    "jsdom": "13.2.0",
```
mejackreed: Should we remove the eslint-inline disable here now?: https://github.com/ProjectMirador/mirador/blob/master/setupJest.js#L3

src/containers/Window.js
```diff
@@ -2,12 +2,13 @@ import { compose } from 'redux';
...
 import { Window } from '../components/Window';
 
+
 /**
  * mapStateToProps - used to hook up connect to action creators
  * @memberof Window
  * @private
  */
-const mapStateToProps = ({ manifests, windows }, props) => ({
+const mapStateToProps = ({ config, manifests, windows }, props) => ({
```
mejackreed: Is config needed here still?

__tests__/src/actions/workspace.test.js
```diff
@@ -49,4 +49,29 @@ describe('workspace actions', () => {
...
+      expect(actions.setWorkspaceViewportPosition({
+        x: 20,
+        y: 20,
+      })).toEqual(expectedAction);
+    });
+  });
+  describe('toggleWorkspaceExposeMode', () => {
+    it('should set the exposeMode to true', () => {
+      const expectedAction = {
+        type: ActionTypes.TOGGLE_WORKSPACE_EXPOSE_MODE,
```
mejackreed: Is `TOGGLE_WORKSPACE_EXPOSE_MODE` just a placeholder in the state to enable future work?

package.json
aeschylus: It's a peer dependency that was uncovered when I was forced to run things on a new machine.

src/styles/index.scss
```diff
@@ -35,6 +43,8 @@
   }
 
   &-window {
+    width:100%;
+    height:100%;
```
mejackreed: Is this perhaps a duplicate?

src/styles/index.scss
```diff
@@ -1,21 +1,29 @@
 @import 'variables';
 
 .mirador {
-  &-app {
-    height: 100%;
-    position: relative;
-    width: 100%;
+  &-viewer {
+    position: absolute;
```
mejackreed: I'm seeing some style warnings in our `npm run lint` task. Should we fix those now or spin off a ticket for that?

package.json
mejackreed: I'm not sure i understand that. Wouldn't the package list that as a dependency if it was needed? Why wouldn't our CI/CD catch that?

mejackreed:
@aeschylus not sure if this is a big fix or not, but I also just noticed that now our multiple "embedded" versions are broken: https://deploy-preview-1815--mirador-dev.netlify.com/__tests__/integration/mirador/embedding.html

aeschylus:
The styling on the integration tests is unrelated. The mirador viewer is meant to fill the bounds of its parent. Therefore the parent will need to have `position:relative` or `position:absolute` to provide the coordinate system. `overflow:hidden` should probably also be required, but that can be addressed elsewhere. In general there are more considerations around how we want to treat the configured parent box.

src/styles/index.scss
mejackreed: I think `height` was specified twice in the css. It seems resolved now.

mejackreed:
I created a spin off ticket to fix this: https://github.com/ProjectMirador/mirador/issues/2007

mejackreed:
Adding this to the PR so its not lost in comments on Slack:
 ### To switch workspace type
```javascript
miradorInstance.store.dispatch(miradorInstance.actions.updateConfig({
  workspace: {
    type: "elastic"
  }
}))
```

PR #1820, opened 2019-02-06
Refactor <App/> component.

charbugs:

PR #1821, opened 2019-02-06
Refactor <Workspace/> component.

charbugs:


charbugs:
Same as #1820 : travis ci tests fail here. but all tests passed on my local machine.

PR #1826, opened 2019-02-07
Add .env support for multiple environments

### LINKED ISSUE #1825: Add .env support for multiple environments
[dotenv](https://www.npmjs.com/package/dotenv) provides access to locally defined configuration variables in `process.env`.  This is necessary to provide multiple build, test and start environments.  

See https://github.com/ProjectMirador/mirador/pull/1766#issuecomment-460274751

christopher-johnson:
adds `dotenv` and `dotenv-expand` packages to devDependencies
adds env initializer script that reads .env files
closes #1825

### DIFF
--- a/config/env.js
+++ b/config/env.js
@@ -1,0 +1,64 @@
+const fs = require('fs');
+const path = require('path');
+const paths = require('./paths');
+
+delete require.cache[require.resolve('./paths')];
+
+const NODE_ENV = process.env.NODE_ENV;
+if (!NODE_ENV) {
+  throw new Error(
+    'The NODE_ENV environment variable is required but was not specified.'
+  );
+}
+
+const dotenvFiles = [
+  `${paths.dotenv}.${NODE_ENV}.local`,
+  `${paths.dotenv}.${NODE_ENV}`,
+  NODE_ENV !== 'test' && `${paths.dotenv}.local`,
+  paths.dotenv,
+].filter(Boolean);
+
+dotenvFiles.forEach((dotenvFile) => {
+  if (fs.existsSync(dotenvFile)) {
+    require('dotenv-expand')(
+      require('dotenv').config({
+        path: dotenvFile,
+      })
+    );
+  }
+});
+
+const appDirectory = fs.realpathSync(process.cwd());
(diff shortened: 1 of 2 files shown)


mejackreed:
I've read the issue in #1825,  can you say more about the Mirador usage of this?

How do you envision this being used by Mirador adopters? Or is this a project setup need?

We use similar setups in many of our applications, but I'm trying to determine how this fits in Mirador just yet.

christopher-johnson:
The default puppeteer test and production webpack configurations and assumptions made therein are blocking external implementations who want to test, build and deploy the <App /> component (i.e. the Mirador library) using alternative means and methods.  They should be able to be overridden.  See [example](https://github.com/ProjectMirador/mirador/commit/a29d9ae249a47980c06b6b37e6a7752ab1d135a9) of using a `.env ` variable to override the jest configuration.

mejackreed:
So is the scenario here that a user of Mirador wants to use Mirador by?:
 - Cloning or referencing the project using git submodules
 - Then running Mirador's CI system using the other projects own CI system?
 - Mirador adopter then can build a custom version that they need ?

Trying to understand the scenario in play here.

christopher-johnson:
This is a stopgap measure to facilitate external development only.    

The best case scenario going forward is that you produce an npm package that exports the root Mirador React Component and an API to initialize it.  

This would allow implementations to do something like:
```javascript
import { App } from 'mirador'
import withMiradorAPI from 'mirador'

const CustomMirador = withCustomization(withMiradorAPI(store, config)(App))
const element = ( 
   <Provider store={store}> 
     <CustomMirador /> 
   </Provider> 
 );
```
everything else would be in a separate repository.  This assumes that none of the global default Mirador implementation would be in that npm package.

mejackreed:
> stopgap measure to facilitate external development only

I'm not sure I still understand that scenario? Can you say more?

christopher-johnson:
I qualify this:

>  to facilitate external _pre-release_ development only.

undefined release schedule and publishing are related issues here, How / if / when  to publish Mirador as a `React.Component`?   

it could be called `mirador-react-app`.  It could include a default store configuration initialize API or that could be published separately.  This is actually trivial, I submit a webpack config and npm task and PR for this.

PR #1828, opened 2019-02-07
Adds test configuration override

christopher-johnson:
current development and CI testing behavior:
- test script runs production build **first** since the integration tests depend on it
- integration tests run always by default

desired development and CI testing behavior
- linting and unit tests not dependent on production build
- production build second (in CI)
- e2e tests third (in CI) or locally can be wired into a dev server for continuous dev artifact watch.


package.json
```diff
@@ -4,11 +4,14 @@
...
   "scripts": {
-    "lint": "node_modules/.bin/eslint ./ && node_modules/.bin/sass-lint -v ./src/styles/**/*",
+    "lint": "eslint ./ && sass-lint -v ./src/styles/**/*",
     "server": "node_modules/.bin/http-server",
-    "test": "npm run build && npm run lint && npm run size && jest -c jest.json",
-    "test:watch": "jest -c jest.json --watch",
+    "pretest": "npm run lint",
+    "test": "node scripts/test.js",
+    "test:watch": "npm run test -- --watch",
+    "prebuild": "npm run test && npm run size",
```
jkeck: One thing that will cause is the inability to `build` on a fresh clone since this will run tests, but those tests will fail since the project has never been built.

This is why the netlify deploy is broken on this PR, AFAICT, it run `npm install && npm run build` on a fresh clone of the project.

christopher-johnson: I fixed this on travis by running `npm run build:ci` that will not run the prebuild test hook.  only unit tests should run in prebuild `test`.  Running the integration tests with the default test script is unconventional.  These should be either optional (the intent of this PR) or in a separate task (e.g. `test:integration`).

christopher-johnson:
irrelevant

PR #1830, opened 2019-02-07
Refactor <WindowIcon/> component.

charbugs:
In this PR I introduced [selectors](https://www.saltycrane.com/blog/2017/05/what-are-redux-selectors-why-use-them/), i.e. functions that are in a way the counterpart of redux actions. While actions usually trigger a write event to the store, selectors are meant to read data from the store and usually transform them for their consumers. 

I think selectors can help to decouple the render logic from the data managment. Also, there are this [reselect](https://redux.js.org/recipes/computing-derived-data) library that can create memoized selectors, which can lead to better performance. I didn't use reselect for the moment only vanilla javascript, but I think would be easy to add reselect later.

### DIFF
--- a/src/state/selectors/index.js
+++ b/src/state/selectors/index.js
@@ -1,0 +1,22 @@
+
+/**
+* Return the manifest that belongs to a certain window.
+* @param {object} state
+* @param {String} windowId
+* @return {object}
+*/
+export function getWindowManifest(state, windowId) {
+  return state.windows[windowId]
+    && state.windows[windowId].manifestId
+    && state.manifests[state.windows[windowId].manifestId];
+}
+
+/**
+* Return the logo of a manifest or null
+* @param {object} manifest
+* @return {String|null}
+*/
+export function getManifestLogo(manifest) {
+  return manifest.manifestation
+    && manifest.manifestation.getLogo();
+}
--- a/__tests__/src/components/WindowIcon.test.js
+++ b/__tests__/src/components/WindowIcon.test.js
@@ -2,40 +2,27 @@ import React from 'react';
 import { shallow } from 'enzyme';
 import WindowIcon from '../../../src/components/WindowIcon';
 
+/** createWrapper */
+function createWrapper(props) {
+  return shallow(
+    <WindowIcon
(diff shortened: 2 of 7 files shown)


src/state/selectors/index.js
```diff
@@ -0,0 +1,22 @@
+
+/**
+* Return the manifest that belongs to a certain window.
+* @param {object} state
+* @param {String} windowId
+* @return {object}
+*/
+export function getWindowManifest(state, windowId) {
```
christopher-johnson: I think it might make more sense for state to have a `registry `of manifests in store with props including current window id(s) and manifest values denormalized.  The lookup would filter for a manifest v5 uuid.  It could look like this:

The selector simply reduces using a uuidv5 (e.g. xxxxxxxx-xxxx-Mxxx-Nxxx-xxxxxxxxxxxx) created from the manifest uri as the filter.
```
{
  "manifests": {
    "xxxxxxxx-xxxx-Mxxx-Nxxx-xxxxxxxxxxxx": {
      "manifest": "http://some.manifest",
      "windows": [],
      "canvases": [],
      "iiifProperty": "String, etc."
    }
  }
}
```

aeschylus: Would this be to avoid the state digging? I can see some advantages to this for request reasons. In terms of this particular lookup, though, wouldn't we then need to loop over all the manifests to get the answer instead of using the id the window has attached to itself on creation? Honestly I can see having multiple selectors for each of these purposes.

I think the PR as a whole is good for now, given the way the manfiestLogo component is implemented above, and for setting up a selector pattern to let us call on the store a bit less.

christopher-johnson: There is no loop, it works like a map, you basically just select the manifest object by index.  

I would suggest a HoC `withManifest(WrappedComponent)`, that gets and sets a `manifestId `prop for a component.  Then, the component can get whatever values it wants from store registry that relates to the manifest using a selector passing the manifest `uuid `as the key.  

The strategy is to decouple window state from the content as much as possible.  The window should just do its job of being a window and doing window actions and not try to manage its content props.

PR #1841, opened 2019-02-08
Refactor <WindowThumbnailSettings/> component.

charbugs:


charbugs:
As far as I could figure out the integration tests fail because the app is unable to fetch the manifest from https://media.nga.gov/public/manifests/nga_highlights.json

mejackreed:
It looks like the cert expired for the NGA manifest :(

charbugs:
my browser said similar

PR #1845, opened 2019-02-08
Refactor ThumbnailNavigation to receive a list of canvases; fixes #1734

### LINKED ISSUE #1734: Do we need to use derived state
From @aeschylus:
> Can you provide a little more context for this. Why is derived state even being used?

> https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html

> Happy to merge now, since this is all in flux, but maybe we can ticket a refactor of this. It seems like we should be able to set up the redux state differently in order to avoid this. For example, would this go away once manifesto is removed from the redux store? I might be missing something obvious, bu at what point is the manifest ever changing in the lifecycle of a window?
> @jbaiter @charbugs - opinions?

Context: https://github.com/ProjectMirador/mirador/pull/1730

cbeer:
I've tried to apply the new selector pattern that @charbugs has been applying to `ThumbnailNavigation`; I think this addresses our concerns in #1734, unless there's an untested side effect that the previous approach dealt with?

### DIFF
--- a/__tests__/src/components/ThumbnailNavigation.test.js
+++ b/__tests__/src/components/ThumbnailNavigation.test.js
@@ -19,11 +19,9 @@ describe('ThumbnailNavigation', () => {
     );
     wrapper = shallow(
       <ThumbnailNavigation
-        manifest={{
-          id: 'http://foo',
-          manifestation: manifesto.create(manifestJson),
-          isFetching: false,
-        }}
+        canvases={
+          manifesto.create(manifestJson).getSequences()[0].getCanvases()
+        }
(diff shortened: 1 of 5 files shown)


charbugs:
I think the refactoring did a got job and it shows the advantage of selectors. We got rid of a lot of transformation code in the \<ThumbnailNavigation/\> component. The component is now simpler, easier to read and has less assumptions about the redux store. The state transformation is now a matter of the selector that belongs conceptually to the store, i.e. the store provides a tailored function to the component and possibly to other components in the future.

Issue #1790, opened 2019-02-02
Book View --> canvas view, window menu option


camillevilla:
This ticket is a first pass at implementing Book View. Additional tickets may be split out as necessary.

![book view](https://user-images.githubusercontent.com/5402927/52156125-2eb43a00-263b-11e9-925f-6f8d5625cb6a.png)

![book view with menu options exposed](https://user-images.githubusercontent.com/5402927/52156127-307dfd80-263b-11e9-94af-ffc03452e5f1.png)

### Acceptance Criteria
- [x] A user can select "Book" from the Window top menu --> View options #1850
  - [x] The "View" selections menu is clearly separated from the Thumbnail options ([MUI Dividers](https://material-ui.com/demos/dividers/)?)

Once Book view is toggled 
  - [x] a user sees two canvases connected
  - [x] The user experiences the two pages as a single canvas (e.g. they zoom and pan as a single object)
  - [x] The window size remains the same
  - [x] Regardless of previous zoom, the viewer zooms out so that both pages can be seen at once
  - [x] Sidebars and companion windows that were previously open remain open (ignore if companion windows haven't been implemented yet)
- [x] Manifests with an odd number of canvases display the first page as a single canvas (could be split out)

### Out of Scope
- Book view thumbnails in the thumbnail strip (#1791)

### Resources
- [Interactive designs](https://github.com/ProjectMirador/mirador-design/wiki/Mirador-3-Interactive-Designs) > views and thumbnails
- [An example of Mirador 2 with book view enabled](https://parker.stanford.edu/parker/catalog/ky326yt7099)

PR #1850, opened 2019-02-11
Setup the ability to select single vs bookview

mejackreed:
Part of #1790 

bookviewselection

This wires up the UI components for selecting the different view types.

I split this out in to its own PR, as I think implementing #1790 in a single PR is going to be quite large.


mejackreed:
I'm using `view` right now as it directly relates to the user interface element in the designs, rather than a direct translation from the specs. It definitely seems like there will be some more work needed to support how `viewingHint` and `behavior` works here, but I'm considering that out of scope for this work as outlined in #1790.

__tests__/src/selectors/index.test.js
```diff
@@ -96,6 +97,7 @@ describe('getThumbnailNavigationPosition', () => {
   });
 });
 
+<<<<<<< HEAD
```
cbeer: Merge conflict?

mejackreed: ah.. nice catch :)

src/state/actions/window.js
```diff
@@ -27,6 +27,7 @@ export function addWindow(options) {
     thumbnailNavigationPosition: 'bottom', // bottom by default in settings.js
     xywh: [0, 0, 400, 400],
     rotation: null,
+    view: 'single',
```
camillevilla:
```suggestion
    viewType: 'single',
```
This is a small suggestion, but we might save ourselves a tiny headache down the line by having this config key match our action types name.

mejackreed: If its ok, can we hold off on this name change? I agree its suboptimal right now. Perhaps we can brainstorm to figure out something better for internal/and external usage?

Issue #1833, opened 2019-02-08
Design for load window plugin


ggeisler:
This is a story ticket that provides an overview of a proposed load window plugin. The specific actionable tickets are:

- Implement 'List view' of load window plugin
- Add 'Layout' option to load window tools menu
- Implement 'Card view' of load window plugin

 See the [interactive mockup](https://xd.adobe.com/view/477b3833-0ae4-47fa-6032-15d9b7bcd9f0-314a/) for a more complete sense of how the plugin is intended to work.

### Invoking the plugin
The FAB button with the plus icon in the workspace sidebar invokes the plugin:

<img width="89" alt="load-button" src="https://user-images.githubusercontent.com/101482/52449093-43894580-2af3-11e9-8468-4ba44a307255.png">

---

### List view
The initial view is the list view, which looks like this (tablet):

<img width="766" alt="tablet-list-view" src="https://user-images.githubusercontent.com/101482/52449114-5a2f9c80-2af3-11e9-9eda-48e25166f04e.png">

---

Selecting the thumbnail or title of a resource would load that resource into a new window in the workspace. (Unlike the catalog mockup proposed last year, we're not suggesting affordances for previewing a resource or opening multiple resources at once with this plugin, in an effort to keep it simpler.)

The layout of list view is very similar across device sizes; the only difference is that below a specific breakpoint (e.g., in mobile) the institution elements will need to wrap below the resource thumbnail and title.

Both list view and card view include an affordance for loading a resource via a URL:

<img border="1" width="373" alt="mobile-add-resource-panel" src="https://user-images.githubusercontent.com/101482/52449302-083b4680-2af4-11e9-871a-4264a2ccc1d9.png">

---

### Switching views
The user could change from list view to card view, or vice versa, using the View option from the Load Window tools menu:

<img width="371" alt="view-options" src="https://user-images.githubusercontent.com/101482/52449345-2e60e680-2af4-11e9-9f90-d67702bfb65b.png">

---

### Card view
The card view looks like this (wide desktop version):

<img width="1196" alt="desktop-wide-card-view" src="https://user-images.githubusercontent.com/101482/52449351-39b41200-2af4-11e9-8804-a06d3bca86fe.png">

---

As with list view, selecting a thumbnail or resource title would load that resource into a new window in the workspace.

marlo-longley:
[...]

PR #1852, opened 2019-02-11
Move adding manifests to the workspace to a new component

cbeer:
This PR moves adding manifests to a separately rendered component to set us up for #1833


src/components/WorkspaceAdd.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import ns from '../config/css-ns';
+import ManifestForm from '../containers/ManifestForm';
+import ManifestListItem from '../containers/ManifestListItem';
+
+/**
+ * Represents a work area that contains any number of windows
```
camillevilla:
```suggestion
 * An area for managing manifests and adding them to workspace
```
I think this was copied over from the `Workspace` component.

cbeer:
I think there should be a way to persist the info responses (I mean, I think they /are/ there already), but would like to handle that in a separate PR that refactors `WindowViewer`.

I did go back and forth about whether the "workspace add" area was a modal on top of the workspace (in which case, the whole thing is moot) or replaced it entirely (as I ultimately did). It should be easy enough to go either way with it, although there are some styling wrinkles to the modal usage that I wanted to avoid for now, in order to unblock follow-on work to this new area.

Issue #1854, opened 2019-02-12
Handle HTML from IIIF metadata


jkeck:

charbugs:
Can you say how html should be handled? There is a wide range from ignoring all to rendering all. Also, can you provide a fixture for that.

jkeck:
This ticket is intended to figure those details out.  I don't have all those answers, and I think it's up to this group to figure out what that means (and how Mirador handled it in the past would be a good start).

I noticed when implementing #1855 that raw HTML was being rendered in the panel, which I assume is not the desired behavior.

charbugs:
Some time ago I worked on an react application that dealt with this problem. There where text messages that came from third party APIs and this texts should be rendered in the UI. It should be possible to allow some elements to render but for safety reasons not the entire HTML set. I used a [HTML santizer](https://www.npmjs.com/package/sanitize-html) for this problem. You can specify a limited set of elements (and even attributes) that you consider harmless for your application. Each other elements will be striped from the text. 

By default, react will not render HTML that you pass as text content to components. But there is this `dangerouslySetInnerHTML` attribute of react that you can use to circumvent that behavior.

### Questions

1. What does the IIIF docs say about HTML content in manifest fields?
2. How is HTML content handled in Mirador 2?
3. Do we want to render HTML that comes with manifests?
4. If so, what set of elements should be allowed, which not?
5. How does this interfere with our design?
6. Do @jvine and @ggeisler already have thoughts on that?

mejackreed:
2. Mirador 2 implementation of sanitization: https://github.com/ProjectMirador/mirador/blob/v2.7.0/js/src/utils/utils.js#L272-L280

ggeisler:
1. The Presentation 2 spec has a section about HTML markup that says "Minimal HTML markup may be included in the description, attribution and metadata properties. It must not be used in label or other properties." (https://iiif.io/api/presentation/2.1/#html-markup-in-property-values) So I think we should display HTML that is found in those properties. 

4. That section of the API (the same section in the Presentation 3 spec looks very similar) also says:

> Clients should allow only a, b, br, i, img, p, and span tags. Clients may choose to remove any and all tags, therefore it should not be assumed that the formatting will always be rendered.

6. Based on the above, maybe the design-oriented question here is whether we are worried about whether the allowable tags might affect display. I think it might depend on the property. 

[...]

PR #1855, opened 2019-02-12
Add canvas (and manifest) level metadata

### LINKED ISSUE #1780: Display canvas level data in the About this item panel
Split out from #1583

canvas-level-md

_Also, see #93 for an older ticket that was also requesting this feature in previous versions_

### Resources
- [Interactive designs](https://github.com/ProjectMirador/mirador-design/wiki/Mirador-3-Interactive-Designs) > Sidebar panel content & navigation

jkeck:
Closes #1780 

Plus, bonus refactors to simplify WindowSideBar* components.
Plus, bonus manifest level metadata (because it was trivial to implement w/i this PR)

canvas-level-md
_Note: the numbers from the window on the left are coming from the IIIF manifest, however that is also the intended fallback behavior for canvas' would labels_

Spun out #1854 to handle HTML entities in metadata

### DIFF
--- a/__tests__/src/components/LabelValueMetadata.test.js
+++ b/__tests__/src/components/LabelValueMetadata.test.js
@@ -1,0 +1,52 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import LabelValueMetadata from '../../../src/components/LabelValueMetadata';
+
+describe('LabelValueMetadata', () => {
+  let wrapper;
+  let labelValuePair;
+
+  describe('when the labelValuePair has content', () => {
+    beforeEach(() => {
+      labelValuePair = [
+        {
+          label: 'Label 1',
+          value: 'Value 1',
+        },
+        {
+          label: 'Label 2',
+          value: 'Value 2',
+        },
+      ];
+      wrapper = shallow(
+        <LabelValueMetadata labelValuePairs={labelValuePair} />,
+      );
+    });
+
+    it('renders a dt/dd for each label/value pair', () => {
+      expect(wrapper.find('dl').length).toEqual(1);
+
+      expect(wrapper.find('dt').length).toEqual(2);
+      expect(wrapper.find('dt').first().text()).toEqual('Label 1');
+      expect(wrapper.find('dt').last().text()).toEqual('Label 2');
+
+      expect(wrapper.find('dd').length).toEqual(2);
+      expect(wrapper.find('dd').first().text()).toEqual('Value 1');
+      expect(wrapper.find('dd').last().text()).toEqual('Value 2');
+    });
+  });
+
+  describe('when the labelValuePair has no content', () => {
+    beforeEach(() => {
+      labelValuePair = [];
+      wrapper = shallow(
+        <LabelValueMetadata labelValuePairs={labelValuePair} />,
+      );
+    });
+
+    it('renders an empty fragment instead of an empty dl', () => {
+      expect(wrapper.find('dl').length).toEqual(0);
+      expect(wrapper.matchesElement(<></>)).toBe(true);
+    });
+  });
+});
--- a/src/components/LabelValueMetadata.js
+++ b/src/components/LabelValueMetadata.js
@@ -1,0 +1,40 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+
+/**
+ * Renders label/value pair metadata in a dl
+ * @prop {object} labelValuePair
+ */
+class LabelValueMetadata extends Component {
+  /**
+   * render
+   * @return {String} - HTML markup for the component
+   */
+  render() {
+    const { labelValuePairs } = this.props;
+
+    if (labelValuePairs.length === 0) {
+      return (<></>);
+    }
+
+    /* eslint-disable react/no-array-index-key */
+    // Disabling array index key for dt/dd elements as
+    // they are intended to display metadata that will not
+    // need to be re-rendered internally in any meaningful way
+    return (
+      <dl>
+        {labelValuePairs.reduce((acc, labelValuePair, i) => acc.concat([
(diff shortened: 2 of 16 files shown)


src/components/LabelValueMetadata.js
```diff
@@ -0,0 +1,40 @@
...
+    const { labelValuePairs } = this.props;
+
+    if (labelValuePairs.length === 0) {
+      return (<></>);
+    }
+
+    return (
+      <dl>
+        {labelValuePairs.reduce((acc, labelValuePair, i) => acc.concat([
+          <dt key={`label-${JSON.stringify(labelValuePair.label[0].value)}`}>
```
jkeck: Could use ideas on better ways to generate this key.  The label can be repeated, and that will cause issues here (at least react seems to throw an error in the console, but things still continue to work).

src/components/WindowSideBarInfoPanel.js
```diff
@@ -1,65 +1,83 @@
...
+      t,
+    } = this.props;
     return (
       <div className={ns('window-sidebar-info-panel')}>
         <Typography variant="h2" className={classes.windowSideBarH2}>{t('aboutThisItem')}</Typography>
-        <Typography variant="h3" className={classes.windowSideBarH3}>{this.manifestLabel()}</Typography>
-        <Typography variant="body2">{this.manifestDescription()}</Typography>
+        {canvasLabel
+          ? <Typography variant="h3" className={classes.windowSideBarH3}>{canvasLabel}</Typography>
+          : null
```
jkeck: Happy to entertain better ways of doing this.  It feels sub-optimal, but I think we should avoid empty headings hanging out if possible.

src/components/LabelValueMetadata.js
charbugs: React docs: "When you don’t have stable IDs for rendered items, you may use the item index as a key as a last resort." So in this case `i` could do it.

src/components/LabelValueMetadata.js
```diff
@@ -0,0 +1,40 @@
...
+            {labelValuePair.label[0].value}
+          </dt>,
+          <dd key={`value-${JSON.stringify(labelValuePair.value[0].value)}`}>
+            {labelValuePair.value[0].value}
+          </dd>,
```
charbugs: I think the component should get label-value-pairs that are already resolved like `[{ label: "foo", value: 1 }, { label: "bar", value: 2 }]`. In the above code the component depends on the peculiarities of the iiif manifest and has to do data transformation. To change this in this way would probably imply to write more specific selectors and pass them to <WindowSideBarInfoPanel/>.

src/components/WindowMiddleContent.js
```diff
@@ -35,12 +35,7 @@ class WindowMiddleContent extends Component {
     const { manifest, window } = this.props;
     return (
       <div className={ns('window-middle-content')}>
-        <WindowSideBar
-          windowId={window.id}
-          manifest={manifest}
-          sideBarOpen={window.sideBarOpen}
-          sideBarPanel={window.sideBarPanel}
-        />
+        <WindowSideBar windowId={window.id} />
```
charbugs: That's how we have to do it. Basically, all subcomponents of <Window/> should only get the `windowId` passed from its parents. The id then can be used to get further things from the store. :+1:

__tests__/src/components/LabelValueMetadata.test.js
```diff
@@ -0,0 +1,46 @@
...
+      );
+    });
+
+    it('renders a dt/dd for each label/value pair', () => {
+      expect(wrapper.find('dl').length).toEqual(1);
+      expect(wrapper.find('dt').length).toEqual(2);
+      expect(wrapper.find('dd').length).toEqual(2);
+    });
+  });
+
```
charbugs: We should also test, for the correct text contents of \<dt/\> and \<dd/\>.

src/components/WindowSideBarInfoPanel.js
charbugs: AFAIK you can return all kinds of falsy values if you don't want to render anything. So undefined would be valid too. This means you can also do:

```javascript
{ canvasLabel && 
  <Typography variant="h3" className={classes.windowSideBarH3}>{canvasLabel}</Typography>
}
```

Saves one line! :sunglasses:

src/components/LabelValueMetadata.js
jkeck: Makes sense.  I'll look into pushing this logic elsewhere.

src/components/LabelValueMetadata.js
jkeck: Ah, okay.  Our linter is failing hard when using the index key, so I may just need to disable it for these lines.  @mejackreed is the preferred method an inline disable here?

cbeer: 👍 to disabling here; it seems very, very unlikely the data will change inside of here without having to re-rending the whole thing anyway.

src/components/LabelValueMetadata.js
cbeer: Am I right that part of the peculiarity of the data structure we're getting from manifesto is something to do with multilingual data (and that flattening it may lose some of the connections expressed in IIIF)?

jkeck: @cbeer correct.  My plan will be to abstract this logic somewhere (most likely a selector) that can have knowledge about which locale is set (once we have that functionality).  Then in the container, where we're mapping state to props, we can use the locale from state when fetching the metadata, and pass the correct thing to the presentation component. (at least that's where my thinking is going so far)

src/components/LabelValueMetadata.js
mejackreed: 👍  sure, inline disable if we need to

jkeck:
Addressed all the feedback.  Thanks all! 🎉

PR #1859, opened 2019-02-12
Enable prefer-stateless-functions

### LINKED ISSUE #1858: Prefer Stateless Functions
in reviewing the docs for 16.8 and the new hooks API, the current direction for React seems to favor stateless functions over classes.  

This rule can be enabled in `eslintrc` and violating classes can be converted.

christopher-johnson:
converts violating classes to functions
add jsdoc headers
closes #1858

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -21,6 +21,6 @@
         "FunctionExpression": true
       }
     }],
-    "react/prefer-stateless-function": "off"
+    "react/prefer-stateless-function": [2, { "ignorePureComponents": false }]
   }
 }
--- a/src/components/CompanionWindow.js
+++ b/src/components/CompanionWindow.js
@@ -1,19 +1,16 @@
-import React, { Component } from 'react';
+import React from 'react';
 import ns from '../config/css-ns';
 
 /**
  * CompanionWindow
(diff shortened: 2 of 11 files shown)


cbeer:
I thought there was an intentional decision early on to stick with classes to avoid unnecessary code thrashing while we discover what things are stateless/pure/etc. It seems like an obvious improvement for some types of components, but some of these don't feel as straightforward.

It also seems like a relatively easy thing to tackle once we have a full suite of components and feel confident what pieces are truly pure.

christopher-johnson:
FYI: the intention here is to _incrementally_ initiate testing of override / plugin component injection hooks that use the new hooks API (only works with functional components).  `miradorWithPlugins`  passes a parent `ref`, which is not optimal for this, as it assumes that all components are class components.  

this would be easier if `miradorWithPlugins` were only implemented on the tested parent components, and not on everything by default.   

I see now that a prerequisite to this PR is to rollback the scope of `miradorWithPlugins` and develop a new component plugin injector that works with functional components.  I work on this currently.

mejackreed:
Closing this PR to clean up our open PR's. I think we decided against using functions at this moment.

PR #1860, opened 2019-02-12
1584 add canvas navigation information panel

glefi:


charbugs:
This PR will conflict with https://github.com/ProjectMirador/mirador/pull/1855. We decided to wait for https://github.com/ProjectMirador/mirador/pull/1855 to get merged then resolve conflicts and repush.

src/state/selectors/index.js
```diff
@@ -34,6 +34,18 @@ export function getManifestCanvases(manifest) {
...
 
+/**
+* Return ids and labels of canvases
+* @ param {Array} canvases
+* @return {Array} - [ {id: 'id', label: 'label' }, ... ]
+*/
+export function getIdAndLabelOfCanvases(canvases) {
+  return canvases.map(canvas => ({
+    id: canvas.__jsonld['@id'], // eslint-disable-line no-underscore-dangle
+    label: canvas.__jsonld.label, // eslint-disable-line no-underscore-dangle
```
jkeck: I'm not sure we want to access the jsonld directly.  We can access the canvas' id simply w/ `canvas.id`.

The label will be a different object based on IIIF version ([3.0](https://iiif.io/api/presentation/3.0/#label) will return a JSON object) so this will cause issues w/ a V3 Manifest.  You'll want to use `getLabel()` instead (or you can simply modify/use the [`getCanvasLabel` selector](https://github.com/ProjectMirador/mirador/blob/041e557993ffb688e7b6eda6fd08ccf612f53f6f/src/state/selectors/index.js#L101-L106)).

charbugs: OK, we will review this issue tomorow.

glefi: @jkeck you're right. I've changed the implementation of the `getIdAndLabelOfCanvases` selector

Issue #1862, opened 2019-02-12
Reducers index should export named reducers.  


christopher-johnson:
depends on "Allow Named Exports"  https://github.com/ProjectMirador/mirador/issues/1707
1. refactor `createRootReducer` into `rootReducer.js`.
2. remove default export from reducers

index.js will become this. 
```javascript
export * from './workspace';
export * from './windows';
export * from './manifests';
export * from './infoResponses';
export * from './config';
```
Why?
If Mirador is imported as a React Component, the reducers individually need to be available for the implementation to construct its root reducer from the export package.

PR #1864, opened 2019-02-12
Disable eslint prefer-default-export

### LINKED ISSUE #1707: Allow named exports
recommend changing the eslint rule to `"import/prefer-default-export": "off"`.  

The current convention of renaming the default export in the import statement (e.g.  `import ConnectedApp from '../components/App';`) smells.  named imports can and should be used wherever possible as this facilitates composition imports and potentially reduces bundle size for consumers.

christopher-johnson:
closes #1707

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -10,6 +10,7 @@
   "parser": "babel-eslint",
   "plugins": ["jest"],
   "rules": {
+    "import/prefer-default-export": "off",
     "no-console": "off",
     "react/jsx-filename-extension": [1, { "extensions": [".js", ".jsx"] }],
     "require-jsdoc": ["error", {


cbeer:
Didn't @charbugs's work to separate containers resolve this?

christopher-johnson:
no.

mejackreed:
> The current convention of renaming the default export in the import statement (e.g. `import ConnectedApp from '../components/App';`) smells.

This seems resolved by @charbugs , no?

> named imports can and should be used wherever possible as this facilitates composition imports and potentially reduces bundle size for consumers.

Is this still true? Are there cases where when we have a single export we don't want it as the default export? I'm all for modifying our style guide approach, I would just like better understanding of why we are doing something.

cbeer:
I guess I'm not seeing us rename the default export any more (`ConnectedXYZ`), hence my confusion.

christopher-johnson:
sorry, I did not ref. https://github.com/ProjectMirador/mirador/issues/1862.  the reducers need to be named exports so that an implementation can create its rootReducer with them.

charbugs:
It is not quite the same issue as \<ConnectedBlaBla/\>. Eslint complains if you only have a single named export. But there are cases where you want to do this.

PR #1879, opened 2019-02-13
Adds named export reducers

### LINKED ISSUE #1862: Reducers index should export named reducers.

christopher-johnson:
refactors createRootReducer from index
closes #1862

### DIFF
--- a/src/state/reducers/index.js
+++ b/src/state/reducers/index.js
@@ -1,22 +1,5 @@
-import { combineReducers } from 'redux';
-import workspaceReducer from './workspace';
-import windowsReducer from './windows';
-import manifestsReducer from './manifests';
-import infoResponsesReducer from './infoResponses';
-import configReducer from './config';
-
-/**
- * Function to create root reducer
- * from plugin reducers.
- * @namespace CreateRootReducer
- */
-export default function createRootReducer(pluginReducers) {
-  return combineReducers({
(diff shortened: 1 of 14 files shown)


src/state/reducers/index.js
```diff
@@ -1,22 +1,5 @@
-import { combineReducers } from 'redux';
```
mejackreed: One question I have about this, could someone not just import these reducers directly if they are provided from the package exports without this refactor?

christopher-johnson: not clear on exactly what you mean.  they are not exported with names currently so they are not available to the consuming implementation.  

the benefit of named imports is [here](https://github.com/ProjectMirador/mirador/blob/ed33b9c61b96f38636ff44f3dc3e5e2236a9cd3f/src/state/reducers/rootReducer.js#L2-L4).  clean abbreviated syntax for imports.

Issue #1880, opened 2019-02-13
Clicking on a the infoPanel before the manifest has loaded throws an error


jkeck:
I believe this is because the selector I wrote (`getSelectedCanvas`) returns an empty object when the manifestation is not yet available.  We then try to load the label from manifesto (using `getLabel()`) and since the POJO doesn't have a `getLabel()` function, we get an error and a WSOD.

charbugs:
Can you give a hint how to reproduce this error?

Not sure if it's the same problem, but when I declare a wrong URL in `__tests__/integration/mirador/index.html` I get an error and a WSOD too:

![mirador-bug-wrong-url](https://user-images.githubusercontent.com/11216290/52788622-d54c0200-3061-11e9-9c82-97866c9bdc97.png)

jkeck:
This issue was triggered by a manifest loading slowly due to connection slowness (where some of the window components had rendered before the manifest had loaded).  You may need to use a slow loading manifest (not sure if we have a good example of that in the wild) or simulate the slowness in another way.

charbugs:
It's possible to test that issue by typing

`miradorInstance.store.dispatch(miradorInstance.actions.requestManifest('foobar'))`
and
`miradorInstance.store.dispatch(miradorInstance.actions.addWindow({ manifestId: 'foobar' }))`

in the dev tool console. Thus we will get an empty window that refers to an unloaded manifest.

PR #1889, opened 2019-02-14
Bugfix: clicking info panel before manifest loaded throws error. Clos…

charbugs:
…es #1880


charbugs:
CI test is pending for over an hour now. Can't figure out why.

jkeck:
It looks like it's the connection between GH and Travis.  The build seems to have passed, it's just not being reported correctly here.

PR #1893, opened 2019-02-14
Add an initial (basic) functionality to open left side-panel content...

### LINKED ISSUE #1752: Window sidebar --> Companion window --> implement "popping out" to  right
This ticket implements toggling the Window sidebar content from the left side of the window to a the right companion window. 

It would be ideal if #1583 is implemented first, but it's not necessarily a blocker. 

sidebar open, companion window closed
sidebar closed, companion window open

### Resources
[Interactive design](https://xd.adobe.com/view/1a628598-6843-4d25-67cb-d9e42eff6617-6c53/screen/4442ab28-e7a0-4ad4-962d-74506597c24d/Companion-right-default)
- Possible architectural consideration: note that an info panel, right companion window, and bottom companion window can be open simultaneously.

### Acceptance criteria
- [ ] The companion window is available for any Window sidebar menu item

Clicking on the companion window button
- [ ] Closes the window sidebar info panel
- [ ] Keeps the window sidebar menu open (i.e. icons showing, but no icon is selected)
- [ ] Opens the right companion window with content
- [ ] ~The image viewing area remains the same size (e.g. image is not clipped); total window area increases to accommodate the companion window~ <-- Technical considerations for this will be split out to another ticket 🎟

The open companion window
- [ ] Has a header area 
- [ ] Has a working close button

### Out of scope
Drag button and behavior / moving content to bottom companion window. Splitting that out to #1754

jkeck:
...into a right companion window.

Closes #1752 

companionwindowbehavior


__tests__/integration/mirador/companion_windows.test.js
```diff
@@ -0,0 +1,27 @@
...
+    await expect(page).toClick('button[aria-label="Open in companion window"]');
+
+    await expect(page).toMatchElement('.mirador-companion-window-right .mirador-window-sidebar-info-panel');
+
+    await expect(page).toClick('button[aria-label="Close this companion window"]');
+
+    // TODO: Write something like the assertion below that verifies the sidebar is no longer visible
+    // await expect(page).not.toMatchElement(
+    //   '.mirador-companion-window-right .mirador-window-sidebar-info-panel',
+    // );
```
jkeck: I cannot get this test to pass for the life of me.  I can set `waitFor`s and run this in headless mode and see that the element appears to be clicked (from the MUI button animations) but nothing happens.  I can then click the button myself and get the test to pass.

src/components/CompanionWindow.js
```diff
@@ -1,19 +1,91 @@
...
   render() {
+    const {
+      classes, closeCompanionWindow, isDisplayed, position, t, windowId,
+    } = this.props;
     return (
-      <div className={ns('window-companion-side')} style={{ display: 'none' }} />
+      <Paper
+        className={[classes.root, ns(`companion-window-${position}`)].join(' ')}
+        style={{ display: isDisplayed ? null : 'none' }}
+        square={true} // eslint-disable-line react/jsx-boolean-value
```
jkeck: MUI has `false` by default so we need to set this to `true` (but eslint complains)

src/components/CompanionWindow.js
```diff
@@ -1,19 +1,91 @@
...
+      >
+        {this.activePanelComponent()}
+        <IconButton
+          aria-label={t('closeCompanionWindow')}
+          className={classes.closeButton}
+          onClick={() => { closeCompanionWindow(windowId, null, position); }}
```
jkeck: I've gone back and forth on the approach for this.

This dispatch prop is really just `setCompanionWindow` under the hood, and we're then passing `null` to effectively "unset` the companion window.

I could also add an explicit `closeCompanionWindow` action then calls `setCompanionWindow` passing the `null` value in itself.  Happy to go either direction.

src/components/CompanionWindow.js
```diff
@@ -1,19 +1,91 @@
...
+ */
+const styles = theme => ({
+  closeButton: {
+    position: 'absolute',
+    right: 0,
+    top: 0,
+  },
+  root: {
+    ...theme.mixins.gutters(),
+    width: '200px',
```
jkeck: I just picked something at random here (and probably is not what we want).  w/o this, the companion window takes up the entire viewer space.

src/components/CompanionWindow.js
mejackreed: Can you just set it like this?

```javascript
...
        style={{ display: isDisplayed ? null : 'none' }}
        square
...
```

src/state/actions/window.js
```diff
@@ -63,6 +82,23 @@ export function toggleWindowSideBarPanel(windowId, panelType) {
...
+ * @param  {String} panelType The type of panel content to be rendered
+ *                            in the companion window (e.g. info, canvas_navigation)
+ * @param  {String} position The position of the companion window to
+ *                           set content for (e.g. right, bottom)
+ * @memberof ActionCreators
+ */
+export function popOutCompanionWindow(windowId, panelType, position) {
+  return ((dispatch) => {
+    dispatch(setWindowCompanionWindow(windowId, panelType, position));
+    dispatch(toggleWindowSideBarPanel(windowId, 'closed'));
```
jkeck: Not sure if this is the best way to go about this, but, it seems to work. ¯\\_(ツ)_/¯

__tests__/src/reducers/windows.test.js
```diff
@@ -121,6 +121,42 @@ describe('windows reducer', () => {
...
+        type: ActionTypes.SET_WINDOW_COMPANION_WINDOW,
+        windowId: 'abc123',
+        position: 'right',
+        panelType: 'info',
+      };
+      const before = {
+        abc123: {},
+      };
+      const after = {
+        abc123: { companionWindows: { right: 'info' } },
```
jkeck: Note the shape of the state here.  I'm not sold on this working long-term, but it's good enough for the acceptance criteria as outlined and would allow us to move the thumbnail content there as well.  (It would not allow for multiple companion windows to be open in the same right/bottom position though).

src/components/WindowSideBarCanvasPanel.js
```diff
@@ -12,24 +13,36 @@ class WindowSideBarCanvasPanel extends Component {
...
+  t: PropTypes.func.isRequired,
 };
 
-export default WindowSideBarCanvasPanel;
+/**
+ * @private
+ */
+const styles = theme => ({
+  windowSideBarH2: theme.typography.h5,
+});
```
jkeck: Is there a good way for us to share these styles, or set them somewhere?

src/components/CompanionWindow.js
```diff
@@ -1,19 +1,91 @@
...
+   * @return React Component
+   */
+  activePanelComponent() {
+    const { windowId, panelContent } = this.props;
+    switch (panelContent) {
+      case 'info':
+        return <WindowSideBarInfoPanel windowId={windowId} />;
+      case 'canvas_navigation':
+        return <WindowSideBarCanvasPanel windowId={windowId} />;
+      default:
```
jkeck: This logic is now duplicated in two places.  Is there a good place to hold shared logic like this?

src/state/actions/window.js
aeschylus: Another way to do it is to make this state change in the reducers. I can understand the urge to avoid repeating the state merging logic, so to DRY it up, I believe a utility function can be written and re-used in each of the appropriate reducers. The utility to merge the companion window into state and the utility to toggle the sidebar panel state would both be called in the reducer that receives popOutCompanionWindow. That way the action log would be clean (only the original intended action is fired), and the side effects would be removed.

src/components/CompanionWindow.js
charbugs: I tink [that's](https://github.com/ProjectMirador/mirador/blob/master/src/components/WindowSideBarPanel.js#L15) the place jkeck refers to.

charbugs:
I think the approach of this PR won't scale. I know it is meant as a basic setup for the companion window implementation. But I think large parts of the code has to be rewritten when we go further with the functionality. There is lot to say here because it's a large PR but I will try to limit my thoughts to the essential.

According to the [design mockups](https://xd.adobe.com/view/1a628598-6843-4d25-67cb-d9e42eff6617-6c53/) there will be multiple companion windows on different positions. Also there can be multiple on the right, on the bottom and so on. This means in \<WindowMiddleContent/\> we have to render the companion windows out of an array or a similar data structure. That probably will look something like this:

```javascript
// in <WindowMiddleContent/>
{ window.compananionWindows.map(cw => <CompanionWindow /* ... */ />) }
```

If fear that the state structure for windows will then look similar to this:

```javascript
windows: {
  'window-xyz': {
    /* ... */
    companionWindows: [
      { position: 'right', content: 'info' },
      { position: 'right', content: 'canvas' },
    ]
  }
}
```

If we go like this the state structure isn't anymore flat or "normalized". This leads to the problem that we have different shapes for different slices of the state which in turn means that we will have different patterns per reducer/action/selector and that we can not apply common utility functions.

So I would suggest to keep the companion windows in a different state slice as we currently do with windows and manifests. It may look like this:

```javascript
state: {
  windows: {
    'window-xyz': {
      /* .. */
      compantionWindowIds: [ 'cw-1', 'cw-2' ]
    }
  },
  companionWindows: {
    'cw-1': {
      windowId: 'window-xyz',
      position: 'right', 
      content: 'info' 
    },
    'cw-2': {
      windowId: 'window-xyz',
      position: 'right', 
      content: 'canvas' 
    },
  }
}
```

This means we stick with the idea that state is made of tables (windows, manifests, companionWindows, ...), where each table is made of rows (cw-1, cw-2, ...), where each row is made of primitive values (windowId, position, content, ...).* This is predictable and can be utilized in many places.

But: if we decide to go this way, then a lot of code in this PR has to be rewritten.

There's another pattern in the code, of which I'm not sure if I like it. It is somehow related to the above, so I will mention it here:

```javascript
// in <WindowMiddleContent/>
<CompanionWindow windowId={window.id} position="right" />

// in <CompanionWindow/>
<Paper /* ... */ style={{ display: isDisplayed ? null : 'none' }} >
```

That is, the \<CompantionWindow/\> will be rendered by \<WindowMiddleContent/\> but whether it finally appears is up to the companion window itself. On the one hand it looks clever to me and the code in the middle content component is clean. On the other hand I ask myself why the companion window (and in the future multiple companion windows) should be rendered if it doesn't appear, why does it even exist?

When we follow the suggestion from above then a companion window will be created in store if the user clicks the pop out button and will be destroyed when the user hits the close button. In this case we can simply go like this:

```javascript
// in <WindowMiddleContent/>
{ window.compananionWindowIds.map(id => <CompanionWindow id={id} />) }
```

That means: 1) only existing companion windows will be rendered, 2) there is no need in the companion window to decide if it should appear, 3) there is only one prop passed from parent to child (and that's :icecream:)



*I know that arrray isn't a primitive value but we can not really avoid it. But the arrays in state should only contain primitive values.

PR #1894, opened 2019-02-14
Fixes #1869 better layout when in book view

### LINKED ISSUE #1869: Determine best type of layout for book view images with differing aspect ratios
Pinging @ggeisler and @jvine on this.

#1867 implements book view. However, for manifests that have canvases that differ wildly in aspect ratio, the view can be a little awkward.

https://deploy-preview-1867--mirador-dev.netlify.com/__tests__/integration/mirador/

Manifest to test: https://media.nga.gov/public/manifests/nga_highlights.json

What is the designed behavior? Everything works ok with more "book like" material

Is this good enough?

bookviewawkward

mejackreed:
Based on feedback in #1869 this PR implements the Mirador 2 behavior

right heights

### DIFF
--- a/__tests__/src/components/OpenSeadragonViewer.test.js
+++ b/__tests__/src/components/OpenSeadragonViewer.test.js
@@ -64,7 +64,7 @@ describe('OpenSeadragonViewer', () => {
   describe('boundsFromTileSources', () => {
     it('generates bounds from a set of tileSources', () => {
       expect(wrapper.instance().boundsFromTileSources()).toEqual(expect.arrayContaining(
-        [0, 0, 250, 201],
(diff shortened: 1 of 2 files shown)


jkeck:
Can you say more about _why_ this change creates a better layout when in book view in the commit message?

mejackreed:
@jkeck 
Good call on that. I've updated the commit message to add additional info:

```
Fixes #1869 better layout when in book view

This new layout matches Mirador 2's implementation. When in bookview,
extend both images to the same height, centered, and then scale each
of their widths out from the center. This is negligible when images have
similar heights (like a book), but is more noticiable with different
sized images.
```

aeschylus:
Nice!

Issue #1834, opened 2019-02-08
Implement 'List view' of load window plugin


ggeisler:
This is part of #1833.

See the [interactive mockup](https://xd.adobe.com/view/477b3833-0ae4-47fa-6032-15d9b7bcd9f0-314a/) for a better sense of how the plugin is intended to work than I can give with screenshots.

A goal in the design of list view for the load window plugin is to keep the arrangement of resource elements consistent across device sizes, and to use that same arrangement in the sidebar of a collection or multivolume work where we show the manifests that are part of the larger work. For example, here is the proposed arrangement for the window sidebar and the load window on various device sizes:

### Window sidebar

<img width="325" alt="window-sidebar-a" src="https://user-images.githubusercontent.com/101482/52449545-f9a15f00-2af4-11e9-8f94-fc5ede5408a1.png">

---
<img width="345" alt="window-sidebar-b" src="https://user-images.githubusercontent.com/101482/52449558-01610380-2af5-11e9-9058-28f420993377.png">

---

### Mobile
<img width="371" alt="mobile-list-view" src="https://user-images.githubusercontent.com/101482/52449613-34a39280-2af5-11e9-9da7-f9ad9b141085.png">

---

### Tablet
At tablet and larger, the device is wide enough to fit all resource elements in the same row:

<img width="766" alt="tablet-list-view" src="https://user-images.githubusercontent.com/101482/52449623-3d946400-2af5-11e9-8b93-c11420ca9eea.png">

---

### Desktop
<img width="1021" alt="desktop-list-view" src="https://user-images.githubusercontent.com/101482/52449629-438a4500-2af5-11e9-9938-b22e7ee2d3c6.png">

---

## Linked elements
We don't want to make the entire row linked (because that makes it hard to scroll without selecting a resource on touch devices) but we do want a good-sized link target to make it easy and quick to open a resource. The thumbnail and title are visually grouped and should be a good link target. For accessibility purposes, we should probably group those two elements in HTML and make the group the link (so we don't have repeated links in screenreaders).

General suggestion: The thumbnail and title section takes 100% width on mobile, and 50% width on larger than mobile viewports (leaving the institutional elements the other 50%).

## Add resource panel (Split to #1857)
[...]

PR #1902, opened 2019-02-15
Add support for configuration initial manifests for the viewer

cbeer:
I guess this is part of #1833 so the initial load window view can be populated with data. It has an indirect dependency on #1901 in order to make the data actually show up in the interface, but is independent and can be merged with or without it.

There's some obvious follow-on work to make sure we lazy-load manifests instead of immediately dispatching requests for data we might not even use, but this hopefully is sufficient for now.

### DIFF
--- a/__tests__/integration/mirador/index.html
+++ b/__tests__/integration/mirador/index.html
@@ -19,7 +19,23 @@
        {
          loadedManifest: 'https://iiif.bodleian.ox.ac.uk/iiif/manifest/e32a277e-91e2-4a6d-8ba6-cc4bad230410.json',
          thumbnailNavigationPosition: 'off',
-       }]
+       }],
+       manifests: {
+         "http://media.nga.gov/public/manifests/nga_highlights.json": { provider: "National Gallery of Art"},
+         "https://data.ucd.ie/api/img/manifests/ucdlib:33064": { provider: "Irish Architectural Archive"},
+         "https://wellcomelibrary.org/iiif/b18035723/manifest": { provider: "Wellcome Library"},
+         "http://dams.llgc.org.uk/iiif/2.0/4389767/manifest.json": { provider: "The National Library of Wales"},
+         "https://demos.biblissima.fr/iiif/metadata/florus-dispersus/manifest.json": { provider: "Biblissima"},
+         "http://beta.biblissima.fr/iiif/manifest/ark:/43093/desc57cb76cd3739a24a9277b6669d95b5f3a590e771": { provider: "Biblissima"},
+         "https://www.e-codices.unifr.ch/metadata/iiif/gau-Fragment/manifest.json": { provider: "e-codices - Virtual Manuscript Library of Switzerland"},
+         "https://wellcomelibrary.org/iiif/collection/b18031511": { provider: "Wellcome Library"},
+         "https://gallica.bnf.fr/iiif/ark:/12148/btv1b10022508f/manifest.json": { provider: "Bibliothèque nationale de France"},
+         "https://manifests.britishart.yale.edu/Osbornfa1": { provider: "Beinecke Rare Book and Manuscript Library, Yale University"},
+         "https://iiif.biblissima.fr/chateauroux/B360446201_MS0005/manifest.json": { provider: "Biblissima"},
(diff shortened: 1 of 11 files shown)


src/lib/MiradorViewer.js
```diff
@@ -67,6 +67,12 @@ class MiradorViewer {
         thumbnailNavigationPosition,
       }));
     });
+
+    Object.keys(mergedConfig.manifests || {}).forEach((manifestId) => {
```
mejackreed: While I like the fact that we can support multiple manifests now, I'm hesitant to introduce this same pattern from M2 without a way to mitigate kicking off so many requests.

Could we huddle up to discuss before shipping this forward?

cbeer: That's fine with me, although I'm out next week, but I'm more than happy for someone else to pick this up. I'd just hope we can find a tractable solution quickly because this feature will be very useful for completing the rest of the load window tickets.

src/state/reducers/manifests.js
```diff
@@ -10,6 +10,8 @@ export const manifestsReducer = (state = {}, action) => {
       return {
         ...state,
         [action.manifestId]: {
+          ...state[action.manifestId],
+          ...action.properties,
```
charbugs: If `properties` is an object we got nested state. Is it possible to place the single properties directly in the manifest object?

src/state/selectors/index.js
```diff
@@ -42,6 +42,10 @@ export function getManifestCanvases(manifest) {
     return [];
   }
 
+  if (!manifest.manifestation.getSequences) {
```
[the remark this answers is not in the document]
cbeer: With the set of test fixtures I got from @ggeisler, some of them did not implement `getSequences`.

__tests__/integration/mirador/index.html
```diff
@@ -19,7 +19,23 @@
        {
          loadedManifest: 'https://iiif.bodleian.ox.ac.uk/iiif/manifest/e32a277e-91e2-4a6d-8ba6-cc4bad230410.json',
          thumbnailNavigationPosition: 'off',
-       }]
+       }],
+       manifests: {
+         "http://media.nga.gov/public/manifests/nga_highlights.json": { provider: "National Gallery of Art"},
```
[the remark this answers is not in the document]
cbeer: See the discussion on #1834 (https://github.com/ProjectMirador/mirador/issues/1834#issuecomment-463304125); apparently IIIF v3 may have some appropriate data, but for now (and in Mirador 2), it's provided data.

src/state/reducers/manifests.js
cbeer: I'm not sure I understand the request or concern?

src/state/selectors/index.js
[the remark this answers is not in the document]
charbugs: So, `manifesto.create(someManifestFixture).getSequences === undefined`? Can you copy that fixture here?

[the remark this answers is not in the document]
cbeer: https://wellcomelibrary.org/iiif/collection/b18031511 (a IIIF collection)

src/state/reducers/manifests.js
charbugs: Oh, sorry, I was wrong. The spread operator already does what I mean.

src/lib/MiradorViewer.js
mejackreed: Gotcha. Do you think it is something we could do within the ManifestListItem itself? So that only when that component mounts it does the async request.

cbeer: And add another action + reducer for "hey, here's a manifest i might want eventually"?

src/state/selectors/index.js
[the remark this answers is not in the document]
charbugs: Ok, I can track it. So, the check makes sense. 

But doesn't that mean that manifesto has a unrelieable API? I mean, isn't it the purpose of manifesto to provide an identical interface for all the peculiarities in the manifest?

src/lib/MiradorViewer.js
mejackreed: I think you could use `fetchManifest` with checking if it has already been fetched perhaps in `componentDidMount` or something similar? that way we only get the manifests when we need to view them.

cbeer: Have a look at this; there's probably some more refactoring in the future, but it seems to delay loading at least.

PR #1903, opened 2019-02-15
Add a workspaceControlPanel.enabled config option for displaying...

### LINKED ISSUE #1885: Create a setting to "disable workspace control panel" and implement
For an embedded Mirador scenario, it may be desirable that the Workspace control bar is not available to users. We should make this configurable.

jkeck:
...the WorkspaceControlPanel component.

Closes #1885

You can modify this in the netlify deploy by putting this in your console (and changing the `enabled` value to `true` or `false` based on if you want it to display or not)

```js
miradorInstance.store.dispatch(
  {
    type: "UPDATE_CONFIG",
    config: {
      ...miradorInstance.store.getState()['config'],
      workspaceControlPanel: { enabled: false }
    }
  }
);
```

toggleworkspacesidebar

### DIFF
--- a/src/config/settings.js
+++ b/src/config/settings.js
@@ -19,5 +19,8 @@ export default {
   },
   workspace: {
     type: 'mosaic',
+  },
+  workspaceControlPanel: {
+    enabled: true,
   }
 };
--- a/src/containers/Workspace.js
+++ b/src/containers/Workspace.js
@@ -9,6 +9,7 @@ import Workspace from '../components/Workspace';
  */
 const mapStateToProps = state => (
   {
+    isWorkspaceControlPanelVisible: state.config.workspaceControlPanel.enabled,
     workspaceType: state.config.workspace.type,
     windows: state.windows,
   }
--- a/__tests__/src/components/App.test.js
+++ b/__tests__/src/components/App.test.js
@@ -14,6 +14,7 @@ function createWrapper(props) {
   return shallow(
     <App
       isFullscreenEnabled={false}
+      isWorkspaceControlPanelVisible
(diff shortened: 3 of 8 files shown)


src/config/settings.js
```diff
@@ -19,5 +19,8 @@ export default {
   },
   workspace: {
     type: 'mosaic',
+  },
+  workspaceControlPanel: {
+    enabled: true,
```
jkeck: I could easily be convinced to flatten this out.

src/containers/Workspace.js
```diff
@@ -9,6 +9,7 @@ import Workspace from '../components/Workspace';
  */
 const mapStateToProps = state => (
   {
+    isWorkspaceControlPanelVisible: state.config.workspaceControlPanel.enabled,
```
jkeck: I could theoretically pass this from `App.js` instead of getting it from state (since it's a prop there), but I wasn't sure if that would be coupling to the parent (since in order to move Workspace somewhere you may have to figure out how to get that out of state again).

I'm happy to go either way.

aeschylus: State makes sense to me. I think there's still a lot of open questions about how the config is getting passed around.

src/config/settings.js
aeschylus: Are there other settings we would want to add here? I think it makes sense to have them grouped, as you have it already. There are a lot of open questions about the config situation. State needs to be normalised, but the config needs to be easily documented and human-memorable. Settings should exactly mirror the config (is how I'm thinking of it at the moment). From that point of view, I prefer what's already here.

PR #1920, opened 2019-02-19
Move determining the workspace layout to component lifecycle methods instead of render()

### LINKED ISSUE #1909: Closing all windows results in undefined window id
**What is the current behavior?**
load a resource, close last window, app crashes.
console reports `TypeError: Cannot read property 'id' of undefined`.
  
**What is the expected behavior?**
close last window, control panel and empty workspace remain visible

Introduced with https://github.com/ProjectMirador/mirador/pull/1905

### LINKED ISSUE #1914: WorkspaceMosaic updates state during render
`updateWorkspaceMosaicLayout` is called from `render()`.  This throws a React warning in the console.  This could be fixed by using lifecycles like this (or alternatively as a functional component using hooks):

```javascript

  componentDidMount() {
    this.determineWorkspaceLayout();
  }

  componentDidUpdate(prevProps) {
    const { windows, workspace } = this.props;
    if (prevProps.windows !== windows || prevProps.workspace !== workspace) {
      this.determineWorkspaceLayout();
    }
  }
```

cbeer:
Fixes #1914 
Fixes #1909


src/components/Window.js
```diff
@@ -15,6 +15,8 @@ class Window extends Component {
    */
   render() {
     const { manifest, window } = this.props;
+    if (!window) return <></>;
```
cbeer: Maybe there's a better way to do this, but it seems like there's some asynchronous behavior that's making this hard to guard against. We clearly ran into this before in the mosaic layout by adding a guard clause like:

```
    const window = windows[id];
    if (!window) return null;
```

but, apparently, between that guard clause and the window container, state can change to remove the window.

PR #1922, opened 2019-02-19
Normalize companion window state

### LINKED ISSUE #1899: Normalize companion window state
Raised originally in https://github.com/ProjectMirador/mirador/pull/1893#issuecomment-464086407

The implementation in #1893 will likely need a refactor so that the state can be normalized.

jkeck:
Closes #1899


src/state/reducers/companion_windows.js
```diff
@@ -0,0 +1,65 @@
...
+  ));
+}
+
+/**
+ * companionWindowsReducer
+ */
+export const companionWindowsReducer = (state = {}, action) => {
+  switch (action.type) {
+    // action params: id
+    case ActionTypes.CLOSE_COMPANION_WINDOW:
```
charbugs: I don't understand why the companion window will not be removed entirely on CLOSE_COMPANION_WINDOW. Is there a wanted UI behavior that makes that necesarry?

jkeck: Closing it (as opposed to removing it all-together) allows the user's scroll position to be retained when re-opening.  That's useful for a single companion window, but won't necessarily make sense once we have multiple companion windows in a particular position.  I'll change this to removing the companion window from state outright so that it doesn't cause any confusion down the road.

charbugs: Ok, I see :+1:

jkeck: Just pushed up the change to replace `CLOSE_COMPANION_WINDOW` with `REMOVE_COMPANION_WINDOW` behavior.  Thanks for pointing that out.

jkeck:
@charbugs good catch.

charbugs:
I found another bug: When you pop out a companion window serveral times then the existing companion window will be removed (as wanted), but the references in window are not deletes. So there are dead references in window.

I figured out that a source of confusion is the `SET_COMPANION_WINDOW` action. It is actually concerned with two things: adding and updating, leading to some conditional operations that are hard to follow.

I started to refactor this by setting up two different actions: `ADD_COMPANION_WINDOW` and `UPDATE_COMPANION_WINDOW`. I thought it would be a easy change, but I finally touched every file of this PR. :flushed: 

It apparently works, but I'm not sure if I should commit it to this PR or if I should create an additional alternative PR. The point is that it overwrites a lot of code of this PR. Or isn't this a problem?

How to proceed?

jkeck:
This is all the boilerplate code I was trying to avoid having to write in the first place (which would have prevented all these bugs, caveats, and clean-up code).

I'll take a look over to think of how to handle that, but it seems like no matter what, with these kinds of approaches, we'll be writing a ton of boilerplate state management code that is going to be constant source of bugs in this project.

charbugs:
Can you say what kind of approach do you exactly taking about?

mejackreed:
@charbugs Could you submit your work as a PR to this branch `1899-normalize-companion-window-state`?  That way we can see the intentional change and move both of these forward?

Thanks!

PR #1927, opened 2019-02-20
Add an error message when add manifest request fails

### LINKED ISSUE #1873: Load window --> Error notification banner for for Add resource
⬇️ Split from #1834: Overview for load window 'List view'

### Add resource error message
If the resource cannot be retrieved, we should let the user know by replacing the placeholder row with a notification banner. If we follow the Material UI guidelines for the [banner](https://material.io/design/components/banners.html#), it is displayed at the top of the content area, is non-modal, and pushes content down rather than overlaying it:

tablet-loading-error

(EDIT from @ggeisler: @jvine and I discussed and we think we will need to ignore the Material UI guidelines in this case and show the notification banner in place of the placeholder row that failed to load, _wherever that placeholder row is in the list_. This is because the top-pinned banner placement won't be as intuitive if there are multiple manifest loading errors, or if the manifest that failed to load is not at the top of the list (e.g., on initial load of the list, several manifests fail to load).) 

- [x] Dismiss: removes the notification banner
- [x] Try again: removes the notification banner and retries the URL, adding the placeholder row

### Acceptance criteria
There are a lot of possible error states that need to be fleshed out. For now, let's implement this UI element and use a simple error case: a manifest that isn't a valid URL (e.g. doesn't begin with `https://`.

### Out of Scope
#1748 - better placeholder state

jkeck:
Closes #1873 

add-item-error

---

add-item-errors


src/state/actions/manifest.js
```diff
@@ -58,7 +58,13 @@ export function fetchManifest(manifestId, properties) {
     return fetch(manifestId)
       .then(response => response.json())
       .then(json => dispatch(receiveManifest(manifestId, json)))
-      .catch(error => dispatch(receiveManifestFailure(manifestId, error)));
+      .catch((error) => {
+        if (typeof error === 'object') { // Returned by JSON parse failure
+          dispatch(receiveManifestFailure(manifestId, String(error)));
+        } else {
+          dispatch(receiveManifestFailure(manifestId, error));
```
jkeck: This whole part feels sub-optimal, but wasn't sure about all the scenarios we'll ultimately want to handle.  The existence of the `error` in the manifest state is more or less used as a boolean at this point (since we don't display this error to the user).  The pre-existing behavior was to store the error string (and I think we'll need something more than a true/false state down the road) so I'm casting it as a string for now.

Before this, we were sending a (very odd) error object originating from the JSON library that didn't have any keys or values but would print a string to the error log.  React was choking on this and just throwing an error, and that needed to be addressed for any of this to work.

__tests__/integration/mirador/invalid-api-response.test.js
```diff
@@ -14,4 +14,25 @@ describe('Mirador Invalid API Response Handler Test', () => {
       expect(e.name).toMatch('TimeoutError');
```
jkeck: Not really part of my PR, but I puzzled over this for a bit.  I don't think this test will ever fail, because all this does is test that there is not an `li` w/ that text in the page (which there isn't, so, yeah, it's going to pass).

I'm not sure what the intent of the test is based on the "breaks Mirador" description, so I'm not really sure what to do with this.

src/components/ManifestListItemError.js
```diff
@@ -0,0 +1,75 @@
...
+  classes: PropTypes.object.isRequired, // eslint-disable-line react/forbid-prop-types
+  manifestId: PropTypes.string.isRequired,
+  onDismissClick: PropTypes.func.isRequired,
+  onTryAgainClick: PropTypes.func.isRequired,
+  t: PropTypes.func.isRequired,
+};
+
+const styles = {
+  errorIcon: {
+    color: 'red',
```
jkeck: Should I be using this?  A hex?  A variable?

PR #1937, opened 2019-02-21
1872 thumbnail list

glefi:


src/components/ValidationCanvas.js
```diff
@@ -0,0 +1,35 @@
+import ManifestoCanvas from '../lib/ManifestoCanvas';
+
```
mejackreed: I'd be fine w/ moving thing directly into `ManifestoCanvas` at some point if that makes sense.

glefi: I thought, as ManifestoCanvas is kind of external code and may be modified and therefore needs to be replaced somedays it is easier to maintain, when inheriting from `ManifestoCanvas`. What do you think?

mejackreed: `ManifestoCanvas` is an internal class that is just providing some additional convenience methods on Manifesto's Canvas objects. I guess the naming is probably not the best and could use some work. I'm good either way.

PR #1941, opened 2019-02-21
1888 app crashes while loading canvas thumbnails

### LINKED ISSUE #1888: App crashes while loading canvas thumbnails
mirdor-bug-canvas-thumbnails

glefi:
I've added a default width, so default width and height can be used to display the defaultImgPlaceholder

Fixes #1888


glefi:
I'm not sure, if this is related to #1990, but the error seems to be still present in master.

mejackreed:
I think its related but not the same as this fix.

PR #1954, opened 2019-02-23
Add a workspace language switcher

### LINKED ISSUE #1876: i18n locale toggle (workspace)
We now have a basic i18n implementation, we need to implement a UI element for switching the workspace's locale.

jkeck:
Closes #1876 

language-switching

Note that I've kept 881855b as a separate commit in case we want to nix it.  I wasn't sure if that was a good idea or I should try and keep props that get sent it "flatter".  It feels like it simplifies the `LanguageSettings` component a bit (and possibly gets us close to something that can be re-used elsewhere).

### DIFF
--- a/__tests__/integration/mirador/language_switching.test.js
+++ b/__tests__/integration/mirador/language_switching.test.js
@@ -1,0 +1,24 @@
+describe('Language Switching', () => {
+  describe('Application Language', () => {
+    it('allows the user to switch the application language', async () => {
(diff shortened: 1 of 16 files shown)


jkeck:
This PR introduces the `Language` option to the workspace menu.  Does anybody (@glefi?) have a good German translation for that?  (`Sprache`/`Sprachen`)?

jkeck:
Thanks @camillevilla, just updated the PR to include `Sprache` for the language toggle.

Issue #1760, opened 2019-01-31
Adjust window layout: bottom + right companion windows


camillevilla:
Related #1714 / #1715 

This ticket shakes out of a discussion @jkeck and I had with @jvine. I think we could still move ahead with basic implementations of companion windows in #1754 and #1752. This design becomes relevant in cases where we have both a bottom companion window and right companion window open. 

I'm interested in @mejackreed's opinions re: implementation, and am tagging this with @jvine and @ggeisler to confirm this layout change.  

## Currently implemented
![window-layout](https://user-images.githubusercontent.com/5402927/52032808-6bffb700-24d7-11e9-8e48-9efb16cf475b.png)

## Proposed
**Image updated March 4**
![proposed window layout](https://user-images.githubusercontent.com/6945691/53747118-6c34fe80-3e57-11e9-821f-67e234f37696.png)

![screen shot 2019-01-30 at 3 57 52 pm](https://user-images.githubusercontent.com/5402927/52032779-4b376180-24d7-11e9-9da3-8734a396df89.png)

## Resources
- [Interactive designs](https://github.com/ProjectMirador/mirador-design/wiki/Mirador-3-Interactive-Designs) --> Companion window behavior
- #1715: original layout design
- [CSS prototyping codepen](https://codepen.io/mejackreed/pen/MLezVX)

jvine:
Comment above says for both window types:

> can be split **vertically** into multiple companion windows

which isn't accurate. It's hard to be unambiguous about the behaviour in terms of "splitting vertically/horizontally", so let me try it this way:

Companion windows are a specific shape (vertical or horizontal) and can stack within their allocated region.

Primary window = Canvas area, including left side menu and panel. 

Horizontal companion window exists below the primary window, and is the same width as the primary window. Multiple horizontal windows can stack below the primary window, maintaining the same width.

Vertical companion window exists to the right of the primary window and is the same height as the primary window + its horizontal companion windows. Multiple vertical windows can stack beside the primary window, maintaining the same height. 

_Alternative approach: Implement React Mosaic framework within the window, allowing full flexibility in placing and splitting companion windows. :-)_

charbugs:
I added this issue to the [companion window card](https://github.com/ProjectMirador/mirador/projects/1#card-17532968) in the project board because I consider it as a prerequisite for #1754 .

aeschylus:
https://github.com/ProjectMirador/mirador/issues/1760

cbeer:
[...]

Issue #1754, opened 2019-01-30
Companion window --> repositioning button / drag handle


camillevilla:
Probably depends on pattern set up in #1752 (initial popping out implementation) 

![companion window on right](https://user-images.githubusercontent.com/5402927/52004705-a12aeb80-247c-11e9-80a9-8f8eb93df296.png)

![companion window bottom](https://user-images.githubusercontent.com/5402927/52005808-77bf8f00-247f-11e9-8a55-b0097f9b8de7.png)

### Resources
- [Interactive design](https://xd.adobe.com/view/1a628598-6843-4d25-67cb-d9e42eff6617-6c53/screen/4442ab28-e7a0-4ad4-962d-74506597c24d/Companion-right-default)
- Overall window layout diagram in #1715

### Acceptance criteria (still finalizing with design team) 
- [ ] Clicking and dragging the companion window to the bottom
  - [ ] Closes the right companion window
  - [ ] Moves content to the bottom companion window
- [ ] The image viewing area remains the same size (e.g. image is not clipped); total window area increases to accommodate the companion window

### Out of scope
- Visual niceties for dragging the window over (e.g. ghost boxes) 
![ghost-box](https://user-images.githubusercontent.com/5402927/52005700-26170480-247f-11e9-89ee-bc68780d5c7a.png)
- Keyboard controls (waiting for larger design / structure)

camillevilla:
@ggeisler @jvine I just want to clarify intended behavior for dragging the window to the right area because it isn't in the interactive mockups. If dragging left:
- the sidebar pops open the contents on the left
- window state should look like this again, no explicit "companion window" label

![screen shot 2019-01-30 at 10 13 05 am](https://user-images.githubusercontent.com/5402927/52007591-9c1d6a80-2483-11e9-867e-e4cbd5b5ebb7.png)

jvine:
I don't believe there is a drag-left for the companion window. Windows can be moved to the right or bottom of the primary window. Primary window consists of image + information panels. Companion window doesn't pop back into the info panels area. Logically if companion windows were to be dragged left, then the result would be positioned outside the primary window, to the left of the sidebar menu, taking the full height of the current layout.

camillevilla:
Thanks for the confirmation. We will not be implementing or designing drag-left. 🎉

charbugs:
@jvine @ggeisler 

[...]

Issue #1947, opened 2019-02-22
Standardized icon set


jvine:
Primarily using icons from the MUI "Sharp" theme: https://material.io/tools/icons/?search=left&icon=play_circle_outline&style=sharp

Custom icons are provided for some core features for M3 specificity, visual consistency, disambiguation with similar icons in the UI/alternate icon meanings. (svgs in comment below.)

Location | Icon
------------ | -------------
Workspace menu | @material-ui/icons/MenuSharp
Workspace fullscreen | @material-ui/icons/FullscreenSharp
Workspace fullscreen exit | @material-ui/icons/FullscreenExitSharp
Settings | @material-ui/icons/SettingsSharp
Share | @material-ui/icons/ShareSharp
Export | @material-ui/icons/SaveAltSharp
Window menu | @material-ui/icons/MoreVertSharp
Window maximize | **CustomWindowMax**
Window maximize exit | **CustomWindowMin**
Close (multiple locations) | @material-ui/icons/CloseSharp
Sidebar toggle | @material-ui/icons/MenuSharp
Tab - About | @material-ui/icons/InfoSharp
Tab - Index | **CustomIndex**
Tab - Search | @material-ui/icons/SearchSharp
Tab - Annotations | @material-ui/icons/CommentSharp
Tab - Layers | @material-ui/icons/LayersSharp
Open sidebar in panel | @material-ui/icons/OpenInNewSharp
Collapse toggle left | @material-ui/icons/ArrowLeftSharp
Collapse toggle right | @material-ui/icons/ArrowRightSharp
Expand content (within panel) | @material-ui/icons/ExpandMoreSharp
Collapse content (within panel) | @material-ui/icons/ExpandLessSharp
View - Single | @material-ui/icons/CropOriginalSharp
View -  Book | **CustomBookView**
View -  Scroll | **CustomScrollView**
View -  Gallery | **CustomGalleryView**
View -  Thumbs off | @material-ui/icons/CropDinSharp
View -  Thumbs bottom | **CustomThumbsBottom**
View -  Thumbs side | **CustomThumbsSide**
Canvas - Previous/Left | @material-ui/icons/PlayCircleOutlineSharp
Canvas - Next/Right | @material-ui/icons/PlayCircleOutlineSharp (rotate 180 degrees)
Canvas - Zoom in | @material-ui/icons/AddCircleOutlineSharp
Canvas - Zoom out | @material-ui/icons/RemoveCircleOutlineSharp
[...]

Issue #1934, opened 2019-02-20
Correct sidebar elevation and relationship with canvas


jvine:
[UI map for context](https://github.com/ProjectMirador/mirador/wiki/M3---UI-map)

Currently the sidebar is overlapping the canvas, and is at the same elevation as the window header. This seems to be a feature of the drawer component, but is not the intent of the design.

### Current layout
<img width="792" alt="screen shot 2019-02-20 at 11 26 46 am" src="https://user-images.githubusercontent.com/6945691/53119273-f39f7b00-3503-11e9-9928-8af37282a6ca.png">

- sidebar and header are at the same elevation (incorrect); canvas and right panel are at 0 elevation (correct)
- canvas is not centered; canvas navigation is partly obscured (depending on window size)

### Intended layout
<img width="798" alt="screen shot 2019-02-20 at 11 25 40 am" src="https://user-images.githubusercontent.com/6945691/53119274-f4381180-3503-11e9-952f-298f127ee59a.png">

- sidebar, canvas, and panel are all at 0 elevation; header is elevated equally above all three sections
- canvas and canvas navigation are centered; neither the sidebar nor the panel overlaps the canvas area
- there should be no shadows cast by sidebar, canvas, or panel.

mejackreed:
I think the elevation issues here were addressed in #2039 and #2018

jvine:
Looking at the current state of this, just want to point out some details. Apologies if you're already aware.

1. navigation is still at a higher elevation than the content of the sidebar, so the shadow of the window header doesn't extend the full width of the window
2. content panel still has a drop-shadow
3. `overflow-y: auto;` would be better than `overflow-y: scroll;` for the content area, so the scrollbar area isn't reserved when it's not required. 

**Current**
<img width="300" alt="screen shot 2019-03-06 at 1 24 23 pm" src="https://user-images.githubusercontent.com/6945691/53915427-bf07e500-4014-11e9-9e7b-fb73fe76163d.png">

**Should be**
<img width="300" alt="screen shot 2019-03-06 at 1 25 10 pm" src="https://user-images.githubusercontent.com/6945691/53915443-c4652f80-4014-11e9-80e3-b2a69ab357c0.png">

PR #1961, opened 2019-02-25
[WIP] Drag companion window to bottom (and stacking)

charbugs:
#1760 
#1754


src/containers/Window.js
```diff
@@ -12,8 +13,41 @@ const mapStateToProps = ({ manifests, windows }, props) => ({
...
+  companionAreaBottom: {
+    flex: '0',
+  },
+  thumbnailArea: {
+    flex: '0',
+  },
+};
+
 const enhance = compose(
   connect(mapStateToProps),
```
christopher-johnson: withStyles needs to be before connect here

src/components/CompanionWindow.js
```diff
@@ -1,6 +1,8 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
+import Divider from '@material-ui/core/Divider';
 import CloseIcon from '@material-ui/icons/Close';
+import OpenWith from '@material-ui/icons/OpenWith';
```
jkeck: Can we make this `OpenWithSharp` to match other Sharp icons now that we've standardized?

src/state/actions/window.js
```diff
@@ -117,6 +116,18 @@ export function closeCompanionWindow(windowId, companionWindowId) {
...
 
+/**
+* Move companion window between right and bottom area
+* @param {String} companionWindowId
+*/
+export function toggleAreaOfCompanionWindow(companionWindowId) {
+  return (dispatch, getState) => {
+    const position = getState()
+      .companionWindows[companionWindowId].position === 'right' ? 'bottom' : 'right';
+    dispatch(updateCompanionWindow(companionWindowId, { position }));
```
jkeck: It looks like we may be missing a unit test for this function, would it be possible/easy to add one?

src/components/CompanionWindow.js
jvine: Or can we use the drag icon listed in #1947 ?   DragIndicatorSharp

jvine:
The issue with multiple windows covering the sidebar might be partly related to #1934. The sidebar should be on the same level as the canvas and companion windows, and technically they should be able to shove each other over/make each other relatively smaller as needed - but there's still a boundary where there are too many windows to be functional. And it's complicated by the fact that the windows will be able to expand the window size in Freeform mode. 

Possibly approach is to limit the maximum area that can be used by companion windows (or maintain a minimum area of visible canvas), and once that limit is reached, block any further companion windows from being opened. Should this be a separate ticket?

src/components/CompanionWindow.js
jkeck: Sure! I suppose I misunderstood 
> I don't think this icon is directly needed yet.

in https://github.com/ProjectMirador/mirador/pull/1967#issuecomment-467540566

jkeck:
I think it's fine to push that to a separate issue (assuming we're okay w/ the interim state where that menu can become inaccessible)

aeschylus:
Compare to https://github.com/ProjectMirador/mirador/issues/1760

charbugs:
@aeschylus 
What are the implications of this comment? :confused:

jvine:
I can't see the preview on this one - I just get the blank white page. The screenshot @jkeck posted looks accurate in terms of layout but I'd like to verify.

charbugs:
Hmm, strange. Work for me here. Is there an error message on the console?

jvine:
Okay, after a few tries the preview randomly works, yay!

If this PR is just about the placement of multiple companion windows - stacking vertically, horizontally, and in relationship to the thumbnail nav - then the layout is accurate. Some later discussion (in a ticket or Slack, I don't recall) added to/superseded #1760:  the thumbnail nav is always at the bottom of the stack (or far right for vertical) and the full width (or height) of the entire window. 👍 

However:
1. what's implemented here is not a drag interaction but a side/bottom toggle. That means some other intended interactions need further specification, like reordering companion windows within a stack. 
2. there needs to be some HTML restructuring (which will apply to the sidebar as well) to make the header (title plus controls) a unit that is separate from the content. Will ticket this separately.
3. companion windows desperately need some kind of border to distinguish them from each other. They're incomprehensible without this. Curse MUI and its borderlessness!

What's here is accurate. But combining the visual design & HTML issues with the hidden sidebar issue above, and given @snydman's comments today - I think it's premature to merge this to master. @ggeisler do you agree?

ggeisler:
Yeah, it seems like there is still quite a bit of fine-tuning needed, so rather than merging now, it might make more sense to first try to work in some of the details @jvine outlined above.

charbugs:
Fine with me. So I will lable the PR as WIP.

cbeer:
I think we've extracted all the behavior into separate PRs and this can be safely closed.

PR #1989, opened 2019-02-27
In book view, next/previous increment the correct number of canvases

### LINKED ISSUE #1870: Update ViewerNavigation to work correctly in "book view"
Currently in BookView, the next/previous buttons increment canvases. They should probably increment two canvases.

mejackreed:
Fixes #1870 

In book view, next/previous increment the correct number of canvases. The last commit removes unused actions/reducers which could be reverted if we need it to.

bookviewnavcontrols


mejackreed:
That's a great question @jkeck . I've spun this out here https://github.com/ProjectMirador/mirador/issues/1994 so we don't forget.

PR #1995, opened 2019-02-28
remove provider icon from window top bar fixes #1962

### LINKED ISSUE #1962: Remove provider logo image from window top bar
When @jvine and I were working out details of typography and spacing for different components, it became clear that the complications of displaying the manifest provider's logo in the window top bar outweigh the benefits. We recommend removing the logo from the window top bar. So in the current version of the app things would look like this:

mirador-2

mejackreed:
## After

screen shot 2019-02-28 at 10 47 11 am

### DIFF
--- a/src/components/WindowTopBar.js
+++ b/src/components/WindowTopBar.js
@@ -7,7 +7,6 @@ import CloseIcon from '@material-ui/icons/CloseSharp';
 import Toolbar from '@material-ui/core/Toolbar';
 import AppBar from '@material-ui/core/AppBar';
 import classNames from 'classnames';
-import WindowIcon from '../containers/WindowIcon';
 import WindowTopMenuButton from '../containers/WindowTopMenuButton';
 import WindowTopBarButtons from '../containers/WindowTopBarButtons';
 import ns from '../config/css-ns';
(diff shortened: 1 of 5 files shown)


src/components/WindowTopBar.js
```diff
@@ -7,7 +7,6 @@ import CloseIcon from '@material-ui/icons/CloseSharp';
 import Toolbar from '@material-ui/core/Toolbar';
 import AppBar from '@material-ui/core/AppBar';
 import classNames from 'classnames';
-import WindowIcon from '../containers/WindowIcon';
```
cbeer: Can we throw away `containers/WindowIcon` now too? I don't know if it's also a good time to refactor/rename `component/WindowIcon` for its only remaining consumer too...

mejackreed: Looking at this a bit. I'm wondering if alternatively we should just import and use `containers/WindowIcon` in `ManifestListItem` rather than just passing through logo. Is there a reason not to just use the container there?

cbeer: Sure 🤷‍♂️

mejackreed: I've pushed up a change that uses `containers/WindowIcon` rather than relying on it being passed through.

Issue #1975, opened 2019-02-26
Overflow settings creating inappropriate scrollbars/placeholders


jvine:
There are three places where the app is incorrectly getting scroll bars or placeholders for scrollbars.

<img width="509" alt="screen shot 2019-02-26 at 10 19 16 am" src="https://user-images.githubusercontent.com/6945691/53436776-d01b7b00-39b0-11e9-8873-ee39b633f4b1.png">

##  1. Workspace in Mosaic mode 
- [x] `.mirador-workspace` has `overflow: scroll;` which is creating an unnecessary scroll-bar sized border around the mosaic workspace. Not sure what the best setting here is, or how Freeform will use this space (do we want scrollbars on Freeform?) - setting it to `overflow: auto;` or removing the setting altogether fixes the issue for Mosaic.

<img width="506" alt="screen shot 2019-02-26 at 10 07 46 am" src="https://user-images.githubusercontent.com/6945691/53437054-7f585200-39b1-11e9-8054-91f522aa5da3.png">

## 2. Thumbstrip has vertical scrollbar
- [ ] Goal is this: 
- companion windows may scroll in both directions when they have non-thumb content
- thumbnail strip may scroll horizontally but should never scroll vertically. Setting `overflow-y: hidden;` on the inner div (`class="ReactVirtualized__Grid"`) seems to work but might not be the most robust solution.

<img width="504" alt="screen shot 2019-02-26 at 10 08 36 am" src="https://user-images.githubusercontent.com/6945691/53438667-330f1100-39b5-11e9-9fac-ae6ddbf17a1f.png">

## 3. Application has vertical scrollbar
- [x] Not clear to me how/why this is happening, but the application as a whole should never have a vertical scroll bar. Elements within the app (menus, workspace, even the app bar etc.) can scroll, but not the app. Currently the app always seems to be ~2x as high as the viewport - it's impossible to stretch the window to eliminate the scrollbar. 

Outcome with all three overflows resolved should look like this:
<img width="509" alt="screen shot 2019-02-26 at 10 09 34 am" src="https://user-images.githubusercontent.com/6945691/53439004-ef68d700-39b5-11e9-9322-602d43764fdf.png">

<img width="508" alt="screen shot 2019-02-26 at 11 06 00 am" src="https://user-images.githubusercontent.com/6945691/53439259-85046680-39b6-11e9-98a9-8e1b9e2735d3.png">

mejackreed:
[...]

PR #1997, opened 2019-02-28
Fix scrolling issues in window

mejackreed:
This PR fixes parts 1 & 3 of #1975 .

I was not able to repeat part 2, but still think it might be an issue so leaving the issue open until we can better look at padding/margins/spacing for the ThumbnailNavigation.

*Note* Border was added to `embedding.html` to allow for easier debugging for team.

### DIFF
--- a/__tests__/integration/mirador/basic.test.js
+++ b/__tests__/integration/mirador/basic.test.js
@@ -8,6 +8,7 @@ describe('Basic end to end Mirador', () => {
   });
   it('loads a manifest and displays it', async () => {
     await expect(page).toClick('#addBtn');
+    await expect(page).toClick('.mirador-add-resource-button');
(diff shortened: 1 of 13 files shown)


mejackreed:
@jkeck I've updated things to pass effectively. I'm still wondering if I should cherry-pick your commit in that changes the fullscreen library.

jkeck:
I think as long as we don't run into problems w/ the 100% height/width then it's not a big deal.

It might be a good idea to cherry-pick it though, because the library that I swap out w/ in my branch appears to be used more, and potentially a bit more maintained.  The one we're using now (`react-fullscreen-crossbrowser`) claims that it's a typescript conversion of the one what I changed to use in my branch (`react-full-screen`) but the typescript version hasn't been updated since 2017 and the original appears to have gotten a few updates and some releases w/i the last few months.

jkeck:
Removing myself as a reviewer now that I have a commit in this PR.

PR #1998, opened 2019-02-28
Apply design style tweaks to the canvas nav panel

### LINKED ISSUE #1978: Thumbnail canvas nav layout corrections
Canvas navigation with thumbnails in the sidebar is easily confused with the thumbnail navigation panel at the moment. Some layout adjustments will help with this. 

- Thumbnails are meant to be very small: 50px maximum width/height
- Canvas labels should be default font size (not 8pt)
- Image and text should be top-aligned (`align-items: flex-start;`)
- Image "column" should be fixed width (50px) so labels are all left-aligned

Compare current (left) & intended (right) displays:

Current | Intended
--------- | ---------
screen shot 2019-02-26 at 1 26 06 pm | screen shot 2019-02-26 at 5 15 50 pm
**Current** | **Intended**
screen shot 2019-02-26 at 5 06 29 pm | screen shot 2019-02-26 at 4 55 46 pm

(Some planned styling will improve whitespace; we may consider truncating at a standard number of lines to avoid the imbalance of long titles.)

cbeer:
This also refactors the `CanvasThumbnail` component to accept width/height/aspect ratio constraints so it can be used in both the canvas nav and the thumbnail bar relatively interchangeably. 

Fixes #1978 

screen shot 2019-02-28 at 15 35 21

### DIFF
--- a/__tests__/fixtures/version-2/emptyCanvas.json
+++ b/__tests__/fixtures/version-2/emptyCanvas.json
@@ -1,0 +1,7984 @@
+{
+    "@context": "http://iiif.io/api/presentation/2/context.json",
+    "@id": "http://demos.biblissima-condorcet.fr/iiif/metadata/florus-dispersus/manifest.json",
+    "@type": "sc:Manifest",
+    "label": "Codex Florus dispersus",
+    "metadata": [
+       {
+		"label":"Object type",
+		"value":"Manuscrit dispersé"
+       },
+       {
+            "label": "Repository",
+            "value": [
+                "Paris, Bibliothèque nationale de France",
+                "Genève, Bibliothèque de Genève",
+                "Saint-Pétersbourg, Bibliothèque nationale de Russie"
+            ]
+        },
+        {
+            "label": "Shelfmark",
+            "value": [
+                "BnF, Latin 11641",
+                "BGE, Latin 16",
+                "BnR, Latin F.papyr.I.1"
+            ]
+        },
+        {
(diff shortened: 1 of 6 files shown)


aeschylus:
Codeclimate basically just penalises for nested if statements because it "interrupts linear flow".

An alternative is to write a switch statement that covers each case of the layout permutations separately. Such a case statement could then be made easier to read by writing another utility function that maps each distinct constellation of conditions to a readable string (or comments). For example:

``` javascript
layout = layoutFromParams(maxHeight, maxWidth, aspectRatio)

switch (layoutType) {
  case 'heightConstrained':
  case 'widthConstrained':
  case 'heightAndWidthConstrained':
  case 'heightWidthAndPortraitConstrained':
  case 'heightWidthAndLandscapeConstrained':
  default:
}
```
where `layoutFromParams` would look something like:

``` javascript
if (maxHeight && maxWidth && aspectRatio && (desiredAspectRatio < aspectRatio)) {
   return 'heightWidthAndPortraitConstrained';
} else if (maxHeight && maxWidth && aspectRatio && !(desiredAspectRatio < aspectRatio)){
   return 'heightAndWidthAndLandscapeConstrained';
} else if (maxHeight && maxWidth) {
      return '...etc'
...

Yes, it's sort of dumb, but can be read linearly and maps to tests better. You can also factor out branches into functions that receive arguments. for example, something like `isPortrait(width, height, aspectRatio)`.

PR #2002, opened 2019-03-01
1874 import workspace config initial implementation

glefi:
I am still facing the following problems:

I wasn't able to update the viewport for newly created windows, because there seems to be an a temporal dependency between the dispatch of `ADD_WINDOW` and `UPDATE_VIEWPORT`. The viewport can only be updated when the window is already created and there seems to be no possbility to pass an initial viewport configuration when creating a new window.

I wasn't able to update the viewport for newly created windows, because there seems to be an a temporal dependency between the dispatch of `ADD_WINDOW` and `UPDATE_VIEWPORT`. The viewport can only be updated when the window is already created and there seems to be no possbility to pass an initial viewport configuration when creating a new window.

While trying to write an integration test for the confiugration import I was struggeling with an expect-puppeteer bug (https://github.com/smooth-code/jest-puppeteer/issues/150) that prevents me from fill out the config import textarea. I also tried setting the textarea's value by manipulation the DOM element directly:

```
await page.exposeFunction('getTextValue', () => return textValue ); // textValue is the stringified config object
await page.waitForSelector('#workspace-import-input');
await page.$eval('#workspace-import-input', el => window.getTextValue());
```

But this way I ran into a timeout.


glefi:
I spoke with Jennifer an Gary. The beta error dialog can be merged and designed later. So the pr can be merged after reviewing it.

src/components/App.js
```diff
@@ -4,17 +4,27 @@ import classNames from 'classnames';
 import { MuiThemeProvider, createMuiTheme } from '@material-ui/core/styles';
 import Fullscreen from 'react-full-screen';
 import { I18nextProvider } from 'react-i18next';
+import _ from 'lodash';
```
mejackreed: Can you import lodash using a similar pattern here: https://github.com/ProjectMirador/mirador/blob/167b86007e48c8b207a48f9c74b7f810d12948c1/src/lib/ManifestoCanvas.js#L1

I think doing it that way limited the file size increases (in my testing at least a few weeks ago). I wonder if that is still the case.

src/components/App.js
```diff
@@ -4,17 +4,27 @@ import classNames from 'classnames';
...
  * This is the top level Mirador component.
  * @prop {Object} manifests
  */
 export class App extends Component {
+  /**
+   *
+   * @param {Object} error the object to check
+   */
+  static hasError(error) {
+    return !isNull(error) && !isUndefined(error) && !isEmpty(error);
```
charbugs: I think the error rendering logic shout not be handled  by App. Rather App should simply render the ErrorDialog without any prop. Everything else should be done by the ErrorDialog. The way it is now the error logic is spread over two places.

src/components/Window.js
```diff
@@ -41,9 +41,9 @@ export class Window extends Component {
       <div id={window.id} className={ns('window')}>
         {this.wrappedTopBar()}
         <WindowMiddleContent
-          window={window}
           manifest={manifest}
           sideBarOpen={window.sideBarOpen}
+          window={window}
```
charbugs: If the intention is the sort the props by name then it's off topic. Same applies to WindowMiddleContent and WindowViewer.

src/components/App.js
```diff
@@ -68,7 +78,12 @@ export class App extends Component {
                 isWorkspaceAddVisible
                   ? <WorkspaceAdd />
                   : <Workspace />
-               }
+              }
+              {
+                <ErrorDialog
```
charbugs: The test for that must be updated.

src/containers/WorkspaceImport.js
```diff
@@ -0,0 +1,32 @@
...
+const mapStateToProps = state => ({ state });
+
+/**
+ * mapDispatchToProps - used to hook up connect to action creators
+ * @memberof App
+ * @private
+ */
+const mapDispatchToProps = {
+  addError: actions.addError,
+  importConfig: actions.importWorkspace,
```
charbugs: Does the name switching here have a reason?

src/containers/WorkspaceImport.js
```diff
@@ -0,0 +1,32 @@
...
+import miradorWithPlugins from '../lib/miradorWithPlugins';
+import { WorkspaceImport } from '../components/WorkspaceImport';
+import * as actions from '../state/actions';
+
+/**
+ * mapStateToProps - to hook up connect
+ * @memberof WorkspaceImport
+ * @private
+ */
+const mapStateToProps = state => ({ state });
```
charbugs: As far as I can see, there is no need for this line.

src/components/ErrorDialog.js
```diff
@@ -0,0 +1,62 @@
...
+   * render
+   * @return
+   */
+  render() {
+    const {
+      errors, removeError, t,
+    } = this.props;
+
+    /* extract 'items' value and get first key-value-pair (an error) */
+    const error = first(values(omit(errors, 'items')));
```
charbugs: :heart_eyes:

src/state/actions/workspace.js
```diff
@@ -68,3 +73,40 @@ export function toggleWorkspaceExposeMode() {
...
+ * importWorkspace - action creator
+ */
+export function importWorkspace(stateExport) {
+  return (dispatch, getState) => {
+    dispatch(importConfig(stateExport.config));
+    const { viewers } = stateExport || {};
+    const imWins = values(stateExport.windows);
+
+    /* re-use existing windows */
+    const exIds = values(getState().windows).map((exWin) => {
```
aeschylus: Can you explain a little bit of the context around this? This seems like an implicit state mutation ("reusing" property values). What were some of the blocks you ran into here that this addresses?

aeschylus: (Immutability is extremely important for maintainability and debugging.)

src/state/actions/workspace.js
```diff
@@ -68,3 +73,40 @@ export function toggleWorkspaceExposeMode() {
     type: ActionTypes.TOGGLE_WORKSPACE_EXPOSE_MODE,
   };
 }
+
+/**
+ * importWorkspace - action creator
+ */
+export function importWorkspace(stateExport) {
```
cbeer: Are there tests for this action? Also, do we want to try to refactor this to avoid multiple dispatches before merging (e.g. as in #2121)? I wonder if that'd help address some of @aeschylus's concerns too.

src/state/reducers/windows.js
```diff
@@ -90,10 +90,6 @@ export const windowsReducer = (state = {}, action) => {
           y: action.payload.size.y,
         },
       };
-    case ActionTypes.NEXT_CANVAS:
```
cbeer: This was dead code, yeah? I wonder if we should put this in a separate PR for clarity?

src/state/selectors/index.js
```diff
@@ -272,3 +272,11 @@ export function getLanguagesFromConfigWithCurrent(state) {
     current: key === language,
   }));
 }
+
+/**
+ * Returns the latest error from the state
```
cbeer: Is this right? Is it the latest error or the first error?

src/state/actions/workspace.js
cbeer: I wonder if this is related to state mutation, but I noticed if I copied some state, added some windows, and then tried to import the previous state, I got a javascript error:

glefi: > Can you explain a little bit of the context around this? This seems like an implicit state mutation ("reusing" property values). What were some of the things you ran into here that required a reuse pattern?

The idea is the following: If the workspace, that is already opened when starting the import process, already contains windows e. g. viewer instances, we can re-use these windows by dispatching an `UPDATE_WINDOW` action to optimize the performance of the import process. All surplus windows will be closed by dispatching a `REMOVE_WINDOW` action. If, on the other hand, the number of windows to be imported is higher than the number of the existing ones, we need to create new instances by dispatching an `ADD_WINDOW` action. But no state manipulation is done directly here.

src/components/ErrorDialog.js
```diff
@@ -0,0 +1,62 @@
...
+    const hasError = !isUndefined(error);
+    return (
+      <div>
+        { hasError && (
+          <Dialog id="workspace-settings" onClose={() => removeError(error.id)} open={hasError}>
```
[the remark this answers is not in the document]
glefi: Currently there is only one error dialog for the whole workspace. But in order to show it window related, the id would be wrong. I'll remove this.

src/state/reducers/windows.js
glefi: Yes, there is no reducer handling this action anymore. I think it was replaced by `SET_CANVAS`. I'll extract this and put it into a separate pr.

src/state/actions/workspace.js
glefi: Added one.

Do you suggest to bundle the whole import process into one single dispatch or only the three dispatches that are done when replacing or creating a new window?

src/components/WorkspaceImport.js
```diff
@@ -0,0 +1,88 @@
...
+   */
+  render() {
+    const {
+      handleClose, open, t,
+    } = this.props;
+    return (
+      <Dialog id="workspace-settings" open={open} onClose={handleClose}>
+        <DialogTitle id="form-dialog-title">{t('import')}</DialogTitle>
+        <DialogContent>
+          <TextField rows="15" multiline onChange={this.handleChange} />
```
cbeer: If we add `variant="filled"`, does that allow us to remove the theme override?

src/components/WorkspaceMenu.js
```diff
@@ -153,6 +157,16 @@ export class WorkspaceMenu extends Component {
...
           </MenuItem>
+          <MenuItem
+            aria-haspopup="true"
+            onClick={(e) => { this.handleMenuItemClick('importWorkspace', e); handleClose(e); }}
+            aria-owns={exportWorkspace.AnchorEl ? 'workspace-import' : undefined}
+          >
+            <ListItemIcon>
+              <ImportIcon />
+            </ListItemIcon>
+            <Typography varient="inherit">{t('importWorkspace')}</Typography>
```
cbeer: If you don't discover this due to merge conflicts, I believe this should be `variant="body"` per the new typography specs (note the typo in the key too; we've copy+pasted that all over)

src/state/actions/workspace.js
cbeer: Yeah, I was wondering if it could be done in a single dispatch, and have the reducers pick up the elements or clear state as needed.

mejackreed:
Are there additional things that need review/update here?

glefi:
> Are there additional things that need review/update here?

The import of the companion windows was added just recently, so maybe someone could/should have a look at this.

PR #2004, opened 2019-03-01
put zoom controls in correct place fixes #1977

### LINKED ISSUE #1977: Zoom controls should be aligned to visible canvas, not the window
When zoom controls are active and there's a companion window open, the zoom controls overlay the content of the companion window:

screen shot 2019-02-26 at 12 51 08 pm

Zoom controls should be associated with the canvas in each open window. 

screen shot 2019-02-26 at 12 51 38 pm

mejackreed:
## After

screen shot 2019-03-01 at 7 18 25 am

### DIFF
--- a/src/components/OpenSeadragonViewer.js
+++ b/src/components/OpenSeadragonViewer.js
@@ -209,8 +209,8 @@ export class OpenSeadragonViewer extends Component {
           ref={this.ref}
         >
           { children }
+          <ZoomControls windowId={windowId} />
         </div>
-        <ZoomControls windowId={windowId} />
       </>
     );
   }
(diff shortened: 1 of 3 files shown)


src/components/OpenSeadragonViewer.js
```diff
@@ -209,8 +209,8 @@ export class OpenSeadragonViewer extends Component {
           ref={this.ref}
         >
           { children }
+          <ZoomControls windowId={windowId} />
         </div>
-        <ZoomControls windowId={windowId} />
       </>
```
cbeer: Should we also remove the `<></>`?

mejackreed: /shrug no idea. It might be nice to leave incase we need to drop anything else in there in the future and reduce the diff? 

I could go either way

cbeer: WFM.

Issue #1936, opened 2019-02-21
Menu issues: location, behaviour, icons


jvine:
<img width="782" alt="screen shot 2019-02-20 at 3 34 20 pm" src="https://user-images.githubusercontent.com/6945691/53132604-070f0e00-3525-11e9-937c-ecdc6b362c44.png">

## Location
1. **Menus should appear next to or below, but not covering, the element that generates them**. The `Toggle Menu Grow` style here https://material-ui.com/demos/menus/#menulist-composition seems closest to the intended behaviour, though the keyboard interaction is slightly different than MUI.

- [ ] Align sidebar menus with the top/right (in LTR mode) or top/left (in RTL) of the icon.
- [ ] Align top bar menus with the bottom/right (in LTR mode) or bottom/left (in RTL) of the icon.

(By "icon" I mean the svg element (24px) of the icon button (svg + padding).)

2. **The window menu should be visually contained within the window that generated it**. It should not overlap multiple windows. Its max width should match the window width; menu content can wrap if necessary. 

## Selection
3. **The icon button that generated the menu should remain in "active" state** while the menu is open. Example from gmail:

<img width="290" alt="screen shot 2019-02-20 at 4 11 38 pm" src="https://user-images.githubusercontent.com/6945691/53134078-76d3c780-352a-11e9-91fe-e9d1e7758c18.png">

4. **No menu item should be highlighted by default.** Current behaviour allows for two items to be in hover state, which is inappropriate.

<img width="277" alt="screen shot 2019-02-20 at 2 49 39 pm" src="https://user-images.githubusercontent.com/6945691/53130351-d4621700-351e-11e9-83d4-c9af28934bda.png">

This seems to be a factor of the menu overlaying the icon button, so may be fixed by changing the location of the menu (item 1). Menus should open with no option in focus; user can hover/set focus using the mouse or arrow keys. 

5. **Dividers should not be focusable elements.** 

Current implementation is using `<hr class="whatever" tabindex="-1">` as the divider element. For whatever reason, this element is focusable, but it should not be.

<img width="272" alt="screen shot 2019-02-20 at 4 27 33 pm" src="https://user-images.githubusercontent.com/6945691/53134595-a4217500-352c-11e9-9df2-fa8f43266fe0.png">

MUI documentation shows this as the divider, FWIW:
`<li class="mdc-list-divider" role="separator"></li>`

## Layout & icons

[...]

PR #2008, opened 2019-03-01
Set the WindowTopMenu MUI-Menu container to the node (by windowId).

### LINKED ISSUE #1785: WindowTopBar Buttons don't work in fullscreen

jkeck:
This is an alternate approach to fix #1785.  Setting an MUI container prop the result of `document.getElementById` is [not](https://github.com/ProjectMirador/mirador/blob/63b6edb45a9257edb56426af72201fff1a4d75df/src/components/WindowSideBar.js#L33) [unheard](https://github.com/ProjectMirador/mirador/blob/63b6edb45a9257edb56426af72201fff1a4d75df/src/components/WindowSideBar.js#L51) of.

However; this causes an issue w/ side-by-side windows where one window will overlap the modal.  I poked at some z-indexes briefly w/ no luck, but this might be the easiest way to address this bug if we can get the modal to display properly.

window-overlay

### DIFF
--- a/src/components/WindowTopMenu.js
+++ b/src/components/WindowTopMenu.js
@@ -19,7 +19,13 @@ export class WindowTopMenu extends Component {
 
     return (
       <>
-        <Menu id={`window-menu_${windowId}`} anchorEl={anchorEl} open={Boolean(anchorEl)} onClose={handleClose}>
+        <Menu
+          id={`window-menu_${windowId}`}
+          container={document.getElementById(windowId)}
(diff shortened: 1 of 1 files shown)


jvine:
There's a related ticket around menu behaviour here: #1936 that would affect the placement/sizing of the window options menu so it doesn't overlap/ isn't overlaid by other windows.

Issue #1620, opened 2019-01-16
Investigate strategy for managing more complex asynchronous process typologies


aeschylus:
Right now using thunks, we have a working pattern for managing simple requests and allowing the UI to provide synchronous rendering of asynchronous state by storing the request metadata in the redux store. How scalable is this, and what other asynchronous area

## Asynchronous Processes
1. Requesting Manifests
1. Requesting Info.json files
1. Requesting Thumbnails
1. Canceling requests that are no longer required to complete by the user's focus (close the window that was still loading thumbnails)
1. Rendering/Displaying Auth Flow State
1. others...?

## Strategies
1. Continue using a combination of thunks in actionCreators to track request status (essentially split
1. Redux Saga
1. RxJS/Observable Streams
1. `async` + `await`

christopher-johnson:
this is essential to formulate.  There is the potential for an "async request cascade" where hundreds of chained async requests could be required to construct a single rendered manifest in the v3 API.  

The manifest at is most basic form is a list of canvas objects with id and type in the first `Items` array.  The spec does not require the manifest to embed nested items.  The client should be prepared to check for the existence of embedded items and if not present then dereference the object id (with an async action) and subsequently deserialize and normalize the response into state.  This can occur with several different types (e.g. `Canvas`, `AnnotationPage`, `Annotation`, `Range`, etc.)  

This is more likely to occur in the AnnotationPages `Items` array, where there may be both a page that contains embedded painting annotations and distinct `AnnotationPage` ids that must be dereferenced to get annotations with other motivations (similar to "`otherContent`" in v2).   

It is necessary to understand how to abstract fetch action dispatching so that the rendering components can be agnostic to the resolution, reduction and evaluation of the API responses.  The rendering component itself should not contain any business logic that does conditional fetch dispatch as I suspect that this may negatively impact performance and usability.
  
Example:
```javascript
if (items) {
  handleItemsResponse(items);
} else {
  fetchItems(objectURI);
}
```
It should just get a denormalized and contracted object or value (or null) from its selector.  IMO, the business logic of async response handling and reduction/resolution should be clearly separated from the rendering components...

aeschylus:
Might be useful soon: https://www.robinwieruch.de/redux-state-keys/

christopher-johnson:
Error handling using try catch blocks.

mejackreed:
Closing this ticket, we have been using strategy 1 right now. Let's reopen if new options come up or there are new things to discuss.

PR #2010, opened 2019-03-01
Adds in basic otherContent annotation fetching

### LINKED ISSUE #2006: Annotations - Retrieve annotations from an annotations list in a IIIF v2 manifest

mejackreed:
Fixes #2006 

This PR is split into two commits:
 - The first, implements the backend state changes needed to support fetching annotations.
 - The second, implements a basic (and intentionally incomplete) annotation fetching from our component system w/ tests.

annos


src/components/WindowViewer.js
```diff
@@ -26,14 +26,18 @@ export class WindowViewer extends Component {
    * Request the initial canvas on mount
    */
   componentDidMount() {
-    const { fetchInfoResponse } = this.props;
+    const { fetchInfoResponse, fetchAnnotation } = this.props;
 
     if (!this.infoResponseIsInStore()) {
       this.currentCanvases().forEach((canvas) => {
-        const { imageInformationUri } = new ManifestoCanvas(canvas);
+        const manifestoCanvas = new ManifestoCanvas(canvas);
```
aeschylus: I think it made some sense as a first pass to include so much of this request logic in the OSD component early on, but I wonder if we can think about how to avoid using lifecycle methods for this. As this gets more complex, and we need cancellations/editing/etc., this is going to become very difficult to debug. Can we hold off on the component side of this until we make time to think through this?

The general condition I'm trying to avoid is requests being initiated from inside components for any reason. This means they exist outside of our state management system in a way that is invisible and now has to be explicitly managed.

aeschylus: In this case, for example, we know in the state that a given window is focusing a given canvas. So initiating these requests can be done in actions (namely on next/previous/setcanvas/addWindow, etc.), using our existing pattern. If that starts to get too complex, then perhaps we should consider using @bodo22's suggestion of refactoring to sagas (https://github.com/ProjectMirador/mirador/pull/1667), and/or discuss it as an architectural pattern in https://github.com/ProjectMirador/mirador/issues/1620.

mejackreed: Maybe we can jump on a call to discuss this. My main concern is holding up progress on additional annotation tickets to implement a refactor/new architectural pattern that we aren't yet using.

PR #2011, opened 2019-03-01
Make the workspace toolbar responsive; fixes #1837

### LINKED ISSUE #1837: Responsive behavior: reposition workspace control bar based on viewport aspect ratio
Looking at these designs, should we also have this reposition?: https://github.com/ProjectMirador/mirador/issues/1833

cbeer:


mejackreed:
I'm noticing a slight gap and "jump" up when pressing the "+" button. Are you seeing that?

gap

cbeer:
Yep, I must have had a local browser override that never made it into the stylesheet. Fixed.

PR #2013, opened 2019-03-01
Change WorkspaceFullScreenButton to include the workspace sidebar as well

### LINKED ISSUE #1974: Fullscreen should be the entire app not just the workspace viewport
Currently the Application Bar/WorkspaceMenu are not available in fullscreen but they should be.

See https://github.com/ProjectMirador/mirador/pull/1967#issuecomment-467540566

Note that since the Application Bar/WorkspaceMenu will now be visible in fullscreen we'll need to modify the fullscreen icon so that it changes to `FullscreenExitSharp` when in fullscreen mode.

This also has potential implications for #1785/#1813 (and we may be able to address that issue holistically with this change)

jkeck:
Closes #1974 

Note that this also introduced the bug #1785 for the menu in the workspace sidebar (since it is now visible).

## Before
fullscreen-before

## After
fullscreen-after

### DIFF
--- a/locales/en/translation.json
+++ b/locales/en/translation.json
@@ -19,6 +19,7 @@
     "dismiss": "Dismiss",
     "downloadExport": "Download/Export",
     "downloadExportWorkspace": "Download/export workspace",
+    "exitFullScreen": "Exit full screen",
     "fetchManifest": "Add",
     "fullScreen": "Full Screen",
     "hideZoomControls": "Hide zoom controls",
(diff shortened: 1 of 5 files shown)


locales/en/translation.json
```diff
@@ -19,6 +19,7 @@
     "dismiss": "Dismiss",
     "downloadExport": "Download/Export",
     "downloadExportWorkspace": "Download/export workspace",
+    "exitFullScreen": "Exit full screen",
```
jkeck: If anybody has a german translation for this, I'm happy to add it.

Issue #1926, opened 2019-02-20
General refactoring of reducers and actions


charbugs:
I like to make a general refacotring of our action and reducer implementation. 

### Context

As you will have noticed, things in the action and reducer files are getting more complex and we're starting to mixing up different patterns to accomplish things. So far, we mainly did state updates that only changes a certain state slice, e.g. changing canvas index of a window, adding a new manifest. But as, for example, PR #1922 shows there will be cases where we have to deal with actions that do more complex things, like closing a window, what implies removing the window and removing all the companion windows that belong to that window. To enable convenient implementations of such features we should apply more predictable patterns for state management.

### Problems with our reducers/actions

* For my understanding there are two general aproaches to handle state in redux. You can either put many logic in the reducer and keep action creators very simple, or the other way around, having only simple reducers and let the actions do the logic (the latter I prefer). We are currently not decided about which way to go and do sometimes so, sometimes so.

* There are different kind of ways to passing properties to action creators, sometimes single objects are required that hold the properties to process, sometimes the properties have to be passed as single parameters.

* In the reducers, the expressions to merge state without mutating it are sometimes very large and hard to read. With libraries like immutable.js this can be done with one-liners.

* The execution path of the application is hard to follow if we handle a certain action type in different reducers. I know that's a nifty feature of redux, but I think it's not a good one.

### Suggestion

I created a [proof of concept](https://github.com/charbugs/store-refactoring-fiddle) for a pattern that seems reasonable to me. The general idea is to make reducers very simple functions that can only add, update and remove the data they are responsible for. Also, for each reducer there is a set of basic action creators (add, update, remove) with an identical signature. On top of that there are combined/derived action creators that use this basic actions to accomblish more specific things to the UI. That is, there is a general predictable abstration of reducer logic. Actually, there should be no need to touch a reducer at all.

I'm currently figuring out if all use cases that we have in our code base so far can be accomplished with that pattern.

I would be happy about your reviews of the POC and comments on that.

mejackreed:
I'd be interested to see where this goes. A bit of historical underpinnings to the diverse ActionTypes. We were originally thinking that adopters who had customized Mirador instances would subscribe and use the provided ActionTypes similar to how current Mirador adopters use the event system. While I like the simplicity of the approach in the proof of concept, do you think there would also be a way to provide additional ActionTypes for external usage?

[...]

PR #2035, opened 2019-03-04
Basic setup for reducer pattern (part of #1926)

charbugs:
Add functions to create reducers and basic actions atomatically.

See also: 

* https://github.com/ProjectMirador/mirador/wiki/M3-State-Management:-Reducers,-Actions,-Events
* https://github.com/charbugs/store-refactoring-fiddle

### DIFF
--- a/__tests__/src/reducers/createActions.test.js
+++ b/__tests__/src/reducers/createActions.test.js
@@ -1,0 +1,143 @@
+import {
+  createTableReducerActions,
+  createSingletonReducerActions,
+} from '../../../src/state/reducers/createActions';
+
+describe('table reducer actions', () => {
+  const actionTypes = {
+    create: 'CREATE_ITEM',
+    update: 'UPDATE_ITEM',
+    delete: 'DELETE_ITEM',
+    order: 'SET_ITEM_ORDER',
+  };
+
+  const idPrefix = 'item';
+
+  const defaultProps = {
+    foo: 'bar',
+  };
+
+  describe('createTableReducerActions', () => {
+    it('should return object with correct function names', () => {
+      const actionCreators = createTableReducerActions(actionTypes);
+      expect(actionCreators.createItem).toBeInstanceOf(Function);
+      expect(actionCreators.updateItem).toBeInstanceOf(Function);
+      expect(actionCreators.deleteItem).toBeInstanceOf(Function);
+      expect(actionCreators.setItemOrder).toBeInstanceOf(Function);
+    });
+  });
+
+  describe('create action creator', () => {
+    it('should return correct action type', () => {
+      const { createItem } = createTableReducerActions(actionTypes);
+      const action = createItem();
+      expect(action.type).toBe('CREATE_ITEM');
+    });
+
+    it('should set the passed id to the action', () => {
+      const { createItem } = createTableReducerActions(actionTypes, idPrefix, defaultProps);
+      const action = createItem({}, 'item-xyz');
+      expect(action.id).toBe('item-xyz');
+    });
+
+    it('should create an id if no id is given.', () => {
+      const { createItem } = createTableReducerActions(actionTypes, idPrefix, defaultProps);
+      const action = createItem();
+      expect(typeof action.id).toBe('string');
+    });
+
+    it('should merge default properties to payload', () => {
+      const { createItem } = createTableReducerActions(actionTypes, idPrefix, defaultProps);
+      const action = createItem();
+      expect(action.payload).toMatchObject(defaultProps);
+    });
+
+
+    it('should merge passed data object to payload', () => {
+      const { createItem } = createTableReducerActions(actionTypes, idPrefix, defaultProps);
+      const action = createItem({ bubu: 'kiki' });
+      expect(action.payload).toMatchObject({ foo: 'bar', bubu: 'kiki' });
+    });
+
+    it('passed data object can overwrite default props and id', () => {
+      const { createItem } = createTableReducerActions(actionTypes, idPrefix, defaultProps);
+      const action = createItem({ foo: 'kiki', id: 42 });
+      expect(action.payload).toEqual({ foo: 'kiki', id: 42 });
+    });
+  });
+
+  describe('update action creator', () => {
+    it('should return correct action type', () => {
+      const { updateItem } = createTableReducerActions(actionTypes);
+      const action = updateItem();
+      expect(action.type).toBe('UPDATE_ITEM');
+    });
+
+    it('should set the passed id to the action', () => {
+      const { updateItem } = createTableReducerActions(actionTypes);
+      const action = updateItem(33, {});
+      expect(action.id).toBe(33);
+    });
+
+    it('should set the data object to the action', () => {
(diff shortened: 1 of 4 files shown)


aeschylus:
First of all, thank you for this extremely detailed and well-researched description @charbugs. I have read through your wiki page and this PR, and think I understand what the proposal is. I still need to look over the repo you posted (that's very helpful, thank you), and I have not read the entire "structuring reducers" section of the redux docs yet, but will do so this week.

My initial reaction is that, while I understand "what" is being proposed, I don't understand "why" we should want this. Right now it feels like far more abstraction than necessary at this point, and makes it much more difficult to see at a glance all the transformations that can be applied to the state (as one currently can in the action-types list). I'm just encountering this now, so it's possible I've missed a few things that have already been answered. I'll enumerate my questions here:
1. **How does this make the code easier to understand both for ourselves and new contributors?** 
Currently, I can see all actions in a short list in the action-types file. I know how to create them easily and there is a template for doing so that is natural from a feature-development standpoint. With this proposed pattern, we now need to translate our feature into tables before working on it. This seems unnecessary to me, and one of the things I'm enjoying about redux is that I don't have to think as much about the data model when developing a feature. The action + reducer mechanism is always immutable, so I'm safe putting data at the top level, without over-modelling it. Maybe this is naïve.
2. **Is there currently a problem with our state management approach?** It seems there has been some confusion over the state normalisation, and perhaps some of the objections in this list will be addressed by making us atomise/normalise state more granularly with this pattern, but in general I think the way we are currently managing state seems to be working for us. What problems are we running into that makes this approach necessary?
3. **This will encourage logic to be pushed to components** If any object can be arbitrarily updated in several different ways from different components, there will be implicit information in the component implementations about the business logic. For example, one component might "UPDATE_WINDOW" by passing in an object that happens to change its size or position, while another "UPDATE_WINDOW" call might change the view settings or register the state of a manifest request. It is no longer possible to debug these in the redux tools because the type of transformation is no longer in the name of the transformation. These same calls now have multiple possible signatures, which will be difficult to debug and document.
4. **Where will request states go?** How are we going to update request states from API responses? Are these also going to be general UPDATE actions on an object? How would this be done in Sagas, which expect to receive actions. Not all updates need to send requests. Do we need to diff the passed-in update description and add logic to determine when the update requires a request and when it does not? Our more specific named action-types currently allow us to know from the type of transformation whether a request is entailed.

The biggest benefit of this I see is the normalisation guidance, which could be very helpful, but the loss of event types with specific and natural names seems large to me, especially for our developer consumers. I think it's important to come to a strategy on our normalisation, and this is a huge step in that direction, and very thorough work. How can we address the above concerns?

aeschylus:
@mejackreed has directed me to https://github.com/ProjectMirador/mirador/issues/1926, so I'm catching up on that. After skimming, it seems some questions may have been answered there. I think the question of providing subscriptions to state changes is a separate issue from this one here. In particular I think the issue of more complex transformations brought up in the ticket above is better handled by having multiple reducers listen for the same event. This is the advantage of redux in my mind; as many different parts of the state can respond to a very specific event as necessary. Each reducer can handle the same event in a different way. Indeed this was part of the rationale for choosing redux: one component changes the state, and different derived state can be calculated easily in a predictable way.

charbugs:
@aeschylus Thank you for the detailed questions. I will answer your question about the current problems of the state managment approach in this comment and will respond to your thoughts about specific vs. basic actions types and logic in reducers vs. logic in combined actions in another comment.

There are two reasons why I suggest the reducer pattern: a) enhance long term maintainability and b) create a solid basis for an advanced plugin system. Here is why I think the code lacks maintainability:

__1. We heavily violate the "don't repeat youself" principle__

The current reducers and actions are doing more or less the same operations. In this early stage of the project there are about 2.000 lines of code that can actually be expressed in about 500 lines via automation. The reason why we are currently forced to write similar actions and reducers again and again is that each of them do their work just a little bit differently. Would we realize what they have in common and build an abstraction than we could save time and error prone code.


__2. There are no constistent state patterns yet__

Apart from the possibility of automation, the state managment code lacks consistency. Here are some points for that:

* We sometimes use very basic action types like `UPDATE_CONFIG`, sometimes we use more specific ones like `TOGGLE_WINDOW_SIDE_BAR`.
* Sometimes we apply thunks to handle complex operations (combined actions), sometimes we accomplish it by let different reducers react to the same action type.
* Sometimes the conditional logic for state updates is handled in reducers, sometimes in thunks.
* Some of the reducers have default props, some have not. Sometimes the default props are treated by actions, sometimes within the reducers.
* There are inconsistencies in naming too. Sometimes we use the term `payload` to refer to an object that should be added to state, sometimes we use `options`, sometimes we use a more specific name like `window`. Sometimes `payload` is an argument to an action, sometimes it's what the action creates within the function.

It may be that we can deal with this inconsistencies because we are working on this code since weeks day by day. But I think people that are new the project have a hard time to reason about the code. And probably we have too, after a longer break.


__3. There is no modularization__

For my understanding, a complex software consists of mutliple modules that have a flexible API but hide their internals from the outer world. This way, a developer who works on module A that uses module B don't have to touch the internals of module B, but rather can simply rely on the API of module B.

This is not the case with the current state implementation. Every time when we work on a feature that needs to write to the state we have to change the reducers, actions an the respective tests. And by that we're ending up whit the inconsistencies from point 2.

Imagine that the code base grows so that part time contributer cannot overview the entire code base. If there is no modularization with clear APIs, the contributer may be forced to touch the internals of modules that he/she not fully unerstands. This is very error prone. 

__4. With reducers reacting on the same action types the execution path is hard to follow__

As far as I can see we currently only have one case where an action type is handled by different reducers. So, this problem might not be obvious for now. But imagine you have an action that is handled by three different reducers. To reason about what this action actually does you to find the respective reducers and  to read what they do. And as I said above the do it each a little different.

In contrast, if your reducers are handling only basic actions with a predictable behaviour (create, update, delete), and these action are combined in a single thunk than it's very easy to follow the exection path in one place. The point is that you see the basic actions and you know what they do, you don't have to figure out how the respective reducers handle this actions.

aeschylus:
Thanks @charbugs, I will make time to respond to more of this soon, but for now I want to thank you especially for the list in your section 2 above. I think these inconsistencies are necessary to address soon in some group pairing work. In particular I think the use of payload should be standardised. I have been using payload as an object key in all actions, as I thought we had agreed on this during our kickoff week, but I'm not finding it in the wiki.

As for thunks, I haven't been taking issue with the current use of them in all my reviews, but I have talked with @jkeck and @sdellis about how to handle actions that need to affect different parts of the state, with some places pointed out to refactor. I didn't realise it was getting that widespread, but we should remove these non-asynchronous thunks. Thunks are a tool for managing asynchronous processes, not calling actions from other actions for no reason. I think using sagas will make this distinction clearer, but for now, we need to avoid ever calling actions from other actions. In asynchronous actions, the actual "thunk" portion isn't actually calling another action (making a state transformation right there), it's sending a request outside of the entire application, and that resolution then triggers a synchronously-applied action at a later time. I've ticketed this here: https://github.com/ProjectMirador/mirador/issues/2049

PR #2043, opened 2019-03-04
Add simple annotation list to window sidebar / companion window

### LINKED ISSUE #2014: Display annotations in a list in the window side panel
Potential fixture: https://iiif.harvardartmuseums.org/manifests/object/299843

screen shot 2019-03-01 at 12 11 15 pm

https://xd.adobe.com/view/f2440f29-045b-4d8d-7b61-0d5c8ece8104-8475/screen/dc313c49-498d-4a64-a406-fedf73d3484b/Side-panel-open-Annotations-available-288r

jkeck:
Closes #2014 

initial-annotations
Manifests in this gif:
- http://storiiies.cogapp.com/holbein/manifest.json
- https://iiif.harvardartmuseums.org/manifests/object/299843


src/state/selectors/index.js
```diff
@@ -96,6 +98,49 @@ export function getSelectedCanvas(state, windowId) {
...
+/**
+ * @param {Array} annotations
+ * @return {Array} [{ id: 'abc123', content: 'Annotation Content' }, ...]
+ */
+export function getIdAndContentOfAnnotations(annotations) {
+  return annotations.map((annotation, i) => ({
+    id: (annotation['@id'] || `annotation-${i}`),
+    content: flatten(new Array(annotation.resource)).map(r => r.chars).join(' '),
+  }));
+}
```
jkeck: I wasn't sure if these nested functions was the way to go.  It almost feels like I want to extract a class similar to what ManifestoCanvas does, but kept it simple for now.  Happy to work on that, split that out into a separate issue, or forget about it.

src/containers/WindowSideBarAnnotationsPanel.js
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * mapStateToProps - to hook up connect
+ * @memberof WindowSideBarAnnotationsPanel
+ * @private
+ */
+const mapStateToProps = (state, { windowId }) => ({
+  annotations: getIdAndContentOfAnnotations(
+    getAnnotationResourcesByMotivation(
+      getSelectedCanvasAnnotations(state, getSelectedCanvas(state, windowId).id),
+      ['oa:commenting', 'sc:painting'],
```
jkeck: Not sure if these are the proper motivations to begin with.  The design has a drop-down that will allow us to switch between particular motivations (which would then be a prop here), but that's for future work.

src/state/selectors/index.js
jkeck: Also, this is pretty much tied to the V2 (Open Annotation) format.  Should I be doing anything to either `try`/`catch` somewhere or explicitly filter for only Open Annotations (based on `@context`).?

christopher-johnson: not clear to me why `canvasId` is being used as the state reference for the annotations.  Passing it as a parameter to the fetch action is somewhat questionable, since it is not used by the fetch action, and is essentially concatenated into the API response by the reducer.  An annotation is self-contained and already contains the target to which it references.  Filtering stored annotations by target would be a more extensible approach, since it is theoretically possible for an annotation to refer to anything,

mejackreed: @christopher-johnson please create a spin off issue for this and link any fixture data that may be relevant or helpful for an implementer.

Issue #2038, opened 2019-03-04
Outcome of "Add resource" is hidden by the form


jvine:
Depending on the side of the viewport and length of the resource list, the Add resource panel may cover the placeholder and resolution (success or failure) of the Add, so the user has no visible feedback.

ggeisler:
If we close the Add resource panel close after the users selects "Add," would this eliminate the problem?

The downside is if the user intends to add several resources consecutively they'd have to reselect the Add resource button each time. But it feels cleaner to me for the panel to close after selecting "Add" and I think that would eliminate the potential problem.

jvine:
I think there's something more complicated going on, actually - the list as a whole doesn't scroll if the viewport isn't long enough to display all of it. So if that is fixed, it might be reasonable to scroll the list to the position where the new item was added. Or let the Add panel push the list up rather than overlay it.

ggeisler:
@jvine @jkeck I think #2048 fixed this? Or is there something else that prevents us from closing this issue?

jvine:
I agree this issue is fixed, and #2126 covers the behaviour change mentioned in the comments above.

PR #2048, opened 2019-03-05
Add a 100% height and auto overflow to workspace add div so that it …

### LINKED ISSUE #1929: Load window add resource panel doesn't work well w/ a long list
Once the list of manifests available is as long as the page, the add resource panel behaves oddly.  It appears that Material-UI hides this panel beneath the viewport, and when trying to scroll to view manifests added beneath the viewport, the add resource panel doesn't stay affixed to the bottom of the viewport.

add-item-error

jkeck:
…can be scrollable (while maintaining the button/panels absolute positioning).

Closes #1929 

workspace-add-after

### DIFF
--- a/__tests__/integration/mirador/basic.test.js
+++ b/__tests__/integration/mirador/basic.test.js
@@ -11,6 +11,7 @@ describe('Basic end to end Mirador', () => {
     await expect(page).toClick('.mirador-add-resource-button');
     await expect(page).toFill('#manifestURL', 'http://localhost:5000/api/sn904cj3439');
     await expect(page).toClick('#fetchBtn');
+    await expect(page).toClick('button[aria-label="Close Menu"]'); // Close menu so new item is visible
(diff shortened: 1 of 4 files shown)


__tests__/integration/mirador/basic.test.js
```diff
@@ -11,6 +11,7 @@ describe('Basic end to end Mirador', () => {
     await expect(page).toClick('.mirador-add-resource-button');
     await expect(page).toFill('#manifestURL', 'http://localhost:5000/api/sn904cj3439');
     await expect(page).toClick('#fetchBtn');
+    await expect(page).toClick('button[aria-label="Close Menu"]'); // Close menu so new item is visible
```
jkeck: I believe this will be solved when dealing w/ #2038, but for now, it seemed reasonable to just close the menu so the item is visible in the test.

PR #2062, opened 2019-03-06
remove vestigial named-asset-importer

mejackreed:
I can't find our usage of this babel plugin or what the use case is for it. Can we remove it or somehow figure out what this is for?

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,6 @@
     "babel-eslint": "10.0.1",
     "babel-jest": "^24.1.0",
     "babel-loader": "^8.0.4",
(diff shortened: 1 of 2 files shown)


christopher-johnson:
this PR explains what this does https://github.com/facebook/create-react-app/pull/3718
It looks to be currently non-functional because "[@svgr/webpack](https://www.npmjs.com/package/@svgr/webpack)" is not in devDependencies.  It is useful if ppl had a need to import svgs as ReactComponents, but calling it vestigial currently is accurate.

aeschylus:
I also have no idea. Doesn't show up in blame.

Issue #1999, opened 2019-02-28
Create separate header/body sections for sidebar & companion windows


jvine:
Spinoff from #1961 

Panels in sidebar and companion windows need separate header and body section.

**Header**

1. [x] remains static at the top of the panel, so the unique content of the panel scrolls underneath it
2. [x] can have a background colour different from the panel
3. [x] contains the panel title ("About", "Index")
4. [x] contains panel controls (popout, ~drag~, close)
  a. [x] popout and close are right-aligned
  b. ~drag is left-aligned~
5. [ ] ~in some panels (e.g. Index, Search, Annotations) the header section may contain additional controls, such as a dropdown or search box. Ideally, the additional controls will appear inline with the title, or below the title, as space allows~

**Body**
1. [x] panel content scrolls **within this section** - i.e. the scrollbar does not extend into the header area.

Any margins or padding will be associated with the content sections within the panel body

<img width="976" alt="screen shot 2019-02-28 at 3 45 09 pm" src="https://user-images.githubusercontent.com/6945691/53606867-38b15600-3b71-11e9-83e5-5717b0568407.png">

PR #2064, opened 2019-03-06
Update companion window header styles

cbeer:
Part of #1999 

screen shot 2019-03-06 at 08 36 24

### DIFF
--- a/src/components/CompanionWindow.js
+++ b/src/components/CompanionWindow.js
@@ -30,8 +30,9 @@ export class CompanionWindow extends Component {
           order: position === 'left' ? -1 : null,
         }}
         square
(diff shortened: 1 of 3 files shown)


cbeer:
I was hoping/assuming when the header styles are applied, magic will happen and things'll look right.

ggeisler:
> I was hoping/assuming when the header styles are applied, magic will happen and things'll look right.

I think this will be true, since the intended font size is likely to be smaller.

jvine:
Font size will be much smaller; we may use a shorter heading; we may need to add a nowrap/overflow behaviour.

It looks like there is still a shadow to the right of the sidebar panel.

PR #2073, opened 2019-03-06
Add canvas labels to thumbnail navigation

### LINKED ISSUE #1932: ThumbnailNavigation styling, labels, and visual design
Our current implementation of ThumbnailNavigation is missing the styles in the mocks.

Needed things:
 - [ ] ~~Effective padding / margins for both single and book view~~
 - [x] Labels added
 - [x] Container areas should work w/ theming

Thumbnails navigation
- [x] use `<nav />` element, add `aria-label="Thumbnail carousel"` (#2101)

jkeck:
Part of #1932

thumbnail-buttons

### DIFF
--- a/src/containers/ThumbnailNavigation.js
+++ b/src/containers/ThumbnailNavigation.js
@@ -1,6 +1,7 @@
 import { compose } from 'redux';
 import { connect } from 'react-redux';
 import { withTranslation } from 'react-i18next';
+import { withStyles } from '@material-ui/core/styles';
 import miradorWithPlugins from '../lib/miradorWithPlugins';
 import CanvasGroupings from '../lib/CanvasGroupings';
 import * as actions from '../state/actions';
@@ -25,9 +26,23 @@ const mapDispatchToProps = {
   setCanvas: actions.setCanvas,
 };
 
+const styles = {
+  canvasLabel: {
+    background: 'linear-gradient(to top, rgba(0,0,0,0.7) 0%, rgba(0,0,0,0.3) 70%, rgba(0,0,0,0) 100%)',
+    textAlign: 'left',
+  },
+  title: {
+    color: 'white',
+  },
+  thumbnailButton: {
+    padding: '0',
+  },
+};
+
 const enhance = compose(
   withTranslation(),
   connect(mapStateToProps, mapDispatchToProps),
+  withStyles(styles),
   miradorWithPlugins,
   // further HOC go here
 );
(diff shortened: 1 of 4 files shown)


src/containers/ThumbnailNavigation.js
```diff
@@ -24,8 +25,22 @@ const mapDispatchToProps = {
   setCanvas: actions.setCanvas,
 };
 
+const styles = {
+  titleBar: {
+    background: 'linear-gradient(to top, rgba(0,0,0,0.7) 0%, rgba(0,0,0,0.3) 70%, rgba(0,0,0,0) 100%)',
+    textAlign: 'left',
+  },
+  title: {
+    color: 'white',
```
jkeck: Is there somewhere I should be getting this color from instead?

jvine:
So on further exploration... it looks like this is rendering invalid HTML. A `<ul>` element can only have `<li>` elements as direct children, and this has a bunch of `<div>`s in between.

jkeck:
I was curious why I wasn't getting any errors from React (which has thrown errors about similar HTML structural issues).

I'm not sure what we can do if we want to use the library (`react-virtualized`) that we're using to handle lazily loading the canvas thumbs.

Would the addition of the button inside the div to give a clickable element be an acceptable improvement?  Otherwise, I can kill the list/button changes all-together and simply add the canvas label as a caption in the `GridListTileBar`.

jkeck:
I've popped on a commit to remove the list but keep the button (now that we're wrapped in a `<nav />` as of #2070 being 🚢ed).  We can squash that down to one commit if it's 👍

src/containers/ThumbnailNavigation.js
cbeer: `theme.palette.common.white`? there's also `theme.palette.getContrastText()`, which I don't understand but sounds promising.

mejackreed:
This looks really great! I did notice one thing where in BookView the second thumb seems to be missing:

screen shot 2019-03-07 at 7 25 12 am

src/containers/ThumbnailNavigation.js
jkeck: I tried this out, but whenever I use `theme` into my styles, everything goes haywire when I switch themes.  Maybe somebody can help me look at this today in the office (or if anybody know of some addl. step I need to do).

cbeer: ✋; sounds good

jkeck:
Closed by #2088

PR #2074, opened 2019-03-06
Pin nodejs version

cbeer:


camillevilla:
^-- Note that this is pinned until jest fixes compatibility with nodejs 11.11

PR #2076, opened 2019-03-07
Remove positive lookbehind regex as it is only supported in chrome.

jkeck:


src/lib/AnnotationResource.js
```diff
@@ -37,9 +37,9 @@ export default class AnnotationResource {
     const { on } = this.resource;
     switch (typeof on) {
       case 'string':
-        return on.match(/(?<=xywh=).*$/)[0].split(',');
+        return on.match(/#xywh=(.*)$/)[1].split(',');
       case 'object':
-        return on.selector.value.match(/(?<=xywh=).*$/)[0].split(',');
+        return on.selector.value.match(/#xywh=(.*)$/)[1].split(',');
```
jkeck: The tests pass, but not 100% sure if there are other cases I'm missing that the lookbehind was getting us.

cbeer: I don't think we need the `#`; at least the previous implementation didn't care.

jkeck: Yeah, looks like while it is recommended, is not required.

Issue #1853, opened 2019-02-12
Window sidebar open/close behaviour


jvine:
Sidebar currently has just one option implemented, using an icon button to open the metadata panel. Icon button does not indicate selected status when its associated panel is opened.

Intended behaviour for the window sidebar is similar to a [tab bar](https://material-components.github.io/material-components-web-catalog/#/component/tabs). (Not clear if MUI's tab bar component allows for vertical tabs, though.)

- [ ] ~1. When the sidebar menu is first opened, the info panel is open be default.~ (**updated March 5 2019**; split to #2059 )

<img width="400" alt="no tab selected" src="https://user-images.githubusercontent.com/5402927/53828898-be4b5200-3f33-11e9-81ff-c43a5d4141e1.png">

- [x] 2. When the user clicks an icon, the associated panel opens, and the icon is in "active" state.

<img width="400" alt="active tab icon" src="https://user-images.githubusercontent.com/6945691/52676220-4a8eca00-2ede-11e9-9d82-cadb0a1ca2cb.png">

- [x] 3. The icon is **not** a toggle: re-clicking an active icon does not hide or close the panel. The open panel has a "hide" toggle: 

<img width="400" alt="hide panel widget" src="https://user-images.githubusercontent.com/6945691/52676372-c8eb6c00-2ede-11e9-8be8-a4dd5cec6b39.png">

- [x] 4. Clicking that toggle hides the panel, but does not change the active state of the icon. User can click the toggle again to un-hide the panel, which remains in exactly the same state as the user left it: scrolled down, filter applied, options selected, etc.

<img width="400" alt="un-hide panel widget" src="https://user-images.githubusercontent.com/6945691/52676471-1bc52380-2edf-11e9-8986-d77d74adf8c7.png">

(Note that the toggle was not shown in this state in previous designs; that was an error.)

- [x] 5. While one icon/tab is selected, user can hover over other icons without changing the active state of the selected icon. 

<img width="400" alt="icon hover state" src="https://user-images.githubusercontent.com/6945691/52676236-54b0c880-2ede-11e9-8044-471a9c3e00c3.png">

- [x] While the panel is visible, clicking:
- the active icon does _not_ hide or close it
- any other icon replaces its content
- the "hide" toggle hides it

- [x] While the panel is hidden, clicking: 
- the "hide" toggle shows the panel in its last open state
- the active icon shows the panel in its last open state
- any other icon opens the panel to that content

[...]

PR #2087, opened 2019-03-07
Convert WindowSideBarButtons to use Material-UI Tabs

cbeer:
Part of #1853

### DIFF
--- a/src/components/WindowSideBarButtons.js
+++ b/src/components/WindowSideBarButtons.js
@@ -1,7 +1,8 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import Badge from '@material-ui/core/Badge';
-import IconButton from '@material-ui/core/IconButton';
+import Tabs from '@material-ui/core/Tabs';
+import Tab from '@material-ui/core/Tab';
 import InfoIcon from '@material-ui/icons/InfoSharp';
 import AnnotationIcon from '@material-ui/icons/CommentSharp';
 import CanvasIndexIcon from './icons/CanvasIndexIcon';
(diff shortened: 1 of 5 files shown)


src/components/WindowSideBarButtons.js
```diff
@@ -39,49 +46,59 @@ export class WindowSideBarButtons extends Component {
...
+          classes={{ root: classes.tab, selected: classes.tabSelected }}
           aria-label={
             this.sideBarPanelCurrentlySelected('canvas_navigation')
               ? t('closeCanvasNavigationCompanionWindow')
               : t('openCanvasNavigationCompanionWindow')
           }
-          onClick={() => (this.toggleCompanionWindow('canvas_navigation'))}
-        >
-          <CanvasIndexIcon
-            color={this.sideBarPanelCurrentlySelected('canvas_navigation') ? 'secondary' : 'inherit'}
```
camillevilla: Nice! We're able to remove this bit of logic because `Tab` has its own set of classes that define CSS for selected icons. This is set below in the `tabSelected` class in the WIndowSideBarButtons container.

src/containers/WindowSideBarButtons.js
```diff
@@ -41,9 +42,27 @@ const mapStateToProps = (state, { windowId }) => ({
...
+  },
+  tabsIndicator: {
+    display: 'none',
+  },
+  tab: {
+    minWidth: 'auto',
+    borderRight: '4px solid transparent',
+  },
+  tabSelected: {
+    backgroundColor: theme.palette.secondary.light,
```
camillevilla: Let's revisit these colors in a more holistic palette review with @jvine and @ggeisler.

PR #2088, opened 2019-03-07
Add the canvas label in a GridListTileBar in the ThumbnailNavigation …

jkeck:
…component.

Part of #1932

new-labels

### DIFF
--- a/src/containers/ThumbnailNavigation.js
+++ b/src/containers/ThumbnailNavigation.js
@@ -1,6 +1,7 @@
 import { compose } from 'redux';
 import { connect } from 'react-redux';
 import { withTranslation } from 'react-i18next';
+import { withStyles } from '@material-ui/core/styles';
 import miradorWithPlugins from '../lib/miradorWithPlugins';
 import CanvasGroupings from '../lib/CanvasGroupings';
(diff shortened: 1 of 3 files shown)


src/containers/ThumbnailNavigation.js
```diff
@@ -25,7 +26,20 @@ const mapDispatchToProps = {
   setCanvas: actions.setCanvas,
 };
 
+/**
+ * Styles for withStyles HOC
+ */
+const styles = {
+  root: {
+    background: 'linear-gradient(to top, rgba(0,0,0,0.7) 0%, rgba(0,0,0,0.3) 70%, rgba(0,0,0,0) 100%)',
```
aeschylus: Not a blocker in any way, but this makes me think we should figure out how we decide to style things; there are so many options! In components? As theme variables? In the stylesheet?

aeschylus: (Love the gradient by the way)

jkeck: Totally agree, I'm still not 100% comfortable w/ deciding about where styles should go (but that may be a lot because css in js is still pretty foreign to me).

PR #2089, opened 2019-03-07
Adds the ability for the ThumbnailNavigation to be displayed on the

### LINKED ISSUE #1983: BookView thumbnail strip - right companion window
Related #1791

mejackreed:
right side.

Fixes #1983 

#2090 spun off to better calculate layout positioning. This will also have implications with some of the work @jkeck has been doing.

thumbnav-right

### DIFF
--- a/src/components/ThumbnailNavigation.js
+++ b/src/components/ThumbnailNavigation.js
@@ -20,6 +20,7 @@ export class ThumbnailNavigation extends Component {
     super(props);
 
     this.cellRenderer = this.cellRenderer.bind(this);
+    this.calculateScaledHeight = this.calculateScaledHeight.bind(this);
     this.calculateScaledWidth = this.calculateScaledWidth.bind(this);
     this.gridRef = React.createRef();
   }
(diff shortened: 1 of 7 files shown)


src/components/ThumbnailNavigation.js
```diff
@@ -110,14 +121,39 @@ export class ThumbnailNavigation extends Component {
...
+
+  /**
+   * calculateScaledHeight - calculates the scaled height of a row for a Grid
+   * in a simple case, a row == canvas. In a book view, a group (or two)
+   * canvases. When in the "bottom" position, this value is static.
+   */
+  calculateScaledHeight(options) {
+    const { config, canvasGroupings, position } = this.props;
+    switch (position) {
+      case 'right':
```
mejackreed: I realize there is some repetitive case statements in may of these methods. I'm open to ideas about how to remove that or improve this. The idea here was to be pretty explicit so a reviewer/developer could understand what is happening.

src/styles/index.scss
```diff
@@ -81,10 +81,6 @@
     margin-left: $window-sidebar-width;
   }
 
-  &-companion-bottom {
```
mejackreed: This isn't really a thing and just kind of left over

src/components/ThumbnailNavigation.js
```diff
@@ -52,12 +53,22 @@ export class ThumbnailNavigation extends Component {
...
-      columnIndex, key, style,
+      columnIndex, key, style, rowIndex,
     } = options;
     const {
-      classes, window, setCanvas, config, canvasGroupings, t,
+      classes, window, setCanvas, config, canvasGroupings, t, position,
     } = this.props;
-    const currentGroupings = canvasGroupings.groupings()[columnIndex];
+    let currentIndex = null;
+    switch (position) {
```
aeschylus: Could do this with a ternary, since I think there will only ever be 2 options.

`currentIndex = (position === 'right') ? rowIndex : columnIndex`

mejackreed: Sure thing. I've made the requested change.

PR #2091, opened 2019-03-07
Remove close button from WindowList component …

### LINKED ISSUE #2061: Remove "close" button from "List open windows" menu
Behaviour of this menu is a bit different than anticipated, and the Close button is no longer appropriate. Not needed, and not accessible by keyboard.

I think it's also appropriate to add `role="presentation"` to the `Open windows` list item, and take it permanently out of the tab order.

jkeck:
…(it is a menu dialog and will close when navigation away)

Closes #2061 

## Before
window-list-before

## After
window-list-after

### DIFF
--- a/__tests__/src/components/WindowList.test.js
+++ b/__tests__/src/components/WindowList.test.js
@@ -84,4 +84,30 @@ describe('WindowList', () => {
       ).toBe(true);
     });
   });
+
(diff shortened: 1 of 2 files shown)


jkeck:
@jvine does the `After` gif look right above?

Issue #2049, opened 2019-03-05
Refactor thunks so multiple actions are never dispatched from action creators


aeschylus:
This came up in discussion here: https://github.com/ProjectMirador/mirador/pull/2035#issuecomment-469722475, and I think a few other places. I want to create a specific thread for it here.

Thunks are a tool for managing asynchrony. There should never be multiple synchronous actions dispatched from a thunk. I think this is mostly happening in the companion window code, but I know @sdellis was working with a similar pattern in the maximise window feature branch. The preferred pattern is to have any reducers that are interested in this action receive and respond to that action. One action, one transformation assertion.

![alt text](https://cdn-images-1.medium.com/max/2600/1*7qtRmuWoMmFyhpnyxoS3MA.png)

Eliminating thunks with sagas may discourage this pattern in general, as it will be impossible to dispatch an action from an action creator, but I don't think we're quite ready for that yet. For now, let's find and refactor instances of actions creators dispatching actions, and record the strategy in our contributor docs.

charbugs:
> Thunks are a tool for managing asynchrony.

This is only half the story. You CAN use thunks to managing asychrony and apparently a lot of people do so. But as the [docs](https://github.com/reduxjs/redux-thunk) say:

> Thunks are the recommended middleware for basic Redux side effects logic, including complex synchronous logic that needs access to the store, and simple async logic like AJAX requests.

That is, if you don't use async calls in thunks they work synchronous as every other action does.

> There should never be multiple synchronous actions dispatched from a thunk.

Can you say way?

>  The preferred pattern is to have any reducers that are interested in this action receive and respond to that action. One action, one transformation assertion.

From what I know so far, it is simply not possible to avoid thunks (or a similar technic) if you deal with complex action logic. Take for example the `closeWindow` action. It includes:

* Getting the companion window IDs from window.
* Delete all companion windows.
* Delete the window.

To accomplish that you need to have a place where you have access to both the `dispatch` and the `state`. Let's look at some places that come into question:

__Reducers:__ Doesn't work! Reducers have no access to the store at all. So the `windowReducer` can not dispatch delete actions to the `companionWindowReducer`.

__React-Redux:__ Doesn't work too! The `mapStateToProps` function only receives the `state` while the `mapDispatchToProps` function only gets the `dispatch` passed.

[...]

PR #2100, opened 2019-03-08
Add a toggle to show/hide the left companion window area

### LINKED ISSUE #1853: Window sidebar open/close behaviour

cbeer:
Fixes #1853

### DIFF
--- a/src/state/actions/companionWindow.js
+++ b/src/state/actions/companionWindow.js
@@ -1,6 +1,6 @@
 import uuid from 'uuid/v4';
 import ActionTypes from './action-types';
-import { updateWindow, setWindowSideBarPanel } from './window';
+import { updateWindow, setCompanionAreaOpen, setWindowSideBarPanel } from './window';
 
 const defaultProps = {
   content: null,
@@ -72,6 +72,8 @@ export function popOutCompanionWindow(windowId, panelType, position) {
 
     if (position === 'left') {
       dispatch(setWindowSideBarPanel(windowId, panelType));
+
+      dispatch(setCompanionAreaOpen(windowId, true));
     }
 
     const action = dispatch(addCompanionWindow({ content: panelType, position }));
--- a/__tests__/src/actions/companionWindow.test.js
+++ b/__tests__/src/actions/companionWindow.test.js
@@ -144,6 +144,12 @@ describe('companionWindow actions', () => {
         windowId: 'abc123',
         panelType: 'info',
       });
+
+      expect(mockDispatch).toHaveBeenNthCalledWith(2, {
+        type: ActionTypes.UPDATE_WINDOW,
+        id: 'abc123',
+        payload: { companionAreaOpen: true },
+      });
     });
   });
 });
--- a/__tests__/src/actions/window.test.js
+++ b/__tests__/src/actions/window.test.js
@@ -66,6 +66,19 @@ describe('window actions', () => {
     });
   });
 
+  describe('setCompanionAreaOpen', () => {
+    it('returns the appropriate action type', () => {
+      const id = 'abc123';
+      const expectedAction = {
+        type: ActionTypes.UPDATE_WINDOW,
+        id,
+        payload: { companionAreaOpen: true },
+      };
+      expect(actions.setCompanionAreaOpen(id, true)).toEqual(expectedAction);
+    });
+  });
+
+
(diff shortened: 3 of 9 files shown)


src/state/actions/companionWindow.js
```diff
@@ -72,6 +72,8 @@ export function popOutCompanionWindow(windowId, panelType, position) {
 
     if (position === 'left') {
       dispatch(setWindowSideBarPanel(windowId, panelType));
+
```
aeschylus: I know this is still not fully resolved, but I believe the point on using multiple dispatches for state management has reached agreement here https://github.com/ProjectMirador/mirador/issues/2049#issuecomment-470148920

We want making actionCreators that create actions, not cause side effects. Redux is meant to operate on inversion of control; this pattern of multiple dispatches reverses the inversion of control and makes us essentially use function calls to update state, meaning that we need to know all necessary state update logic in the actionCreators.

The [redux docs](https://redux.js.org/recipes/structuring-reducers/beyond-combinereducers#sharing-data-between-slice-reducers) give several patterns for "updating state across slices/branches". The first is to refactor the state. The second pattern, which I think makes the most sense, is to add more data to the action that is created that will allow multiple reducer listeners to have what they need to make the transformation. We can still use `getState()` in action creators to make this data available, but never dispatch more than one action that does not have the same name as the actionCreator. An example of this is given here https://github.com/ProjectMirador/mirador/issues/2049#issuecomment-469951071

The exception to this last rule is for async functions, whose main bodies never dispatch any actions, but only initiate requests that themselves have meaningfully-named actionCreator side-effects. See: https://github.com/ProjectMirador/mirador/pull/2035#issuecomment-469816402

cbeer: We spent a good hour yesterday thinking about how to not do this; it seems to me, until we come to an agreement about the new approach, this is the pattern we've been using and can easily be refactored later.

jkeck: Yeah, totally agree that the multiple dispatch is sub-optimal.  But as @cbeer mentions, at this point, it's all encapsulated in this one place since we're already doing multiple dispatches in this action.

Once the pattern that we'll use for this type of thing emerges, refactoring this piece of code to use that new pattern should be relatively easy (as opposed to us trying to figure that pattern now).  I think we should certainly discourage any new actions being added/modified to include addl. dispatches.

mejackreed: @aeschylus you bring up some going points thanks for linking those tickets. Until the larger question is resolved can we move this work forward?

The action creator in question is already dispatching additional actions, adding one more doesn't holistically change the already existing pattern happening. I'm sure the team would be open to addressing these questions at a macro level, in which this work can be refactored, no?

aeschylus: Will merge. Thanks for considering guys.

PR #2107, opened 2019-03-08
Address menu icons and position

jkeck:
Closes of #1936 

## Tasks
- [x] Address location of menus
- [x] ~~Address default focus in menus~~ _(Spinning off into new issue #2115)_
  -  This is intentionally done by the MUI library, and I _believe_ it is for keyboard a11y.  It appears that you need to have focus on a visible element w/i the menu in order to be able to use arrow-key navigation.  There is some interesting discussion in mui-org/material-ui#5186 explaining some of the rationale.  (Note, adding the display:none `MenuItem` at the beginning of the list gives the desired behavior visually, but we lose any keyboard arrow navigation).
  - It might be possible to do something with [MenuList composition](https://material-ui.com/demos/menus/#menulist-composition), but that seems like it'll get a bit complicated, and think it may be best to spin that off into a separate PR, as it seems like the other issues addressed here are a good step forward.
- [x] Make dividers non-focusable
- [x] Make the icon that triggers a menu remain "selected"

### Before
before
top-before

### After
after
top-after

### DIFF
--- a/src/containers/WindowTopMenuButton.js
+++ b/src/containers/WindowTopMenuButton.js
@@ -13,6 +13,9 @@ const styles = theme => ({
   ctrlBtn: {
     margin: theme.spacing.unit,
   },
+  ctrlBtnSelected: {
+    backgroundColor: theme.palette.action.selected,
+  },
 });
 
 const enhance = compose(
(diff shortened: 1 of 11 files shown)


src/containers/WindowTopMenuButton.js
```diff
@@ -13,6 +13,9 @@ const styles = theme => ({
   ctrlBtn: {
     margin: theme.spacing.unit,
   },
+  ctrlBtnSelected: {
+    backgroundColor: theme.palette.action.selected,
```
jkeck: CodeClimate is complaining about this being duplicated.  I _could_ extract this into an abstract mirador menu icon component, but not sure if it's really worth it at this point.

PR #2116, opened 2019-03-08
Add a simple i18n lint that prints warnings during our build

jkeck:
Hoping this can help us keep on-top of our i18n hygiene.

example


scripts/i18n-lint.js
```diff
@@ -0,0 +1,73 @@
+const glob = require('glob'); // eslint-disable-line import/no-extraneous-dependencies
+const fs = require('fs');
+const chalk = require('chalk'); // eslint-disable-line import/no-extraneous-dependencies
```
jkeck: I wasn't sure if this, and the `glob` require were legit.  These are listed in our `package.json` as `devDependencies` (which this seems to be to me) but I may be misunderstanding the purpose of the different types.  Happy to add these as explicit dependencies if that's the way to go.

mejackreed: I think `devDependency` is the right thing here

aeschylus:
This will be really nice to have going forward.

PR #2118, opened 2019-03-08
Add toggle for moving companion windows to the bottom position

cbeer:
Part of #1754

### DIFF
--- a/src/components/CompanionArea.js
+++ b/src/components/CompanionArea.js
@@ -9,6 +9,15 @@ import ns from '../config/css-ns';
 /** */
 export class CompanionArea extends Component {
   /** */
+  areaLayoutClass() {
+    const {
+      classes, position,
+    } = this.props;
+
+    return position === 'bottom' ? classes.horizontal : null;
+  }
+
+  /** */
   render() {
     const {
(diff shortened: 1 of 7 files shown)


src/components/CompanionArea.js
```diff
@@ -16,7 +16,7 @@ export class CompanionArea extends Component {
     } = this.props;
 
     return (
-      <div className={classes.root} style={{ minHeight: 0, display: 'flex' }}>
+      <div className={[classes.root, position === 'bottom' ? classes.horizontal : null].join(' ')} style={{ minHeight: 0, display: 'flex' }}>
```
aeschylus: I realise this has kind of been lost in the fray of classes and styles getting moved around between containers and components and parents (oh my!), but a pattern I've seen that handles this nicely is doing all of the classes string composition in a component function, similar to having a function for `elementStyle(some_props)`. I'm not sure what existing cruft you ran into here, so feel free to say "no" on refactoring this. But would it be possible to abstract this somewhat (or perhaps ticket this so we know it exists)?

cbeer: Sure, I can pull out the new logic to a method.

aeschylus: Thanks. This is a bit more compact.

PR #2121, opened 2019-03-08
 Allow multiple companion windows in companion areas

### LINKED ISSUE #1900: Ability to support multiple companion windows in a given position/plane
Right companion window should stretch to the bottom the window
- can be split vertically into multiple companion windows

Bottom companion window should reach the edge of the window viewer
- can be split vertically into multiple companion windows

Something roughly like this (ignoring the shadow):

screen shot 2019-02-22 at 8 02 20 am

With the thumbnail navigation enabled:
screen shot 2019-02-22 at 10 37 22 am

cbeer:
Fixes #1900

### DIFF
--- a/src/state/reducers/windows.js
+++ b/src/state/reducers/windows.js
@@ -96,6 +96,31 @@ export const windowsReducer = (state = {}, action) => {
       return setCanvasIndex(state, action.windowId, currentIndex => currentIndex - 1);
     case ActionTypes.SET_CANVAS:
       return setCanvasIndex(state, action.windowId, currentIndex => action.canvasIndex);
+    case ActionTypes.ADD_COMPANION_WINDOW:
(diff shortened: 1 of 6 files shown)


src/state/reducers/windows.js
```diff
@@ -96,6 +96,31 @@ export const windowsReducer = (state = {}, action) => {
       return setCanvasIndex(state, action.windowId, currentIndex => currentIndex - 1);
     case ActionTypes.SET_CANVAS:
       return setCanvasIndex(state, action.windowId, currentIndex => action.canvasIndex);
+    case ActionTypes.ADD_COMPANION_WINDOW:
+      if (action.payload.position === 'left') {
```
camillevilla: 🎟 @aeschylus

PR #2130, opened 2019-03-09
use redux-debounced to debounce updateViewport dispatch

### LINKED ISSUE #2129: UpdateViewport could be debounced by redux
In testing the new minimize and maximize actions, it appears that using lodash to debounce the `onViewportChange` function that dispatches `UpdateViewport` does not work properly if another action is fired before the lodash debounce is invoked. 
 
Peek 2019-03-09 14-02

I ran a few tests with [redux-debounced](https://github.com/ryanseddon/redux-debounced), and it appears to improve this since a redux debounced action will always execute at least once even if another action is fired.

christopher-johnson:
closes #2129

### DIFF
--- a/__tests__/src/actions/canvas.test.js
+++ b/__tests__/src/actions/canvas.test.js
@@ -1,6 +1,8 @@
 import * as actions from '../../../src/state/actions';
 import ActionTypes from '../../../src/state/actions/action-types';
 
+const debounceTime = 300;
+
 describe('canvas actions', () => {
   describe('setCanvas', () => {
     it('sets to a defined canvas', () => {
@@ -17,6 +19,11 @@ describe('canvas actions', () => {
     it('sets viewer state', () => {
       const id = 'abc123';
       const expectedAction = {
+        meta: {
+          debounce: {
+            time: debounceTime,
+          },
+        },
         type: ActionTypes.UPDATE_VIEWPORT,
         windowId: id,
         payload: {
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "react-rnd": "^9.1.1",
     "react-virtualized": "^9.21.0",
     "redux": "4.0.1",
+    "redux-debounced": "^0.5.0",
     "redux-devtools-extension": "^2.13.2",
     "redux-thunk": "^2.3.0",
     "sass-loader": "^7.1.0",
--- a/src/components/OpenSeadragonViewer.js
+++ b/src/components/OpenSeadragonViewer.js
@@ -1,7 +1,6 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import OpenSeadragon from 'openseadragon';
-import debounce from 'lodash/debounce';
 import ns from '../config/css-ns';
 import ZoomControls from '../containers/ZoomControls';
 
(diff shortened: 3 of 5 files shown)


aeschylus:
Thanks @christopher-johnson. A performance improvement is certainly needed here, however, I believe what is needed is [not a debounce, but a throttle](http://demo.nimius.net/debounce_throttle/). There will be other components that need to know the viewport state about as fast as rendering can be done, for example, to implement a mini-map that updates as the user moves the viewport, or the drag behaviour on windows that is intended to pan the workspace programmatically as the user drags their window off the edge of the screen (also not yet implemented).

In the Mirador proof-of-concept repo, I user raf-throttle for this purpose on the events that crowd the store/action stream. Could we perhaps modify this PR to use throttling instead of debouncing?

christopher-johnson:
The main purpose of this PR is to address the issue with a trailing edge debounce applied to a viewer event handler function that fires a dispatch action with that event's value.  If the component is updated (for example from another connected component changing the store) before the debounced caller fires the action, then that event value is lost.  I believe that this behavior can be reproduced on netlify (also as shown in the video with this issue).

If the debounce is managed by redux, however, the last event value is not lost on subscriber state update.  I think that this is what is required here. The dispatch control mechanism should be in the context of redux middleware and not in the component caller method.

Regarding throttling vs. debounce in redux, I attempted to do an "apples to apples" performance analysis (see attached).  These are Chrome devtools profiles, so they should be portable.  The main difference appears to be in the timings metric.  With action throttling, React does a periodic tree reconciliation which does not seem to be as fps performant (less spikey) than with debounce.   When the reconciliation is performed at every throttle interval, fps drops significantly.

[debounced-Profile-20190311T110938.txt](https://github.com/ProjectMirador/mirador/files/2951777/debounced-Profile-20190311T110938.txt)
[throttled-Profile-20190311T110710.txt](https://github.com/ProjectMirador/mirador/files/2951779/throttled-Profile-20190311T110710.txt)

 Evaluating ms level performance is nonetheless somewhat arbitrary.  If you can prove that throttling performs better and will fix this issue, than that could be a alternative.  It is also possible to reduce the debounce time to accommodate fast clickers and other components that will subscribe to the viewport state.  200 ms seems to perform similarly to 300 ms.

mejackreed:
Great discussion and resources on both of these options. After some investigation this morning, I'm perhaps convinced we need both options (at least debouncing at the state level).

Here is the case where it seems like debouncing at the state level is needed:
 - As @christopher-johnson put it, the state can be modified by another action before a debounce or throttle has time to finish. The component can even be destroyed which means that the state update never got through. Debouncing at redux seems to resolve this issue. I don't see a way to do that from within component without doing something in `componentWillUnmount`.

```
ViewportChanged -> Debouncing happening -> ComponentUnmounted -> debouncing never finished
```

aeschylus:
I can see the advantages of rate control being handled at the redux/state level. I want to make sure we leave open the ability for components to be updated in "real-time" (which is to say "user perception time"). For example, other components may need to update as a window drags or is resized, or as the viewport pans, or as a panel resizes, or as an annotation region is being expanded. I would aim for no slower than 45fps, and ideally 60fps for these kinds of interactions.

aeschylus:
Looking more closely at this, I think this is fine for the updateViewport event in particular, but we may need to return to it if the annotation code is refactored. Looks good to merge if we can get a rebase in.

christopher-johnson:
Note: I changed the debounce time to 100ms.  As noted in the code, this type of a constant should probably live in a separate application settings registry.  This will basically capture all input interactions that affect viewport state as quickly as possible.  

A benefit with debounce as redux middleware is that if the input interaction is continuous, (like an `onWheel` zoom), then the action is dispatched when it stops.  This also fixes the current problem of a debounced viewport change event dispatch persisting a bogus viewport zoom level.

I think that this produces an acceptable combination of performance and persistence for the viewport state.

aeschylus:
Thanks for the rebase @christopher-johnson. Looks good to go.

PR #2151, opened 2019-03-12
add missing translations

glefi:
@charbugs Grundsatzfrage: Duzen oder Siezen wir den Anwender? Oder vermeiden wir jedwede Anrede?


aeschylus:
So interesting.

camillevilla:
Is there any indication of which form of address was used in Mirador 2? https://github.com/ProjectMirador/mirador/blob/v2.7.0/locales/de/translation.json 

(I don't speak / read German at all 🙂)

aeschylus:
Good link @camillevilla. Looks like both @morpheus-87 and @jbaiter were contributors to that file, which is using _Sie_ (formal address). In light of that, I'm going to go ahead and merge this, but we can always amend later.

charbugs:
We mostly use the neutral form. If this is not possible or sounds weird we use formal "Sie".

glefi:
I talked to @charbugs and @GrosserRichardUBL and we think we should keep it like in Mirador 2 (formal address). We could provide an informal translation as well later on, if the need comes up.

PR #2158, opened 2019-03-12
TOC stuff

### LINKED ISSUE #2068: Add Table of contents to Index panel
Add Ranges structure to the index panel as a hierarchical list navigation. 

The ToC is an alternative to the canvas thumbs navigation, and should be the **primary** navigation where it exists. Complete feature will also require a dropdown menu to select alternate Index options.

**Mockup**
screen shot 2019-03-06 at 10 58 06 am

**Fixtures**
https://github.com/ProjectMirador/mirador/wiki/M3-Fixtures-&-sample-IIIF-data#table-of-contents-and-sequences

**Implementation**
The sidebar menu on the Material Design sidebar is a nice example of an expanding hierarchical list.
https://material.io/design/foundation-overview/

cbeer:
Fixes #2068 

### Fixtures:
https://dms-data.stanford.edu/data/manifests/Parker/nm203xw8381/manifest.json
https://scta.info/iiif/graciliscommentary/lon/manifest
https://iiif.lib.harvard.edu/manifests/drs:5981093
https://iiif.bodleian.ox.ac.uk/iiif/manifest/390fd0e8-9eae-475d-9564-ed916ab9035c.json

### DIFF
--- a/__tests__/src/components/WindowSideBarCanvasPanel.test.js
+++ b/__tests__/src/components/WindowSideBarCanvasPanel.test.js
@@ -3,6 +3,8 @@ import { shallow } from 'enzyme';
 import List from '@material-ui/core/List';
 import ListItem from '@material-ui/core/ListItem';
 import Typography from '@material-ui/core/Typography';
+import ExpansionPanel from '@material-ui/core/ExpansionPanel';
+import ExpansionPanelSummary from '@material-ui/core/ExpansionPanelSummary';
 import manifesto from 'manifesto.js';
 import { WindowSideBarCanvasPanel } from '../../../src/components/WindowSideBarCanvasPanel';
 import { CanvasThumbnail } from '../../../src/components/CanvasThumbnail';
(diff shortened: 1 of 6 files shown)


jvine:
Clarification:  Ranges are not meant to list the canvases within them, only the range labels. As implemented in Mirador2.

camillevilla:
@jvine I chatted with @cbeer about this a bit. Canvas labels have been removed, but the UI is definitely a rough first implementation. Could you and @ggeisler take a pass at this re: styling and HTML elements? 

4 items displayed with table of contents panels showing

ggeisler:
Typography updates will be in a forthcoming ticket JV is writing up. We plan to work on spacing/margins tomorrow. So there will be suggestions for those soon.

I don't think we want the gray background on the panel elements (aside from the header and select menu).

jvine:
It's more than the styling. 

The structure needs to be hierarchical, with the child elements opening when the parent is selected. Here's a more informative manifest to show a complex range structure:
https://iiif.lib.harvard.edu/manifests/drs:5981093
I think it's the `within` property that creates the hierarchy.

This structure has the same interaction and keyboard access requirements as other nav elements:
- indicate selected state for the currently-displayed item
- tab into the control (focus lands on currently selected item)
- arrow within the list
- enter on a parent item toggles its sublist open/closed _and_ displays the first canvas in that range
- enter on a node displays the first canvas in that range

Compare in Mirador2: the hierarchy, navigation, and state aspects are there; keyboard control is not. 

Note that when I open that Harvard manifest, I get a blank element at the top that collapses the entire list.

Oh! Just realized something else, and this may be a separate ticket. The range title "Table of contents" is what should appear in the dropdown. We can't assume that the manifest's range is effectively a table of contents; it's just a sequence they've defined. If the manifest has more than one sequence (it can), we need to list them all in the dropdown with their manifest-provided titles. @aeschylus can you confirm this, please?

regisrob:
> Oh! Just realized something else, and this may be a separate ticket. The range title "Table of contents" is what should appear in the dropdown. We can't assume that the manifest's range is effectively a table of contents; it's just a sequence they've defined. If the manifest has more than one sequence (it can), we need to list them all in the dropdown with their manifest-provided titles. @aeschylus can you confirm this, please?

Yes, a Manifest Range is not necessarily a table of contents, it could be any kind of structure within an object. And there could be multiple structures within a given Manifest (e.g. a ToC of the textual works in a manuscript, its codicological structure, and an index of the miniatures), so a viewer should be able to handle that by giving users the ability to switch between them.

The relevant parts of the 2.1 spec are:
- https://iiif.io/api/presentation/2.1/#members: "In order to present a table of the different ranges..."
- https://iiif.io/api/presentation/2.1/#viewinghint (see viewingHint "top")

An alternative sequence is something conceptually different, it is just a different ordering of the Canvases.

aeschylus:
https://github.com/ProjectMirador/mirador/blob/release-2.x/js/src/widgets/toc.js#L94-L231

PR #2162, opened 2019-03-13
Add tooltips to non-text icons (that displays the aria label)

### LINKED ISSUE #2106: Add tooltips to icon buttons
Icon buttons need tooltips that display the aria-label for the button. Okay to start with default behaviour & placement. (Depending on the outcome, we may want to specifically position in some areas.)

jkeck:
Closes #2106

Issues w/ this implementation (that I would like to spin off into separate issues if that's :ok_hand:)
* Tooltips do not work in full-screen
* Tooltips for Buttons that either move position on click or trigger menu dialogs require a click to close.

## Tab Icons
tab-icon

## FAB Icon
fab-icon

## Custom Component
mirador-menu-button


src/components/MiradorMenuButton.js
```diff
@@ -0,0 +1,34 @@
...
+*/
+export function MiradorMenuButton(props) {
+  const { 'aria-label': ariaLabel, disabled } = props;
+  const { icon, ...iconButtonProps } = props;
+  return (
+    <Tooltip title={ariaLabel}>
+      {
+        disabled // Wrap IconButton in span if it is disabled
+          ? <span><IconButton {...iconButtonProps}>{icon}</IconButton></span>
+          : <IconButton {...iconButtonProps}>{icon}</IconButton>
```
cbeer: `IconButton` also has a disabled property. Should we be passing that through to the button when it's in the disabled state?

src/components/MiradorMenuButton.js
```diff
@@ -0,0 +1,34 @@
...
+
+
+/**
+ * MiradorMenuButton ~ Wrap the given icon prop in an IconButton and a Tooltip.
+ * This shares the passed in aria-label w/ the Tooltip (as title) and the IconButton
+ * All props besides icon are spread to the IconButton component
+*/
+export function MiradorMenuButton(props) {
+  const { 'aria-label': ariaLabel, disabled } = props;
+  const { icon, ...iconButtonProps } = props;
```
cbeer: I'm curious about the choice of an `icon` property instead of `children`?

src/components/MiradorMenuButton.js
```diff
@@ -0,0 +1,34 @@
...
+  return (
+    <Tooltip title={ariaLabel}>
+      {
+        disabled // Wrap IconButton in span if it is disabled
```
cbeer:
```suggestion
        disabled // Wrap IconButton in span if it is disabled so it can receive mouse events (e.g. show the tooltip)
```

src/components/MiradorMenuButton.js
```diff
@@ -0,0 +1,34 @@
...
+          ? <span><IconButton {...iconButtonProps}>{icon}</IconButton></span>
```
cbeer: I wonder if there's an idiomatic way to conditionally wrap a thing?

```
const Wrapper = disabled ? `span` : `Fragment` 
...
<Wrapper>
  <IconButton ...>
</Wrapper>
```

maybe?

src/components/MiradorMenuButton.js
jkeck: Yeah, I went back and forth on that a bit.  I would be a happy to make it required `children` instead of an explicit `icon` prop.  I was poking around some of the MUI code and saw that they have an `icon` prop in some cases (e.g. [`Tab`](https://material-ui.com/api/tab/)) (although they also have children sometimes).

src/components/MiradorMenuButton.js
jkeck: Sure 🤷‍♂️

src/components/MiradorMenuButton.js
cbeer: I think the thing that'd push me to `children` is because this seems like an `IconButton` wrapper, and it uses `children`. 🤷‍♂️

src/components/MiradorMenuButton.js
jkeck: Doesn't seem to work, unfort.  I'm going to just wrap in a span (and comment about why) unconditionally.

PR #2165, opened 2019-03-13
2034 optimize sidebar structure for keyboard support

glefi:
After trying to to encapsulate the methods for the keyboard navigation, I've come to the conclusion, that the tab navigation is way too special, to re-use the implementation for other navigation controls. I ended up with a container for event handlers and a big handler clutter in the using component, so I reverted the changes.

### DIFF
--- a/src/components/WindowSideBarButtons.js
+++ b/src/components/WindowSideBarButtons.js
@@ -35,7 +35,8 @@ export class WindowSideBarButtons extends Component {
   componentDidMount() {
     // eslint-disable-next-line react/no-find-dom-node
     this.tabs = Array.from(ReactDOM.findDOMNode(this.tabRef).getElementsByTagName('button'));
-    this.tabBar = this.tabs[0].parent;
+    console.log(this.tabs[0]);
+    this.tabBar = this.tabs[0].parentElement;
     this.selectTab(this.tabs[0]);
   }
 
--- a/src/components/WindowSideBarButtons.js
+++ b/src/components/WindowSideBarButtons.js
@@ -13,14 +13,22 @@ import CanvasIndexIcon from './icons/CanvasIndexIcon';
  *
  */
 export class WindowSideBarButtons extends Component {
+  /**
+   * selects the given tab
+   * @param {*} tab the tab to activate
+   */
+  static activateTab(tab) {
+    tab.removeAttribute('tabindex');
+    tab.setAttribute('aria-selected', 'true');
+  }
+
   /** */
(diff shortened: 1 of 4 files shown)


src/components/WindowSideBarButtons.js
```diff
@@ -16,23 +17,148 @@ export class WindowSideBarButtons extends Component {
...
     this.handleChange = this.handleChange.bind(this);
+    this.handleKeyUp = this.handleKeyUp.bind(this);
+    this.handleKeyDown = this.handleKeyDown.bind(this);
+    this.handleTabRef = this.handleTabRef.bind(this);
+
+    this.keys = {
+      end: 35,
+      home: 36,
+      up: 38,
+      down: 40,
```
jkeck: MUI uses `event.key` and compares that to things like `Home` and `ArrowUp`.  This may be more reliable that the key codes? https://github.com/mui-org/material-ui/blob/7e1a0f1e1574d36188d333d8a91288e6f516db09/packages/material-ui/src/MenuList/MenuList.js#L43-L82

glefi: Yes, you're right, `keyCode` is marked as deprecated. But unfortunately 'key' doesn't work for Safari. I've switched to 'key' and 'which' (this combination should work in all modern browsers)

jkeck:
Just poking around how MUI handles testing this, I wonder if we could use something along of their strategy: https://github.com/mui-org/material-ui/blob/d1a7d76db79b2f1f03068c7c3bc0c9f28a7c5a64/packages/material-ui/test/integration/MenuList.test.js#L60-L109

src/components/WindowSideBarButtons.js
```diff
@@ -12,78 +12,201 @@ import CanvasIndexIcon from './icons/CanvasIndexIcon';
...
+    };
   }
 
-  /** */
+  /**
+   *
+   */
+  componentDidMount() {
+    this.tabs = Array.from(document.querySelectorAll(`#${this.containerId} button[role=tab]`));
```
jkeck: I'm wondering if some of our testing woes may be coming from the use of `document.querySelectorAll` in the `componentDidMount` here.  I'm going to poke at if refs could be at all useful.

glefi:
fixed with https://github.com/ProjectMirador/mirador/pull/2284

camillevilla:
ah ok! so #2284 actually merged changes into this branch, @glefi. We'll want someone to review this branch so that all of it gets merged into master.

PR #2170, opened 2019-03-13
Try harder to get manifest thumbnails

### LINKED ISSUE #2139: Use first canvas thumbnail to represent resource in load window
In the current load window, it looks like if the manifest for a resource does not include a `thumbnail` property, we don't show anything in the thumbnail area:

Screen Shot 2019-03-11 at 1 54 17 PM

In issue #1966 I suggest adding a placeholder for unavailable thumbnails, but if it is possible it would be better to use a thumbnail version of the first canvas of the resource before resorting to a placeholder image. So to populate the thumbnail area of the resource in the load window, the order of preference would be:

1. Use the `thumbnail` value in the manifest
2. Use a thumbnail version of the first canvas image of the resource
3. Use a placeholder thumbnail as described in #1966

cbeer:
Fixes #2139


ggeisler:
Nice!

PR #2172, opened 2019-03-13
Fix NestedMenu keyboard nav (for Language Settings)

### LINKED ISSUE #2022: Submenus should be accessible from keyboard
Submenus in the "Appbar" menu list should be accessible from the keyboard, but are not.

screen shot 2019-03-01 at 4 09 32 pm

jkeck:
Fixes #2022 

NestedMenuArrowNav

### DIFF
--- a/__tests__/src/components/NestedMenu.test.js
+++ b/__tests__/src/components/NestedMenu.test.js
@@ -38,12 +38,6 @@ describe('NestedMenu', () => {
     expect(wrapper.find('WithStyles(Typography)').children().text()).toEqual('GivenLabel');
   });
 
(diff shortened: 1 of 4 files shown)


cbeer:
Is it weird the submenu options are now peers of the other options in the menu? 🤷‍♂️

jkeck:
AFAICT it's the only way to get the Menu keyboard navigation (and styling) to work w/o rolling our own.   The MUI menu keyboard navigation assumes sibling navigable elements only.

Discussed this w/ @jvine and @ggeisler a bit yesterday and I _think_ they said it was "okay".  Not sure if there is something aria wise we can/should do here (that could be spun off into another issue or grouped w/ other aria work).

Issue #1693, opened 2019-01-24
Mirador API Requirements (Developer UX Specification)


aeschylus:
This ticket is intended to be a gathering place for many other issues, but I want to ensure we have properly captured all the desired API functionality in detail. In practice, these areas all need to be realised as integration tests, and imply we get our distribution channels handled. My hope is that this becomes the seed of our comprehensive API documentation, and the One True Thread for debate on the external API surface area. I will edit this comment as discussion continues below.

## Prior Art
These are some previous attempts to organise feedback on the shape of the Mirador API surface area and capabilities.
[wiki page on embedding/import contexts]()
[plugin types]()
[Known Consuming Contexts](https://github.com/ProjectMirador/research-and-demos/wiki/Known-Mirador-Consumer-Contexts)
[Known Mirador Plugins](https://github.com/ProjectMirador/research-and-demos/wiki/Known-Mirador-Plugins)

## Deployment Contexts
Mirador is a javascript _library_ to be included in several different kinds of software contexts. These slides from the Mirador Community Update in Washington D.C. last year sums up the various contexts where Mirador needs to run _as a library_.
![image](https://user-images.githubusercontent.com/145874/51778809-93f6b100-20b8-11e9-9d5a-c8e6688f8f22.png)

![image](https://user-images.githubusercontent.com/145874/51778739-3eba9f80-20b8-11e9-8d22-73b4796d8f92.png)

These contexts and their attendant use cases will determine the shape of its API surface area, which may not be identical for each context. Our job is to make it a pleasure to use (as a developer) from all four of these different perspectives. 

## Packaging/Publication
In general, we want to publish the main package with NPM
#### Developer Experience
The developer types `npm install mirador` from their project folder.  
In their project Code, they can now type
``` Javascript
import Mirador from `mirador`;

const viewer = Mirador({id:'container'});
```
Where the DOM node with an id `container` could come from a jade template, for example.
 
How should this work in `rails`?
How should this work within a react application?

``` Javascript
import Mirador from `mirador`;

```
### CDN
It should be possible to grab the total Mirador bundle, or any of its sub-bundles, from a CDN for creating and sharing examples in sandbox websites, static pages, presentations, or local use.
**Example**  
``` html
<script crossorigin src="https://unpkg.com/mirador/umd/mirador.min.js"></script>
```
## Configuration
Configuration should be human-readable and easily writable. An average webmaster should be able to easily tweak the configuration after reading a few examples.
## Theming
[...]

PR #2173, opened 2019-03-13
Enable top level es module usage and components

### LINKED ISSUE #2083: Export the top-level react component for parent react applications
see: https://github.com/ProjectMirador/mirador/issues/1693

`import miradorComponent from 'mirador'` (or something similar), should provide the top level react component.

mejackreed:
Fixes #2083 

This PR does a few things with a bunch of lessons learned that I hope I can share.

### Commits (by order added)
1. Moves import of `index.scss` into the top level component rather than a generic JavaScript 
We need to do this, so that `index.scss` is shown when `./src/containers/App.js` is required outside the context of implementing through `MiradorViewer`.
2. We can use `@babel/cli` to build a version of Mirador that essentially just replicates our `src` directory structure. This enables installers of the Mirador npm package to be able to import individual `containers`, `components`, and other classes directly. We still have the container/state dependencies that need to happen, but this enables some more flexibility.
   #### Example
   ```javascript
    import MiradorApp from 'mirador/dist/es/src/containers/App';
    // or
    import { CanvasThumbnail } from 'mirador/dist/es/src/components/CanvasThumbnail'
    ```
3. Moves `react` and `react-dom` dependencies to `devDependencies` and `peerDependencies`. This is needed as React >16.8 has a check for "multiple" instances of react being required in a page. This will cause issues for users wanting to install the Mirador package that also are using `react` or `react-dom`. For more reading about this, checkout https://reactjs.org/warnings/invalid-hook-call-warning.html#duplicate-react
4. Since we are already using `locales` in `src/i18n.js`, lets just move them in the same directory. The only reason for this is to simplify the `build:es` task

### How do we know this works?
Good question. Without changes to the project structure I'm not sure we have a way to test this. Open to other's ideas here. But you can check it out for yourself.

1. Run `npm pack` which will create a local tarbal packed up build `mirador-3.0.0-alpha.0.tgz`
1. Use `create-react-app` to quickly create a new React application
    ```sh
    $ npx create-react-app my-app 
    $ cd my-app
    $ yarn install ../../mirador/mirador-3.0.0-alpha.0.tgz
    $ yarn start
    ```
   You will also need to update (in your cra my-app) App.js https://gist.github.com/mejackreed/5462dc8f9319ac0d806f8bc511d177af

Screen Shot 2019-03-13 at 3 58 10 PM

cra!


package.json
```diff
@@ -48,13 +50,11 @@
     "node-sass": "^4.9.2",
     "openseadragon": "^2.4.0",
     "prop-types": "^15.6.2",
-    "react": "^16.8.3",
-    "react-dom": "^16.8.3",
     "react-full-screen": "^0.2.4",
     "react-i18next": "^10.2.0",
     "react-mosaic-component": "^2.1.0",
     "react-placeholder": "^3.0.1",
-    "react-redux": "^6.0.0",
```
cbeer: Can you say more about why we're downgrading and pinning `react-redux`?

mejackreed: Yikes, thanks for catching that. I've removed this change to move back to what it was previously.

aeschylus:
Merging, but this has been good at uncovering a bunch more design considerations that are being spun off.

christopher-johnson:
FYI: There are positioning issues with the mirador default scss in the HSP implementation.  

I need to continue to manage the scss outside of mirador because of this.  Ideally, there would be no named css classes and everything could be done by jss overrides and themes.  You  want to avoid setting `top`, `left` and making assumptions about `padding` and `position: absolute` in the scss where ever possible.  These hard coded values could be made into theme variables instead.

For reference, you can see the extent of the positioning style overrides here: https://sleepy-curie-a284a7.netlify.com/view

sdellis:
@christopher-johnson we definitely don't want any css collisions, which is why we are namespacing. I agree that we should not be setting styles on HTML elements, but I don't see any evidence of this. As long as `position:absolute` elements are wrapped in `position:relative` elements, there should not be a problem. Can you provide more details on the conflict you're seeing in HSP, with a screenshot?

christopher-johnson:
example:
```
  &-workspace-maximized-window {
    left: 0;
    position: absolute;
    top: 0;
    z-index: 10000;
  }
```
this will put the maximized window at the top (over all parent elements).  removing  `position: absolute` works (also for `&-viewer`).  
Screenshot (3)

Not sure about your assertion about wrapping absolute with relative, but [this ](https://stackoverflow.com/questions/42057043/how-to-fully-wrap-an-absolute-position-div-with-a-relative-position-div) is informative. The app bar is assigned a min-height (it makes no difference for constraining the child AFAICT).

The example provided in this issue does not really wrap Mirador.  React implementations may want Mirador as a deeply nested child component....

charbugs:
So, with this approach:

* Mirador will not be bundled and minimized at all.
* It's up to the adopters to provide React and bundle Mirador in their apps.
* It's up to the adopters to write an initialization similar to `MiradowViewer.js`
* Adopters can import arbitrary modules from Mirador.

This is a very flexible and low-level API and it's a big benefit for use cases like the HSP. But does it really match the requirements from #2083 / #2197 / https://codesandbox.io/s/wkyw556xp7? I mean, their should be a higher level import of Mirador that avoids writing the intialization code.

```javascript
import Mirador from 'mirador';

function MyApp(props) {
  return (
   <div>
    <MyStuff />
    <Mirador config={config} plugins={plugins} />
    <MyOtherStuff>
  </div>
  );
}
```
I think that the API introduced by this PR is what @aeschylus calls "Composable Component API" in #1693 and that the above example is a different thing.

mejackreed:
We recognize this does not address all of the cases and needs expressed in #1693 and hopefully use this as a starting point to build towards that.

PR #2182, opened 2019-03-14
Initial pass at a more keyboard accessible WindowTopMenu

### LINKED ISSUE #2029: Window top bar menu options are not accessible via keyboard
screen shot 2019-03-01 at 5 11 52 pm

Menu dropdown should also be constrained to its parent window and the layout should be closer to the original mockup as seen here:

screen shot 2019-03-01 at 5 25 01 pm

jkeck:
Closes #2029 

WindowTopBar-keyboard-nav

Discussed this a bit w/ @jvine and @ggeisler and it sounds like a flat `<li>` that just has ArrowUp ArrowDown support to begin with (better than the current which is not keyboard accessible at all).

I've also added support to have the first selectable item in the list selected by default (which is necessary to be able to navigate via keyboard arrows w/o needing to tab into the menu), however; I find it basically impossible to test (Enzyme says that's browser behavior and you should use a browser env to test, and while puppeteer claims that `document.activeElement` is supported, I could not get anything but `undefined` in the tests (even while validating it outside of `HEADLESS` mode).

I'm still not able to get the `ListSubheader` to be not focusable at all (which I _think_ is the desired behavior).  I have some thoughts on how to handle that, but think it would probably be best to spin that off into a separate issue (and maybe do the same thing in `WindowList`)

### DIFF
--- a/src/components/WindowThumbnailSettings.js
+++ b/src/components/WindowThumbnailSettings.js
@@ -1,8 +1,7 @@
 import React, { Component } from 'react';
 import FormControlLabel from '@material-ui/core/FormControlLabel';
-import Radio from '@material-ui/core/Radio';
-import RadioGroup from '@material-ui/core/RadioGroup';
-import Typography from '@material-ui/core/Typography';
+import ListSubheader from '@material-ui/core/ListSubheader';
+import MenuItem from '@material-ui/core/MenuItem';
 import ThumbnailsOffIcon from '@material-ui/icons/CropDinSharp';
 import PropTypes from 'prop-types';
 import ThumbnailNavigationBottomIcon from './icons/ThumbnailNavigationBottomIcon';
@@ -22,10 +21,10 @@ export class WindowThumbnailSettings extends Component {
   /**
    * @private
    */
-  handleChange(event) {
+  handleChange(value) {
     const { windowId, setWindowThumbnailPosition } = this.props;
 
-    setWindowThumbnailPosition(windowId, event.target.value);
+    setWindowThumbnailPosition(windowId, value);
   }
 
   /**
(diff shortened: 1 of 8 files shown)


src/components/WindowThumbnailSettings.js
```diff
@@ -34,42 +34,65 @@ export class WindowThumbnailSettings extends Component {
...
       <>
-        <Typography>{t('thumbnails')}</Typography>
-        <RadioGroup aria-label={t('position')} name="position" value={thumbnailNavigationPosition} onChange={this.handleChange} row>
+        <ListSubheader
+          role="presentation"
+          selected={false}
+          disabled
+          tabIndex="-1"
+        >
+          <Typography>{t('thumbnails')}</Typography>
```
jkeck: These `Typography` elements cause the headers to be quite small (as noticeable in the gif).  Not sure if there is a `variant` I should be using here or what.

src/components/WindowThumbnailSettings.js
```diff
@@ -34,42 +34,60 @@ export class WindowThumbnailSettings extends Component {
...
     return (
       <>
-        <Typography>{t('thumbnails')}</Typography>
-        <RadioGroup aria-label={t('position')} name="position" value={thumbnailNavigationPosition} onChange={this.handleChange} row>
+        <ListSubheader role="presentation" tabIndex="-1">
+          <Typography>{t('thumbnails')}</Typography>
+        </ListSubheader>
+
+        <MenuItem className={classes.MenuItem} onClick={() => { this.handleChange('off'); handleClose(); }}>
           <FormControlLabel
```
jkeck: `<FormControlLabel>` seems like an odd choice here, but it gives us the combination of the Icon and Label styled nicely together.  I guess it's not technically any less valid than it was before (given that it never hooked up a proper `for` attribute or anything like that).

If this is unacceptable, I can try to roll something custom for this (or if anybody knows an existing MUI component that would be better, I'm all ears).

jkeck:
If anybody has tips on how to test the `componentDidMount` using a ref, I would be happy to hear them.  I struggled w/ this a lot this afternoon, and didn't find anything that didn't feel like I was writing addl. code just to serve the tests.

PR #2189, opened 2019-03-14
Make the ThumbnailNavigation just another companion window area

cbeer:
🤷‍♂️


cbeer:
I'm not sure I see the differences you are alluding to; can you be more specific?

Issue #2187, opened 2019-03-14
enhanced plugin system: 'replace' type


camillevilla:
Follows work from #2098 and uses the pattern established in the proof-of-concept repo. 
Pre-requisites are #2184 and #2186

mejackreed:
Resolved in #2199

Issue #2186, opened 2019-03-14
enhanced plugin system: 'remove' type


camillevilla:
Follows from #2098 and uses work in the proof-of-concept repo.

mejackreed:
Resolved in #2199

Issue #2185, opened 2019-03-14
Develop initial pattern for handling multiple "add" plugin types


camillevilla:
Follows work from #2098 
#2184 is a pre-requisite. 

This work modifies the pattern established in in the proof-of-concept repo so that multiple plugins can be added. 

This work is NOT negotiating the hierarchy of components (e.g. what happens when I try to add something to a component that has been removed?).

Issue #2184, opened 2019-03-14
initial implementation of new plugin pattern into M3 - "add"


camillevilla:
This work follows from #2098. This work would implement the pattern demonstrated in the @charbugs proof-of-concept repo. For this ticket let's focus exclusively on the `add` action.

mejackreed:
Resolved in #2199

PR #2192, opened 2019-03-14
#2184 initial implementation of enhanced plugin pattern

charbugs:
* Implements the add (#2184), delete (#2186 ), replace (#2187) and wrap (#2191) plugin types. 
* Support for multiple add plugins (#2185) is not implemented here.
* No component uses `withPlugin` so far and therefore their are no integration tests.

Next step: write integration tests.


charbugs:
> Function _withPlugins has a Cognitive Complexity of 10 (exceeds 5 allowed). Consider refactoring.

I firmly reject that!

src/lib/miradorViewer.js
```diff
@@ -0,0 +1,68 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { Provider } from 'react-redux';
+import deepmerge from 'deepmerge';
+
+/** Default Mirador instantiation */
+export default function ({
```
mejackreed: Did `MiradorViewer` need to be refactored to accomplish this? Can you talk a little bit about that need if so?

charbugs: * The processing of plugins works different from the current implementation. a) The `plugins` parameter is an array of objects, b) plugins will be stored within the application. Because this is a little simpler in some regards, the `processPlugins` function becomes unnecessary.

* The `MiradorViewer` pretends to be a class, but it doesn't return a instance, but rather a plain object. So, it's actually a function and the refactoring reflects that.

* For the sake of testing I decoupled the default function from it's dependencies to other mirador modules. The init function now has the task to clue things together. I think that's the very notion of an init function.

mejackreed: I'm trying to understand the necessity of this refactor to accomplish the plugin goals. I recognize that:

> The MiradorViewer pretends to be a class, but it doesn't return a instance, but rather a plain object. So, it's actually a function and the refactoring reflects that.

Yet does it this refactor need to happen extensively to accomplish your plugin goals?

I'm trying to understand the smallest possible changes needed here to facilitate our plugin system. So that later in the future we can see the gradual change.

> The plugins parameter is an array of objects

Could `MiradorViewer` just take that argument also?

charbugs: Yes, it's possible with less changes. Will change it arccodingly and treat the refactoring in a separate step.

PR #2193, opened 2019-03-14
Cascade window positions in elastic mode; fixes #2122

### LINKED ISSUE #2122: Cascade new windows in elastic mode
Currently when switching to elastic mode with existing windows, or adding a new window while in elastic mode, all windows are opened in the same workspace location. This behavior is understandable, since we do want windows in open in a noticeable location, but the drawback is that:

- When moving from mosaic to elastic all windows are stacked directly on top of each other, with only the top window visible. A user might not understand that there is more than one window present in the workspace.

- When adding an additional window in elastic mode, if an existing window hasn't been moved or resized from the default, it will be completely obscured by the window that has just been added.

Ideally, we should cascade new windows in elastic mode similar to the mockup below, by checking if a window exists in the default location, for example, before adding the new one, and adjusting the new window placement accordingly.

Screen Shot 2019-03-08 at 4 23 23 PM

cbeer:
@ggeisler - I'm wondering if this implementation of cascade is sufficient; it doesn't check if an existing window is already at a position, but sets the new window position to a unique position based on the number of windows.

The only drawback I can think of is that windows won't consistently appear in the same position, but that'd also be true if someone panned the elastic workspace anyway.


cbeer:
> Not sure if this should be part of a different ticket, but when you add a new window I think it should get the top bar focus indicator

Let's make that a different ticket, preferably one that lists all the cases where windows should be "focused"

ggeisler:
Sure, sounds good. As you said this earlier, now that you have the basic focus indicator working, we can play with the app and figure out all the cases when a window should have the indicator and when it should not.

PR #2196, opened 2019-03-14
Pan the elastic workspace to the focused window when selected in the window list

### LINKED ISSUE #2124: Window list selection should move focus to selected window
- [x] We've proposed renaming the workspace menu option "List all open windows" to "Jump to window" (though not sure that has been ticketed; if not, it could done as part of this ticket). When the user selects a window title from the "Jump to window" list, that window should gain focus (#2095).  

- [x] Also, in elastic mode, the workspace should also be panned to make the selected window visible in the workspace (especially important for situations where the selected window is outside of the visible workspace viewport).

- [x] the workspace should panned to make the selected window _centered_

cbeer:
This is a part of #2124. It's not yet centered (because we'd have to do a bunch more math to know how big the viewport is, how big the windows are, etc), but the new windows appear in a consistent location at least.

### DIFF
--- a/__tests__/src/actions/window.test.js
+++ b/__tests__/src/actions/window.test.js
@@ -2,6 +2,55 @@ import * as actions from '../../../src/state/actions';
 import ActionTypes from '../../../src/state/actions/action-types';
 
 describe('window actions', () => {
+  describe('focusWindow', () => {
+    it('should return correct action object with pan=true', () => {
+      const expectedAction = {
+        type: ActionTypes.FOCUS_WINDOW,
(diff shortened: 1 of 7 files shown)


aeschylus:
I really like this solution as opposed to trying to manage animation parameters as part of the app. Using CSS will also make the feeling and behaviour of the panning interaction theme-able down the line. The way it sets up the action is also very flexible.

I have a bunch of the coordinate plumbing code written from some branches I created while spiking on this workspace. I will submit it as a separate PR and we can ticket the centring behaviour from there.

PR #2203, opened 2019-03-15
Refactor MiradorViewer. Closes #2200

### LINKED ISSUE #2200: Refactor MiradorViewer.js
* `MiradorViewer` is defined as a class, but it works like a function as it doesn't return a class instance but a plain object.

* It's tightly coupled to a bunch of modules which makes testing difficult.

* The respective unit test is actually an integration test.

charbugs:


mejackreed:
Can we hold off on this right now? I agree that perhaps we should return an instance instead. But there may be other things that we want to do with an instance and provide more API functionality from what is returned.

charbugs:
No problem.

For my understanding, can you say which advantages you see for classes over functions in this case? The only thing I can think of is inheritance.

mejackreed:
> For my understanding, can you say which advantages you see for classes over functions in this case? 

For me it's less about inheritance. I want to leave open the ability to provide a convenience API for umd build users. For instance, if we refactored `MiradorViewer` to return an instance:


```javascript
var miradorInstance = Mirador.viewer({
  id: 'mirador' // id selector where Mirador should be instantiated
});

miradorInstance.addWindow({manifest: "http://foo"})
```

This would use the underlying redux api functionality, but provide a less customized application the ability to tap in and use things in an easy way.

aeschylus:
I agree we need to hold off on this and have tried to link some related issues together in the issue for this PR (#2200).

charbugs:
@mejackreed  Yes, we definitely need this API features. But you don't need classes to accomplish that:

```javascript
function addWindow() { /* ... */ }

function donateToMirador() { /* ... */ }

function MiradorViewer() {
  // ...
  return {
    addWindow,
    donateToMirador,
    whatEver,
  }
}

const mirador = MiradorViewer();
mirador.addWindow();
```

Moreover, imagine that API doesn't live in the MiradorViewer but in a separate module:

```javascript 
import * as api from './api'

function MiradorViewer() {
  // ...
  return {
    ...api,
    whatEver,
  }
}

const mirador = MiradorViewer();
mirador.addWindow();
```

This isn't just a hack to avoid classes, it's the common [factory function pattern](https://medium.freecodecamp.org/class-vs-factory-function-exploring-the-way-forward-73258b6a8d15).

PR #2204, opened 2019-03-15
Elastic minimap

aeschylus:
A prototype for a minimap. The pink box shows the bounding box around all windows. It probably shouldn't be visible in the minimap, and this bounding box dimensions should probably inform the layout at some point, since the whole workspace can be a lot larger than the bounding box. This was originally added to try to get an idea of how to structure the calculation for the transform and scale values for the exposé interaction.

Interested in any feedback, design/interaction-wise. The code is very rough.

minmap


camillevilla:
I don't have much feedback re: the technical approach, but functionally this is a nice improvement on Elastic mode, @aeschylus.

I think it would be important to allow users to toggle the minimap on and off via the Workspace settings UI. I can imagine it would take up a bit of screen real estate on smaller browsers.

PR #2209, opened 2019-03-15
Use react-image to provide placeholder images (in case there is no image url, the image doesn't load, etc)

### LINKED ISSUE #1966: Add placeholders to load window
Part of https://github.com/ProjectMirador/mirador/issues/1963

The load window currently has skeleton screens to represent a manifest resource that is loading and an error message if the resource cannot be loaded. It needs placeholders for thumbnails that fail to load, however:

load-window-failed

If either the resource thumbnail or the provider logo cannot be loaded, the relevant skeleton screen element should be replaced by a similar gray rectangle, but without the animation:

load-window-placeholders

cbeer:
Fixes #1966 

Screen Shot 2019-03-15 at 11 06 12


src/components/ManifestListItem.js
```diff
@@ -119,7 +122,15 @@ export class ManifestListItem extends React.Component {
             </Grid>
 
             <Grid item xs={4} sm={2}>
-              <WindowIcon className={ns('manifest-list-item-logo')} manifestId={manifestId} />
+              <Img
```
mejackreed: Is it possible to use our own `CanvasThumbnail` class also to do something similar?

cbeer: I'm not sure what you mean?

mejackreed: Nope it doesn't thanks!

Issue #2208, opened 2019-03-15
Improve window sidebar open animation


ggeisler:
I know there is work going with the window sidebar in a couple of other tickets, but we wanted to note that while the animation of opening the sidebar was improved in #2125, it is still not ideal. Here's how it currently looks:

![window-open-animation](https://user-images.githubusercontent.com/101482/54450588-0164ac80-470e-11e9-9b3e-e444f6330986.gif)

The problem is how the sidebar nav appears to slide in from the left (which is fine) but the panel just appears in place, so during the interval in which the sidebar nav is sliding in there is a visible gap between the sidebar nav container and the panel. The sidebar panel should not appear to be detached from the nav container; they should appear to slide into place as a single unit.

cbeer:
I think we looked at this when we were replacing drawers. We can trivially get the right animation if we wrap the companion area in a `<Slide>` element from material-ui, but the "hide/show" button that floats outside the window area is problematic.

cbeer:
I've pushed up #2210 demonstrating the problem. If the toggle is inside the slide (so it slides in with the companion area), it is hidden when the area slides out; if the toggle is outside the pane (as it is in #2210), it snaps to the right position instead of sliding with the area. 🤷‍♂️

ggeisler:
I see. For the former behavior (the one not in the PR), what exactly does it mean for the hide/show toggle to be "hidden when the area slides out"? Hidden just during the slide animation, or hidden even after the animation is complete?

If it is only hidden during the animation, that seems like a better compromise than the PR behavior. But if it is hidden after the animation, then yeah, that is clearly not a viable option. 

@jvine in case you want to weigh in on this while I'm out.

cbeer:
Hidden after the animation, because it's part of the thing sliding out.

ggeisler:
@cbeer Okay, @jvine and I discussed and we think the behavior in #2210, while not ideal, is better than the behavior currently in master.

camillevilla:
Closed by #2210... for now.

PR #2210, opened 2019-03-15
Wrap companion areas in a slide so they transition smoothly

cbeer:
Related to #2208


aeschylus:
You probably already know this (hence draft), but the outer sidepanel wrapper isn't sliding with its interior window.

camillevilla:
This needs a rebase, but moving this out of draft per @ggeisler's comment in the issue:
>@cbeer Okay, @jvine and I discussed and we think the behavior in #2210, while not ideal, is better than the behavior currently in master.

camillevilla:
Slide! 
dog on slide

Issue #2214, opened 2019-03-16
Whitespace details (App, main workspace, windows)


jvine:
## Appbar
 
- [x] `WorkspaceControlPanel` `root` change width from 96 to 64
- [x] `WorkspaceControlPanel` `toolbar` change `paddingLeft` and `paddingRight` to 0
- [x] Change left margin of workspace from 100px to 68px 
- [x] set a media query for the FAB:
  - horizontal (mobile) margins: 4px 8px 4px 0;
  - vertical (desktop) margins: 8px 4px 8px 4px

<img width="75" alt="Screen Shot 2019-03-15 at 6 21 13 PM" src="https://user-images.githubusercontent.com/6945691/54468957-32b19c80-474f-11e9-8061-4fc41b712011.png"> <img width="238" alt="Screen Shot 2019-03-15 at 6 21 34 PM" src="https://user-images.githubusercontent.com/6945691/54468955-32b19c80-474f-11e9-9ff3-c356fc0b68d9.png">

## Window top bar

- [x] `windowTopBar` add 4px right padding
- [x] Unlike other buttons in the window top bar, the `WindowTopMenuButton` button has an added `{margin:8px;}`. Removing the margin from this button will make the top bar the correct height and correct the spacing between the buttons.
- [x] `<h2>` element add 4px left padding

<img width="593" alt="Screen Shot 2019-03-15 at 11 57 28 AM" src="https://user-images.githubusercontent.com/6945691/54467660-d9903b80-4743-11e9-811f-62036427d220.png">

## Window sidebar nav

- [x] `WindowSideBar` change width from 55 to 48

<img width="57" alt="Screen Shot 2019-03-15 at 12 13 22 PM" src="https://user-images.githubusercontent.com/6945691/54467666-e01eb300-4743-11e9-8996-9bd4a69c661a.png">

## Companion window 

- [x] `.mirador-companion-window-right` add
  -  `border-right: 1px solid rgba(0,0,0,0.12);`
  - `box-sizing: border-box;`
- [x] `.mirador-companion-window-bottom` add
  - `border-top: 0.5px solid rgba(0,0,0,0.12);`
  - `box-sizing: border-box;`

### Panel header

- [x] Left sidebar
    - Left gutter 16px; right 0px. 
- [x]  right/bottom companion windows
    - Left gutter 0px; right 0px. 
- [x] when the `window-title-controls` section is empty, the right (pop-out / close) button is pushed to the left. It should always be flush against the right border.
[...]

PR #2216, opened 2019-03-17
optimizes and corrects dense variant WindowTopBar and button styling

christopher-johnson:
removes conflicting CSS button class bindings and duplicate default declarations
part of #2214

### DIFF
--- a/src/components/WindowTopBar.js
+++ b/src/components/WindowTopBar.js
@@ -9,7 +9,6 @@ import Toolbar from '@material-ui/core/Toolbar';
 import AppBar from '@material-ui/core/AppBar';
 import classNames from 'classnames';
 import WindowTopMenuButton from '../containers/WindowTopMenuButton';
-import WindowTopBarButtons from '../containers/WindowTopBarButtons';
 import { MiradorMenuButton } from './MiradorMenuButton';
 import ns from '../config/css-ns';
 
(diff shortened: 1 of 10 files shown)


christopher-johnson:
Please note that because the `WorkspaceFullscreenButton` is used also in the `WorkspaceControlPanel`, removing the 8px padding applied from `theme.spacing.unit` affected the alignment of those buttons.  However, this will be required to reduce the width of the control panel to 76px, which is the minimum needed for 16px gutters and a small 40px Fab `WorkspaceAddButton`.  Demo here: https://sleepy-curie-a284a7.netlify.com/view

Screenshot at 2019-03-17 11-27-36

src/components/WindowTopBar.js
```diff
@@ -28,32 +27,32 @@ export class WindowTopBar extends Component {
       maximizeWindow, maximized, minimizeWindow, focused,
     } = this.props;
     return (
-      <AppBar position="relative">
-        <Toolbar disableGutters className={classNames(classes.windowTopBarStyle, focused ? classes.focused : null, ns('window-top-bar'))} variant="dense">
+      <AppBar
+        className={classNames(classes.windowTopBarStyle, focused ? classes.focused : null, ns('window-top-bar'))}
+        color="secondary"
+        position="static"
```
cbeer: I think we explicitly don't want `static` styling here, or the viewer doesn't work well in embedded contexts, right?

src/components/WindowTopBar.js
```diff
@@ -28,32 +27,32 @@ export class WindowTopBar extends Component {
...
-            color="inherit"
             onClick={toggleWindowSideBar}
           >
             <MenuIcon />
           </MiradorMenuButton>
-          <Typography variant="h2" noWrap color="inherit" className={classes.title}>
+          <Typography variant="h2" noWrap className={classes.title}>
             {manifestTitle}
           </Typography>
-          <WindowTopBarButtons windowId={windowId} />
```
cbeer: It seems premature to remove this; I'd anticipate some of the plugin work to go through and update components as appropriate.

src/containers/WorkspaceFullScreenButton.js
```diff
@@ -22,20 +22,8 @@ const mapStateToProps = state => ({
  */
 const mapDispatchToProps = { setWorkspaceFullscreen: actions.setWorkspaceFullscreen };
 
-/**
```
cbeer: It seems like some of these styles were important to keeping the toolbar layout appearing right; it's possible the work in #2225 addressed this though?

src/components/WindowTopBar.js
```diff
@@ -28,32 +27,32 @@ export class WindowTopBar extends Component {
       maximizeWindow, maximized, minimizeWindow, focused,
     } = this.props;
     return (
-      <AppBar position="relative">
-        <Toolbar disableGutters className={classNames(classes.windowTopBarStyle, focused ? classes.focused : null, ns('window-top-bar'))} variant="dense">
```
cbeer: I'm not sure I understand why this PR removes the mirador-namespaces CSS classes; I was under the impression that those classes was our preference to support some downstream theming use cases?

christopher-johnson: Namespaced css classes are bloat.  Prefer providing class references as theme variables instead.  Also. Integration tests need not depend on class selectors with jss.

PR #2224, opened 2019-03-18
Enables annotations to be translatable by CanvasWorld location

### LINKED ISSUE #2133: Display annotations on canvases in BookView
While #2132 implements basic single canvas viewing, there will need some enhancements to display within book view

Enhancements likely needed:
 - [x] ~Enhance OpenSeadragonCanvasOverlay to handle calculations for multiple image sources (maybe by index?)~ not needed did in a different way
- [x] Update selector to handle selection for multiple canvasIds
- ~Determine if WindowSidebarAnnotation needs any design modifications~ Part of #1994

mejackreed:
This fixes #2133 by enabling translation of annotations. Adds a
refactor of world calculations for an OpenSeadragon canvas for
future extensibility.

bookview annos

## Updated 
bookview annos

### DIFF
--- a/src/components/OpenSeadragonViewer.js
+++ b/src/components/OpenSeadragonViewer.js
@@ -133,8 +133,8 @@ export class OpenSeadragonViewer extends Component {
     const { currentCanvases } = this.props;
     const context = this.osdCanvasOverlay.context2d;
     annotations.forEach((annotation) => {
-      const offset = new CanvasWorld(currentCanvases).offsetByCanvas(annotation.target);
       annotation.resources.forEach((resource) => {
+        const offset = new CanvasWorld(currentCanvases).offsetByCanvas(resource.targetId);
         const fragment = resource.fragmentSelector;
         fragment[0] += offset.x;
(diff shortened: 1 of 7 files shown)


src/components/OpenSeadragonViewer.js
```diff
@@ -77,7 +80,7 @@ export class OpenSeadragonViewer extends Component {
         tileSources.map((tileSource, i) => this.addTileSource(tileSource, i)),
       ).then(() => {
         if (tileSources[0]) {
-          this.fitBounds(...this.boundsFromTileSources(), true);
+          this.fitBounds(...new CanvasWorld(currentCanvases).worldBounds(), true);
```
mejackreed: I realize we are recalculating this everytime we need it. It's not an expensive calculation, so it seems fine to do this way. There could be an oppy in the future to not need this, but I don't this its really necessary.

mejackreed:
@camillevilla A follow on commit just added to support work added #2198 . See "Updated" gif above. Also resolves the issue you point out about changing workspace type.

Issue #2221, opened 2019-03-18
Integration test for plugins


charbugs:
Should go in `__tests__/integration/mirador/plugins.html` and `__tests__/integration/mirador/plugins.test.js`

 - [ ] add plugin
 - [ ] wrap plugin
 - [ ] order of plugins
 - [ ] correct connection to store

PR #2229, opened 2019-03-18
add back basic plugin "replace" integration test

mejackreed:
Adds some basic plugin system integration tests for add and replace.

Part of #2221


charbugs:
For larger tests we can place `<script src="https://unpkg.com/babel-standalone@6/babel.min.js"></script>` in the page so we can use JSX in HTML ([see here](https://reactjs.org/docs/add-react-to-a-website.html#quickly-try-jsx)).

PR #2230, opened 2019-03-18
Enable window close and window maximize to be configurable from settings

mejackreed:
This wasn't originally ticketed, but I thought it would be a useful add for our embedded scenario (while I was browsing our ready column)


camillevilla:
Thanks for this @mejackreed. Wondering if we should start looking into documentation options for the config...

PR #2235, opened 2019-03-18
Make the canvas list style selector a "without label" type and style…

### LINKED ISSUE #2181: Update styling of TOC select menu
Given the MUI/React components are not ideal, we're hoping we can improve the styling of the select menu used in the TOC panel. Currently it looks like this:

select-current

It appears that it is using a variant of the select component that includes space for a label? In any case, I think we want to use the variant that is called "Without label" in the [documentation](https://material-ui.com/demos/selects/):

select-example

In addition, I think we're going to have to make some CSS adjustments if we want it to look decent in the panel header context. I don't really know how much flexibility we have in terms of fine-tuning the component-related CSS, or how to be more specific about what we need to update, but this is what we are after (I adjusted the background color and some padding values on various divs to get this):

select-goal

I'm happy to work with a developer on this if it makes sense to discuss what is possible, CSS-wise.

jkeck:
…appropriately.

Closes #2181 

### Before
before

### After
Screen Shot 2019-03-18 at 2 43 02 PM

### DIFF
--- a/src/containers/WindowSideBarCanvasPanel.js
+++ b/src/containers/WindowSideBarCanvasPanel.js
@@ -38,6 +38,14 @@ const styles = theme => ({
     borderBottom: '0.5px solid rgba(0,0,0,0.12)',
     paddingRight: theme.spacing.unit,
   },
+  select: {
+    '&:focus': {
+      backgroundColor: theme.palette.background.paper,
+    },
+  },
+  selectEmpty: {
+    backgroundColor: theme.palette.background.paper,
+  },
 });
 
 const enhance = compose(
--- a/src/components/WindowSideBarCanvasPanel.js
+++ b/src/components/WindowSideBarCanvasPanel.js
@@ -96,13 +96,14 @@ export class WindowSideBarCanvasPanel extends Component {
         id={id}
         windowId={windowId}
         titleControls={(
-          <FormControl variant="filled">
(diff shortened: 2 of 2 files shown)


src/containers/WindowSideBarCanvasPanel.js
```diff
@@ -31,13 +31,32 @@ const mapDispatchToProps = { setCanvas: actions.setCanvas };
...
+      backgroundColor: theme.palette.background.paper,
+      borderTopLeftRadius: theme.shape.borderRadius,
+      borderTopRightRadius: theme.shape.borderRadius,
+    },
+  },
+  selectEmpty: {
+    backgroundColor: theme.palette.background.paper,
+    borderTopLeftRadius: theme.shape.borderRadius,
+    borderTopRightRadius: theme.shape.borderRadius,
+    marginTop: theme.spacing.unit * 2,
```
cbeer: I think we'll also need to do something responsive when the companion window is in the bottom position; right now it ends up looking pretty funny with a huge top margin.

cbeer: I wonder, too, if some of this top margin belongs on the titleControls div in the companion window?

jkeck:
Killed all the niceties I was trying to add, none of it was specified in the ticket.

PR #2247, opened 2019-03-19
Improve thumbnail navigation styling

### LINKED ISSUE #2142: Thumbnail navigation styling
This issue is about the visual styling and theming flexibility of the thumbnail navigation strip.

Currently, a lot of absolute position calculation is causing overlap of elements. Goals are:

1. possibility for user selection of smaller/larger thumbnails (initially designed as a size/border drag handle on the companion window)
2. consistent, potentially customizable/theme-able, margins/sizes for the container, the images, and their borders
3. accommodation for a long-side scrollbar in the container so it doesn't obscure content
4. elimination of the short-side scrollbar
    - `<div class="ReactVirtualized__Grid">` has an inline style of `overflow: auto;` which is allowing scrollbars in both directions. This may be okay if the container height is calculated to allow for all content (see below) but it might be safer to specify `overflow: auto hidden;` for the horizontal nav and `overflow: hidden auto;` for the vertical.

I don't know if the grid component supports this, but it seems like this control would be much more flexible if the positioning and sizing were relative rather than absolute. 

Screen Shot 2019-03-11 at 3 12 29 PM

- Container: 
  - has a theme-able background
  - 2px padding on all sides
  - additional 14 px on the long outside (right or bottom, depending)
- Images have:
  - 2px padding
  - 2px border
    - unselected items have transparent border
    - selected item has secondary-colour border
  - 2px margin

thumb nav horizontal

thumb nav vertical

mejackreed:
Updates ThumbnailNavigation styling and fixes #2142 

Split off #2248 to deal with other browser/OS sizes

updatethumbstyling

### DIFF
--- a/__tests__/src/components/ThumbnailNavigation.test.js
+++ b/__tests__/src/components/ThumbnailNavigation.test.js
@@ -64,7 +64,7 @@ describe('ThumbnailNavigation', () => {
     expect(setCanvas).toHaveBeenCalledWith('foobar', 0);
   });
   it('sets up calculated width based off of height of area and dimensions of canvas', () => {
-    expect(renderedGrid.find('.mirador-thumbnail-nav-container').first().prop('style').width).toEqual(90);
+    expect(renderedGrid.find('.mirador-thumbnail-nav-container').first().prop('style').width).toEqual(95);
   });
   it('renders canvas thumbnails', () => {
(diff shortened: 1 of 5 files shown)


camillevilla:
Great! This moves us in the right direction. Per group discussion, implementing #2269 will give us a lot more re: styling flexibility and accessibility.

Issue #1877, opened 2019-02-13
i18n locale toggle (information panel metadata)


camillevilla:
If multiple languages are provided in the manifest metadata, we need to indicate that to the user and provide a way of toggling between the languages. 

Resources:
- [IIIF v3 spec - Language and Property Values](https://iiif.io/api/presentation/3.0/#43-language-of-property-values)

jvine:
Hm. My reading of this bit in the spec was that the application language setting would capture the user's language preference, and the manifest language would respond to that - i.e. if the user selects French at the app level, and an open manifest has French content, that content would display. But the missing piece is when there is no app-level language option for content that exists in one of the manifests in the workspace. (Implementer provides English, Spanish, French; manifest has Japanese.) Hm.

eroux:
related: https://github.com/ProjectMirador/mirador/issues/1959

camillevilla:
Initial backend implementation was added via #2250. Further design work for this feature is in #2477

PR #2250, opened 2019-03-19
Add a language picker to the info panel when the manifest metadata is multilingual

cbeer:
Add a language picker to the info panel when the manifest metadata is multilingual. Part of #1877.

Screen Shot 2019-03-21 at 09 26 16

(I'm using https://iiif.hab.de/object/mss_272-helmst/manifest.json as the manifest)

----

I don't know that the UI plumbing is right, but it'll provide something to react to.

I also don't know if the backend plumbing, where we reinstantiate the manifesto instance for a language is a great long-term pattern, but it at least kicks the can down the road until we decide on its fate.

### DIFF
--- a/__tests__/src/components/LocalePicker.test.js
+++ b/__tests__/src/components/LocalePicker.test.js
@@ -1,0 +1,54 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import { LocalePicker } from '../../../src/components/LocalePicker';
+
+/**
+ * Helper function to create a shallow wrapper around LanguageSettings
+ */
+function createWrapper(props) {
+  return shallow(
+    <LocalePicker
+      availableLocales={[]}
+      locale={undefined}
+      setLocale={() => {}}
+      {...props}
+    />,
+  );
+}
+
+describe('LocalePicker', () => {
+  let wrapper;
+
+  it('hides the control if there are not locales to switch to', () => {
+    wrapper = createWrapper({ availableLocales: ['en'] });
+
+    expect(wrapper.find('WithStyles(WithFormControlContext(Select))').length).toBe(0);
+  });
+
+  it('renders a select with the current value', () => {
+    wrapper = createWrapper({ availableLocales: ['en', 'de'], locale: 'en' });
+
+    expect(wrapper.find('WithStyles(WithFormControlContext(Select))').length).toBe(1);
+    expect(wrapper.find('WithStyles(WithFormControlContext(Select))').props().value).toBe('en');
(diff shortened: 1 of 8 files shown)


camillevilla:
👏 Nice! Explicitly tagging @jvine and @ggeisler here so they can weigh in on UI approach

jvine:
The UI for this needs some thought - we weren't aware of it as a priority for this sprint, and haven't focused on it at all. Not even prepared to comment whether this is the approach we'd take. (Apart from translating the language codes into language names.)

Is it possible to match the language of the manifest to the language of the UI by default? Once I have selected a language in the UI, I don't want to have to select that language again in the manifest to see labels in my preferred language.

cbeer:
I think defaulting to the user's language is fine, but I thought #1877 was particularly about when the either the user's current language isn't one of the options in the manifest (say, I've selected English in the app, and load a manifest with German + French metadata..), the case where there isn't a full translation of the metadata and I might want to switch between them, etc.

As I noted in my comment, this PR is about getting the groundwork in place for those behaviors, and the UI and backend implementation are almost certainly in flux.

camillevilla:
Ah, ok. So @cbeer is right that I was focused on the case where metadata is provided in a language that isn't provided by the application. 

@jvine I think there needs to be a separate ticket added to the i18n work, which is inferring the metadata language if it's available in the application language. Currently, there isn't any wiring between the info panel and the application language, so this PR provides the foundation for setting up inferred behavior.

🎟 tickets 🎟
- infer metadata language from application language, if available
- translate language codes to language names (We might have rely on some package for this if the language code isn't [one that the implementer is providing](https://github.com/ProjectMirador/mirador/blob/73c7984c5ff4f6e984692d880ad25e059b5d8a03/src/config/settings.js#L125-L128); the translated language names aren't provided by the manifest)

How do those sound?

The backend issues that @cbeer mentions are definitely non-trivial (e.g. some potential incompatibilities with manifesto), so there will be some further review and discussion of that in this PR as well.

camillevilla:
Er alright, I guess @cbeer just beat me to it. 
e0fd4b implements
>infer metadata language from application language, if available

jvine:
And another ticket for the design/interaction?

Yay for inferring the language.

camillevilla:
Per discussion at sprint planning, moving this work forward will involve:
- ripping out language selector UI + merging backend work
- ticketing language selector for separate design work

cbeer:
Instead of ripping out the code, I've just put the UI behind a props flag.

PR #2255, opened 2019-03-20
Add toggle to switch between All and Selected annotations

### LINKED ISSUE #2135: Implement annotation toggling of "All" vs "Selected"
Basic annotations on canvas implemented in #2132 this enhancement adds the ability to view "All" annotations for a canvas (or a CanvasGrouping) or only selected ones (when clicked on in list see #2134)

This could be implemented independently (but coordinated with #2134)

@ggeisler also mentions using `Switch` component from: https://material-ui.com/demos/selection-controls/#switches

Screen Shot 2019-03-11 at 12 21 09 PM

## Design
https://xd.adobe.com/view/f2440f29-045b-4d8d-7b61-0d5c8ece8104-8475/screen/dc313c49-498d-4a64-a406-fedf73d3484b/Side-panel-open-Annotations-available-288r

jkeck:
Closes #2135 

AnnotationToggling


src/state/actions/annotation.js
```diff
@@ -93,3 +93,17 @@ export function deselectAnnotation(windowId, canvasId, annotationId) {
     type: ActionTypes.DESELECT_ANNOTATION, windowId, canvasId, annotationId,
   };
 }
+
+/**
+ * toggleAnnotationDisplay - action creator
+ *
+ * @param  {String} windowId
+ * @param  {String} canvasId
+ * @param  {String} annotationId
```
glefi: 'canvasId' and 'annotationId' are copypasted

src/components/AnnotationSettings.js
```diff
@@ -0,0 +1,38 @@
...
+  render() {
+    const { displayAll, t, toggleAnnotationDisplay } = this.props;
+
+    return (
+      <FormControlLabel
+        control={(
+          <Switch
+            checked={displayAll}
+            onChange={toggleAnnotationDisplay}
+            value={displayAll ? 'all' : 'select'}
```
glefi: I wonder if we should add `disabled={true}` if there is only 1 annotation.

glefi: I wonder if we should add `disabled={true}` if there is only 1 annotation in total.

jkeck: This is a setting that apples to all canvases, so I'm not sure that would be desirable.

src/state/actions/annotation.js
jkeck: Good eye. I've updated the comment.

src/components/AnnotationSettings.js
jkeck: I suppose it could switch between disabled/enabled when paging through canvases, but this wasn't specified anywhere in the designs AFAICT.  @jvine @ggeisler am I missing a requirement about disabling the toggle?

jvine: @ggeisler what do you think? Making the control wink on and off as the user pages through canvases with varying numbers of annotations might be distracting. But if there are no annotations for a canvas, it might be odd to have a control that does nothing visible on this canvas but does affect what you'll see on another canvas.

ggeisler: Yeah, there wasn't a design for this specifically because the annotation mockups were only a first-pass at the general potential features.

My inclination is to go with the disabled-when-appropriate behavior, though I see the potential distraction it might cause when paging through canvases. But maybe that situation won't come up all that often, and maybe when it does won't be as distracting as it seems it might, and it seems better to always have logical behavior for the control.

jkeck: Now disabling for `0` or `1` annotations on a canvas.

PR #2257, opened 2019-03-20
Update how annotations in the panel are styled

### LINKED ISSUE #2264: Change how annotation selection behaves
Currently, annotations are added to a list of selections specific to each canvas.  The intent of the feature is to change the selected annotation from click to click (as more of a navigational construct).

A nice feature may be to allow a canvas to have multiple annotations selected at once (using some modifying keys like shift or command).

#2257 was created and pushed up to begin to think through the implications/behavior of this kind of select (or to decide to abandon all-together for a simple single select).

jkeck:
This PR simply improves the selected annotation styling (and includes a minor function abstraction to cleanup the click event on the component).

Closes #2264

Updates to behavior are now described in #2383 

annotation-style-improvement

### DIFF
--- a/src/components/WindowSideBarAnnotationsPanel.js
+++ b/src/components/WindowSideBarAnnotationsPanel.js
@@ -13,15 +13,35 @@ import ns from '../config/css-ns';
 */
 export class WindowSideBarAnnotationsPanel extends Component {
   /**
+   * constructor -
+   */
+  constructor(props) {
+    super(props);
+
+    this.handleClick = this.handleClick.bind(this);
+  }
+
+  /**
+   * Handle click event of an annotation.
+  */
+  handleClick(event, annotation) {
+    const {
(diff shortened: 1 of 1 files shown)


__tests__/src/reducers/windows.test.js
```diff
@@ -295,6 +295,20 @@ describe('windows reducer', () => {
...
+      };
+      const expectedState = {
+        abc123: { selectedAnnotations: { cId: ['aId'] } },
+      };
+
+      expect(windowsReducer(beforeState, action)).toEqual(expectedState);
+    });
+  });
+
   describe('SELECT_ANNOTATION', () => {
```
jkeck: If we go this route, I'll update `SELECT_ANNOTATION` to be something like `ADD_SELECTED_ANNOTATION`.

jkeck: We did not go this route, so I've kept the action names the same.

PR #2258, opened 2019-03-20
Use reselect for manifest selectors in order to remove manifesto instance from the redux store

cbeer:
I started work on #1877 and it seems like manifesto is deficient in several key ways that will make that work somewhat challenging to do cleanly. In a face-to-face discussion, we decided that it's probably appropriate to at least be cautious in how we use manifesto (perhaps going as far as replacing it). As a first step, it seems helpful to get the manifesto instance out of the redux store.

### DIFF
--- a/src/state/selectors/index.js
+++ b/src/state/selectors/index.js
@@ -14,11 +14,20 @@ export function getManifest(state, { manifestId, windowId }) {
   ];
 }
 
+/** Instantiate a manifesto instance */
 export const getManifestoInstance = createSelector(
   [getManifest],
   manifest => manifest && manifest.json && manifesto.create(manifest.json),
 );
 
+/**
+ * Get the logo for a manifest
+ * @param {object} state
+ * @param {object} props
+ * @param {string} props.manifestId
+ * @param {string} props.windowId
+ * @return {String|null}
+ */
 export const getManifestLogo = createSelector(
   [getManifestoInstance],
   manifest => manifest && manifest.getLogo(),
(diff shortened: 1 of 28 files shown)


src/state/selectors/index.js
```diff
@@ -1,10 +1,37 @@
...
 import ManifestoCanvas from '../../lib/ManifestoCanvas';
 import CanvasGroupings from '../../lib/CanvasGroupings';
 
+/** */
+const uncurry = curriedFn => (...args) => (
+  args.reduce((left, right) => left(right), curriedFn)
+);
+
+/** */
+const createSelectorN = (...selectors) => {
```
charbugs: Can you give some background for this. I know what currying is and I see what the code does but I cannot reason about why it' needed.

cbeer: You're right, it's unnecessary here (for now). I think it starts to become useful when the selectors start to do more work and/or we end up with selectors with slightly different input (like `getDestructuredMetadata` that acts on either a manifest or canvas), but, yes, let's only add them when the time comes.

src/state/selectors/index.js
```diff
@@ -1,72 +1,82 @@
...
-* @param {object} state
-* @param {String} windowId
-* @return {object}
-*/
-export function getWindowManifest(state, windowId) {
-  return state.windows[windowId]
-    && state.windows[windowId].manifestId
-    && state.manifests[state.windows[windowId].manifestId];
+/** Get the relevant manifest information */
+export function getManifest(state, { manifestId, windowId }) {
```
sdellis: This seems to return a manifest by either a `manifestId` or a `windowId`? Does it matter if two windows are using the same manifest?

cbeer: I don't think so. I guess we might not reuse the same manifesto instance for each window, but I'm not sure we care?

PR #2267, opened 2019-03-20
Fix GalleryView selector

cbeer:


camillevilla:
Related: #2258, #2266, #2265

PR #2270, opened 2019-03-20
fixes #2234 by removing span and living with disabled buttons not hav…

### LINKED ISSUE #2234: Redundant titles in span-wrapped buttons are read twice by screenreader
I've been noticing this pattern in the HTML:

```
<span class="" title="Toggle window sidebar">
<button ... aria-label="Toggle window sidebar">
```

Is there a reason for the title attribute on the `<span>` element? It causes VoiceOver to announce "Toggle window sidebar" twice. I know not all AT will behave the same way, but I can't see any reason for the user to be aware of the wrapping element.

sdellis:
…ing tooltips

fixes #2234

### DIFF
--- a/__tests__/src/components/MiradorMenuButton.test.js
+++ b/__tests__/src/components/MiradorMenuButton.test.js
@@ -33,13 +33,7 @@ describe('MiradorMenuButton', () => {
     wrapper = createWrapper();
 
     expect(wrapper.find('WithStyles(Tooltip)').props().title).toEqual('The Label');
-    expect(wrapper.find('WithStyles(Tooltip) span WithStyles(IconButton)').props()['aria-label']).toEqual('The Label');
-  });
-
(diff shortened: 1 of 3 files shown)


cbeer:
I think you'll also want to revert the commit that introduced the `wrapperClassName` (I can dig it up if it isn't obvious).

sdellis:
@cbeer 	I was able to rebase, but when I tried to revert the commit you mentioned, I got lots of errors and wasn't sure how to back out. I am out tomorrow and will revisit this next week with a fresh pull.

sdellis:
I think I did this right. Rebased, tests pass, and `wrapperClassName` commit was reverted. This was the first time I did this type of reversion, so @cbeer please double check that I did it right. 

FYI - Code coverage on GitHub is complaining, but I don't think it's accurate.

PR #2291, opened 2019-03-21
Use jss to style Gallery view so it can be more easily themed via MUI.

### LINKED ISSUE #2251: Gallery view - hook up styling to theme
Split from #2154 #2032. 
There was some discussion re: the similarity of styling between Gallery view items and the thumbnail nav styling implemented in #2142. Options discussed:
- Consolidate styles between the two features via shared classes
- Leave the two classes alone, and hook up GalleryView to the theme the MUI way

jkeck:
Closes #2251 

GalleryCanBeThemed

NowWithTheme

### DIFF
--- a/src/containers/GalleryView.js
+++ b/src/containers/GalleryView.js
@@ -1,5 +1,6 @@
 import { compose } from 'redux';
 import { connect } from 'react-redux';
+import { withStyles } from '@material-ui/core/styles';
 import * as actions from '../state/actions';
 import { withPlugins } from '../extend';
 import { GalleryView } from '../components/GalleryView';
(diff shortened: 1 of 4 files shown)


src/containers/GalleryView.js
```diff
@@ -16,6 +17,43 @@ const mapStateToProps = (state, { window }) => (
...
+      transition: '.1s transform ease-out',
+    },
+    boxSizing: 'border-box',
+    cursor: 'pointer',
+    display: 'inline-block',
+    height: '160px',
+    margin: '10px 5px',
+    maxWidth: '100px',
+    overflow: 'hidden',
+    padding: '5px',
```
jkeck: I'm not sure if the padding/margins here should be using theme variables or not (I'm just moving it from index.scss here).  If anybody thinks that it would make sense to use the spacing vars or anything, let me know.

cbeer: I'd certainly be in favor of using the theme spacing (unless there's a strong reason we need 1px more spacing in gallery view than everywhere else we're adding padding + margins)?

Issue #1985, opened 2019-02-27
Brightness / Contrast plugin


camillevilla:
We want to use this as an opportunity for testing out plugin architecture. 
- Not in projectmirador/mirador repo

jvine:
![plugin menu adjust](https://user-images.githubusercontent.com/6945691/54859865-7cf6c880-4ccf-11e9-95b1-14a7303c5cd7.gif)

- add "Adjust image" option to the window menu, in the section below view/thumbnails
- on select, close the menu and open the "Adjust image" toolbar
- toolbar is horizontal above the mobile breakpoint; vertical below it
   - note that we may want a configuration option to set the default orientation for the toolbar - manuscripts with vertical text will prefer a vertical toolbar that doesn't cut off the first character of each column
- toolbar has its own close button
- icon tooltips appear below or to the left, depending on the orientation of the toolbar
- toolbar pattern is meant be used for other tools & plugins as well - annotations, ruler, etc. 

The illustrated set of icons represents what's implemented in Mirador2:
<img width="1462" alt="Screen Shot 2019-03-22 at 6 40 05 PM" src="https://user-images.githubusercontent.com/6945691/54860071-0c04e000-4cd2-11e9-8e5d-a437f7c80c48.png">

Partial implementation with Brightness/Contrast alone would look like so:
<img width="369" alt="Screen Shot 2019-03-22 at 6 28 45 PM" src="https://user-images.githubusercontent.com/6945691/54859951-67ce6980-4cd0-11e9-8496-3d52f9742c51.png">

Tooltip | Icon (MUI name)
--- | ---
Adjust image (menu option) | tune
Brightness | brightness_5
Contrast | custom (ask @jvine for this one)
Greyscale | tonality (different from mockups)
Invert colors | invert_colors
Rotate right | rotate_right
Rotate left | rotate_left
Flip horizontal | compare_arrows
Flip vertical | compare_arrows rotated
Revert image | refresh
Close toolbar | close

mejackreed:
I started working on this https://github.com/mejackreed/mirador-image-tools however, I think its blocked until we can resolve the "Where does the PluginHook thing happen for all the things?" question.

mejackreed:
Ok so updates.. I've moved the repo here: https://github.com/projectmirador/mirador-image-tools

Will start creating spin off tickets there and here to move work/plugins forward.

mejackreed:
[...]

PR #2299, opened 2019-03-22
Add plugin hook to window menu. Closes #2298

### LINKED ISSUE #2298: Plugin hook in window menu
There should be a place in the window menu for plugin components. This place should only show up if a plugin wants to render a component there, otherwise it's not visible. I know there is design needed and generally we should come up with a concept for plugin UX stuff. But to move the plugin system forward and be able to demo it, it would be nice to have this simple plugin playground.

charbugs:
Please see #2298 

plugin-section-in-window-menu


charbugs:
> we don't want to maintain targets for every possible plugin idea in code, but provide a generic method of adding plugins.

I agree with that in general and I wished there would be a generic solution for adding plugin components. But from what I understand so far there is no other *clean* solution for that. For deleting, replacing and wrapping components by plugins there are already generic methods in `withPlugins`. But *adding* a plugin component to the render tree of a target component always means that the target must treat the plugin explicitly.

Take for example this component:

```javascript
const TopBar = (props) => (
  <List>
    <Item>
      <MenuButton />
    </Item>
    <Item>
      <CloseButton />
    </Item>
  </List>
);
```

There is a generic solution to inject a plugin component into this render tree: You could target for example the `MenuButton` and render the plugin component as a sibling of it. This is what the original `miradorWithPlugins` function did. But there are two issues: a) To inject a plugin to `TopBar` you have to target the `MenuButton` (or another component of the render tree) and this is semantically messy. b) It will break the layout as the first Item now consists of `MenuButton` and the plugin component.

> What are the other target areas we are likely to create and how do we document and maintain them?

Unless we find another solution we indeed have to define and document plugin areas which I think souldn't be a big problem. There is another one I would like to implement: `BackgroundPluginArea`. This is a hidden component where plugins can add to which want to implement a feature without a visible UI component.

Can you give some additional explanations of your concerns? In which way could mainting be a problem here? Can you outline what plugin approach is used by Mirador 2?

camillevilla:
Thanks for providing this seam, @charbugs. 

@aeschylus, I don't think this approach *prevents* the more generic approach  for adding plugins. I believe that as long as the component's container imports `withPlugins`, a plugin author would be able to add a seam there, right? 

I'm not sure if you're referring to prototype discussions or the meeting back in January, but I thought we agreed in January that we should provide at least one recommended plugin seam in the UI. The rationale was preventing UI pollution. I think this PR gives a nice incentive to plugin authors to develop towards this seam.

We're not there yet, but other seams that I can imagine coming down the pipeline:
- wherever image manipulation tools end up after #1985
- adding new types of companion windows

cbeer:
> we don't want to maintain targets for every possible plugin idea in code, but provide a generic method of adding plugins.

I thought that was the pattern we were using; is the concern just that we don't want a `WindowPluginButtons` component, but we'd rather drop `PluginComponent` into `WindowTopMenu`
directly to remove some of the indirection?

I guess I'm with @camillevilla; this unblocks putting plugins in the window menu and when we get designs for how plugins should be displayed, it seems trivial to refactor.

charbugs:
> is the concern just that we don't want a WindowPluginButtons component, but we'd rather drop PluginComponent into WindowTopMenu
directly to remove some of the indirection?

Right, that's better. I will change it.

One note for better understanding: Not all plugins need the *add* mode to place components to mirador. E.g. we in Leipzig want to provide our own WorkspaceAdd panel to enable a discovery service. In order to do that we will probably use the *replace* mode. This can of course be considered as adding a custom component, but technically it works differently. What I said in the comment above only applies to the *add* mode.

PR #2303, opened 2019-03-22
Make the resource add button a medium-sized fab; fixes #2294

### LINKED ISSUE #2294: Reduce size of app bar FAB button to medium
We're currently using the `large` version of the "Add resource" / "Close form" FAB button in the app bar. The `medium` size would look better visually in the app bar, while still being big enough for user interaction:

fab

The FAB should remain centered in the app bar after changing sizes.

The size property is described here: https://material-ui.com/demos/buttons/#sizes

cbeer:
Screen Shot 2019-03-22 at 08 45 29

### DIFF
--- a/src/components/WorkspaceAddButton.js
+++ b/src/components/WorkspaceAddButton.js
@@ -19,6 +19,7 @@ export class WorkspaceAddButton extends Component {
     return (
       <Tooltip title={isWorkspaceAddVisible ? t('closeAddResourceMenu') : t('addResource')}>
         <Fab
+          size="medium"
           color="secondary"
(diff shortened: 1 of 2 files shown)


src/components/WorkspaceAddButton.js
```diff
@@ -19,6 +19,7 @@ export class WorkspaceAddButton extends Component {
     return (
       <Tooltip title={isWorkspaceAddVisible ? t('closeAddResourceMenu') : t('addResource')}>
         <Fab
+          size="small"
```
jkeck: Should this be `medium`?

PR #2307, opened 2019-03-22
Have window focused class extend windowTopBarStyle so that the …

### LINKED ISSUE #2305: Investigate window focus regression
breadcrumbs here: https://github.com/ProjectMirador/mirador/projects/1#card-18680623

jkeck:
…borderTop rules don't clash.

Fixes #2305 
Screen Shot 2019-03-22 at 10 11 19 AM

### DIFF
--- a/src/containers/WindowTopBar.js
+++ b/src/containers/WindowTopBar.js
@@ -34,15 +34,16 @@ const mapDispatchToProps = (dispatch, { windowId }) => ({
  * windowTopBarStyle: {minHeight: number, paddingLeft: number, backgroundColor: string}}}
  */
 const styles = theme => ({
-  focused: {
-    borderTop: `2px solid ${theme.palette.secondary.main}`,
-  },
+  focused: {},
   title: {
     ...theme.typography.h6,
     flexGrow: 1,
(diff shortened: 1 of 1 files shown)


src/components/WindowTopBar.js
```diff
@@ -29,7 +29,7 @@ export class WindowTopBar extends Component {
     } = this.props;
     return (
       <AppBar position="relative">
-        <Toolbar disableGutters className={classNames(classes.windowTopBarStyle, focused ? classes.focused : null, ns('window-top-bar'))} variant="dense">
```
cbeer: I think the idiomatic way is to do some trickery in the styles:

```
  focused: {},
  ....
  windowTopBarStyle: {
    '&$focused': {
      borderTop: `2px solid ${theme.palette.secondary.main}`,
    },
```

(e.g. https://github.com/mui-org/material-ui/blob/4a4650572c037190ef44e290c7f3bcb93c32c1b1/packages/material-ui/src/ExpansionPanelSummary/ExpansionPanelSummary.js#L25)

jkeck: makes total sense

PR #2308, opened 2019-03-22
Extract additional selectors from containers

cbeer:


src/state/selectors/windows.js
```diff
@@ -58,6 +58,14 @@ export const getCompanionWindowIds = createSelector(
   window => (window && window.companionWindowIds) || [],
 );
 
+export const getCompanionWindow = createSelector(
+  [
+    state => state.companionWindows,
+    (state, { companionWindowId }) => companionWindowId,
```
christopher-johnson: state is declared, but never read on line 64.  Rather than referencing an unused param, you could use `{} ` instead

cbeer: Using `{}` violates the eslint rule https://eslint.org/docs/rules/no-empty-pattern.

christopher-johnson: maybe you could disable the rule for these cases.  are unused declarations worse than empty patterns?    I guess [eslint no-unused-variables](https://eslint.org/docs/rules/no-unused-vars) is not savvy enough to read selectors, but tsc does.

cbeer: I'm not sure it's worth it, is it? If nothing else, keeping the explicit `state` nicely parallels all our other selectors.

jkeck: It seems that consistency is more beneficial here that one-off disabling / rule manipulations.

Issue #2304, opened 2019-03-22
Fix hide sidepanel toggle 


ggeisler:
Not sure if for some reason it is just me, but I can no longer hide the sidepanel because the tooltip for it obscures the hit target of the toggle:

<img width="362" alt="Screen Shot 2019-03-22 at 9 09 46 AM" src="https://user-images.githubusercontent.com/101482/54836978-902f7700-4c82-11e9-8008-b6fec23fe7a4.png">

camillevilla:
Ideal behavior: something like Google Maps
<img width="215" alt="Screen Shot 2019-03-22 at 9 22 01 AM" src="https://user-images.githubusercontent.com/5402927/54837580-fcf74100-4c83-11e9-95a9-3f2659aacbcc.png">

jkeck:
I _think_ this is closed now by #2321/#2318.  The placement isn't perfect (due to absolute positioning I believe).  We can re-open or create a new ticket if this needs more work.

PR #2318, opened 2019-03-23
Add a reselect selector to get companionAreaVisibility.

jkeck:
Part of #2304 

This doesn't do anything w/ the styling of the tooltip/button, but instead addresses the click behavior bug.

### DIFF
--- a/src/state/selectors/windows.js
+++ b/src/state/selectors/windows.js
@@ -95,3 +95,20 @@ export const getViewer = createSelector(
   ],
   (viewers, windowId) => viewers[windowId],
 );
+
+/**
+ * Returns the visibility of the companion area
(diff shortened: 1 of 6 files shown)


src/state/selectors/windows.js
```diff
@@ -95,3 +95,20 @@ export const getViewer = createSelector(
...
   (viewers, windowId) => viewers[windowId],
 );
+
+/**
+ * Returns the visibility of the companion area
+ * @param {object} state
+ * @param {object} props
+ * @return {Boolean}
+ */
+export const getCompanionAreaVisibility = createSelector(
```
jkeck: I'm not 100% sure if this was the original intent of this or not.

PR #2332, opened 2019-03-28
Basic plugin validation. Closes #2331

### LINKED ISSUE #2331: Basic plugin validation
Plugin objects should be checked for correct props.

charbugs:


src/extend/pluginValidation.js
```diff
@@ -0,0 +1,71 @@
...
+/** */
+const checkTarget = (plugin) => {
+  const { target } = plugin;
+  return isString(target);
+};
+
+/** */
+const checkMode = (plugin) => {
+  const { mode } = plugin;
+  return some(['add', 'wrap'], s => s === mode);
```
mejackreed: Is there a difference here between lodash's `some` and just es6 `some`?

src/extend/pluginValidation.js
```diff
@@ -0,0 +1,71 @@
...
+    || isNull(mapDispatchToProps)
+    || isFunction(mapDispatchToProps)
+    || isObject(mapDispatchToProps);
+};
+
+/** */
+const checkReducers = (plugin) => {
+  const { reducers } = plugin;
+  return isUndefined(reducers)
+    || (isObject(reducers) && every(values(reducers), isFunction));
```
mejackreed: Similar to previous comment, can we just use es6's `every`?

charbugs: yep

PR #2335, opened 2019-03-28
Add background component for plugins. Closes #2333

### LINKED ISSUE #2333: Adding a BackgroundPlugins component
Some plugins don't need a UI, but doing their work entirely in the background. But with the current plugin system a plugin needs to provide an component in order to get connected to state and dispatch. Therefore there should be a background component where such plugins can add to. The background component will be invisible.

charbugs:


__tests__/src/components/BackgroundPluginArea.test.js
```diff
@@ -0,0 +1,24 @@
...
+  const PluginComponentA = props => <div>A</div>;
+  /** */
+  const PluginComponentB = props => <div>B</div>;
+
+  const wrapper = shallow(
+    <BackgroundPluginArea
+      PluginComponents={[PluginComponentA, PluginComponentB]}
+    />,
+  );
+
```
glefi: May be there could be a check again, if the BackgroundPluginArea is rendered as well, when passing PluginComponents, like in the first test.

Issue #2286, opened 2019-03-21
Update dialogs (story)


ggeisler:
The application currently includes several dialogs, all of which need updating. Here are the actual work issues:

- [x] #2287 - Update the "Settings" dialog to be "Change theme"
- [x] #2288 - Update the "Select workspace type" dialog
- [x] #2289 - Update the "Download/Export" dialog
- [x] #2290 - Update the "Import workspace" dialog

## General info on dialogs

The current dialogs should follow one of two component types:

- [Simple dialogs](https://material-ui.com/demos/dialogs/#simple-dialogs)
- [Form dialogs](https://material-ui.com/demos/dialogs/#form-dialogs)

All dialogs should follow these styling guidelines:
- Title of the dialog is left-aligned
- Title uses Headline 2 typography defined in #2093 
- Text buttons use the BUTTON class typography defined in #2093 

### Simple dialogs
[Simple dialogs](https://material-ui.com/demos/dialogs/#simple-dialogs) enable the user to pick from a list of visible options. The selected option takes effect immediately and closes the dialog at the same time.

### Form dialogs
[Form dialogs](https://material-ui.com/demos/dialogs/#form-dialogs) contain action buttons. The user can select an action or cancel to close the dialog.

For all form dialogs:

- Buttons use the secondary color
- Buttons are right-aligned
- The leftmost one is always labeled "CANCEL" and is a [text button](https://material-ui.com/demos/buttons/#text-buttons)
- The rightmost one is the action that the dialog is intended to provide and is a [contained button](https://material-ui.com/demos/buttons/#contained-buttons)

Issue #2093, opened 2019-03-07
Typography classes & styles


jvine:
Below are the font sizes, weights, and letter-spacing for the classes defined here:
https://material.io/develop/web/components/typography/

(Specify rems; px are for reference.)

class | rem | px | weight | line-height | letter-spacing | other
--- | --- | --- | --- | --- | --- | ---
Headline 1 | 2.822rem | 45.159 | 300 | 1.2em | -.015em
Headline 2 | 1.575rem | 25.192 | 300 | 1.33em | 0
Headline 3 | 1.383rem | 22.127 | 300 | 1.33em | 0
Headline 4 | 1.215rem | 19.436 | 400 | 1.45em | 0.007em
Headline 5 | 1.138rem | 18.215 | 400 | 1.55em | 0.005em
Headline 6 | 1.067rem | 17.072 | 400 | 1.6em | 0.01em
Subtitle 1 | 0.937rem | 14.995 | 300 | 1.6em | .015em
Subtitle 2 | 0.878rem | 14.053 | 500 | 1.75em | .02em
Body 1 | 1rem | 16px | 400 | 1.6em | 0
Body 2 | 0.878rem | 14.053 | 400 | 1.6em | .015em
BUTTON | 0.878rem | 14.053 | 500 | 2.25**rem** | 0.09em | uppercase
Caption | 0.772rem | 12.344 | 400 | 1.6**rem** | 0.33em
OVERLINE | 0.678rem | 10.842 | 500 | 2em | .166em | uppercase

## assign classes

Some specific class assignments are needed to override default sizes. 

element | assign class
--- | ---
window title `<h2>` | Heading 6
companion window `<h3>` | Subtitle 1
`<dt>` |  Subtitle 2
`<dd>` | Body 1 
"Current item" etc. labels in About panel | OVERLINE
window dialog `<h3>` | Heading 2 (we don't have any of these yet)

<img width="367" alt="typography" src="https://user-images.githubusercontent.com/6945691/53993257-ba5e3200-40e3-11e9-953f-838a575ea21b.png">

PR #2338, opened 2019-03-29
#2287: replace selection list by simple list for change theme dialog

### LINKED ISSUE #2287: Update the "Settings" dialog to be "Change theme"
Related to #2286 

The current implementation uses a select list for the two theme options. We want to change this to use the [simple dialogs](https://material-ui.com/demos/dialogs/#simple-dialogs) approach, where we can instead present the options in a simple list:

theme

Also, the theme switcher dialog is currently triggered by the **Settings** workspace menu choice. We have changes to the workspace menus planned that we'll present in the near future, but for now, we should just rename the **Settings** menu choice to **Change theme** and that will trigger the Change theme dialog:

theme-menu

### Details
- Remove icon from "Settings" menu item
- Change label of "Settings" menu item to "Change theme"
- Rename title of dialog to "Change theme"
- Replace select list with simple list of the two theme options, using the [simple dialogs](https://material-ui.com/demos/dialogs/#simple-dialogs) approach
- Icon for both options is `@material-ui/icons/PaletteSharp`
- Icon colors: the dark version is `#000000` and the light version is `#BDBDBD`
- When the user mouses over or keyboards to one of the options, it gets the item highlighting background, as seen in the [simple dialogs demo](https://material-ui.com/demos/dialogs/#simple-dialogs):

theme-highlighted

glefi:
Closes #2287


ggeisler:
@glefi Nice start on this. I'm not sure if this PR is intended to be complete or a work-in-progress but here are a couple of minor issues that should be updated, in case you are not already aware of them:

- The title heading for the dialog window should use Headline 2 typography defined in #2093, rather than the `H6` it currently uses. (This was noted in the [dialog story ticket](https://github.com/ProjectMirador/mirador/issues/2286).)

Some things that don't match the component demo at https://material-ui.com/demos/dialogs/#simple-dialogs but should:
- Hover highlighting as you move over the different theme options
- Cursor should be the hand cursor when highlighting any part of the theme options; currently it shows a pointer cursor when hovering over the icon and changes to a selection cursor when hovering over the theme option text
- The dialog should close automatically when selecting an option; currently the theme changes when selecting an option but the dialog doesn't close until you click again

Also, the dialog doesn't seem accessible via keyboard?

glefi:
@ggeisler Thanks for your feedback. The dialog should now meet all the requirements. It would be nice if you could take another look at it.

ggeisler:
Yep, the Change theme dialog looks good to me.

ggeisler:
Just to be clear, my comments above on this PR were for only the Change theme dialog, not the Select workspace type dialog.

@glefi I assume you know that the Change workspace type dialog still needs:

- Adding the SVG icons I pointed you to
- Showing highlighting for the currently active workspace type

The latter issue does work in the Change theme dialog but last I saw when you open the Change workspace type dialog it does not show any highlighted for the currently selected type.

Issue #1906, opened 2019-02-15
Update 'light' theme color palette


ggeisler:
Here's a first pass at specifying the `light` color palette. We expect to have to make multiple passes at this to get it where we want it, but it'll be easier to fine-tune the palette after getting some fundamental updates incorporated in the app.

There are two parts to this, I believe:

- Specifying a basic color palette for the light theme
- Updating the existing color-category-to-component mapping

### Color palette specifications

```
palette: {
  type: 'light',
  primary: {
    main: '#f5f5f5',
    light: '#ffffff',
    dark: '#eeeeee',
  },
  secondary: {
    main: '#1967d2',
    light: '#64b5f6',
    dark: '#0d47a1',
  }
}
```

For the other palette categories, we can start with what I think are the defaults:

Background: #ffffff
Surface: #ffffff
Surface - dark: #f5f5f5 (not sure this is part of the default palette or not)
Error: #b00020

### Color-category-to-component mapping

There are some components that we want to map to color categories that are different from (the best I can tell) the defaults:

- Application bar: primary - main
- Workspace surface: surface - dark
- Window surface: primary - dark
- Window top bar: primary - light
- Thumbnail strip surface: primary - dark
- Selection controls: secondary
- Buttons: secondary
- FABs: secondary
- Chips: secondary
- Text links: secondary

For other components, such as menu and dialog surfaces, the default (#ffffff) is fine. All of the "on text" defaults and things like text fields will probably be okay (or handled automatically based on our palette).

We'll likely want to adjust some things in the future (e.g., in the final palette, chips probably should not be the secondary color) but we'll be able to make better judgements if we can get these initial palette and mappings in place.

camillevilla:
Closed by #1979 and and #1980

mejackreed:
@ggeisler @camillevilla I had some questions about this.

Specifically: 

> Workspace surface: surface - dark

And similar areas that reference "dark". Is it possible that we should not reference "dark" or "light" for a specific area, but should reference it based off of what theme is available?

ggeisler:
@mejackreed I'm not sure that would produce reliable results. The `light` and `dark` variants are just additional colors added to the color palette so there are more color options available in the palette. Depending on the color palette, trying to automatically calculate `light` based on the primary color, for example, might be fine for palette A but not be the ideal color for palette B, if that makes sense.

[...]

PR #2340, opened 2019-03-29
Updates theme approach

mejackreed:
Inspired by embedding M3 in sul-embed, this PR builds off of discussion in #1906 and work completed in #1979 .

Based off [Material-UI documentation](https://material-ui.com/customization/themes/#type-light-dark-theme) my understanding is that setting `light` and `dark` primary values is used for setting the color when in a specific theme. This pull request resolves this by setting a new `lighten` and `darken` keys. While maybe not the final way we could do this, it is a refactor without modifying the colors currently being used.

Future alternatives could be to use the `lighten` and `darken` methods from Material-UI to do these calculations on the fly.

@ggeisler happy to chat more about this approach

### Before
Screen Shot 2019-03-29 at 10 06 08 AM

### After
Screen Shot 2019-03-29 at 10 06 18 AM

### Before
Screen Shot 2019-03-29 at 10 06 36 AM

### After
Screen Shot 2019-03-29 at 10 06 49 AM

### DIFF
--- a/src/config/settings.js
+++ b/src/config/settings.js
@@ -8,13 +8,17 @@ export default {
       type: 'light', // dark also available
       primary: {
         main: '#f5f5f5',
-        light: '#ffffff',
-        dark: '#eeeeee',
       },
       secondary: {
         main: '#1967d2',
-        light: '#64b5f6',
-        dark: '#0d47a1',
+      },
+      darkened: { // custom colors used for a specific offset in some places
+        dark: '#000000',
+        light: '#eeeeee'
+      },
+      lightened: { // custom colors used for a specific offset in some places
+        dark: '#424242',
+        light: '#ffffff',
       },
       error: {
         main: '#b00020',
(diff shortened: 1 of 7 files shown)


ggeisler:
@mejackreed Okay, great. I'll read up on the theme config stuff and make sure I understand it better, and check out the results from this PR in the themes. If @jvine and I think we need to still adjust colors, we can discuss.

src/config/settings.js
```diff
@@ -8,13 +8,17 @@ export default {
...
         main: '#1967d2',
-        light: '#64b5f6',
-        dark: '#0d47a1',
+      },
+      darkened: { // custom colors used for a specific offset in some places
+        dark: '#424242',
+        light: '#eeeeee'
+      },
+      lightened: { // custom colors used for a specific offset in some places
+        dark: '#000000',
```
ggeisler: It seems odd that (for the Dark theme, if I'm understanding things correctly) a "lightened" color would be `#000000` or black, while the "darkened" color would be `#424242`, which is an off-black. Should those be reversed?

I haven't yet examined the theme in detail to how or if these values are relevant so feel free to ignore this; it just seems on the surface that the "darkened" version should be darker than the "lightened" version.

mejackreed: Yeah it totally could be changed.. this is really just a placeholder as my understanding was that dark theming hasn't really taken place yet.

ggeisler: Sure, makes sense. If you don't mind making the update, I'd switch those two values just so we're starting with something that makes more logical sense (having the darkened variant darker than the lightened variant).

mejackreed: Of course! I've made the update.

ggeisler: Thanks Jack

PR #2349, opened 2019-03-29
Updates the Window and Workspace fullscreen icons

### LINKED ISSUE #2343: Update icons used for window and application maximize
The **workspace** maximize/full-screen icon correctly uses the icon specified in #1947, but when in full-screen mode, the icon is not switched to the `@material-ui/icons/FullscreenExitSharp` icon specified in that ticket. The icon should change when in full-screen mode.

The **window** maximize/full-screen icon does change when in window maximize mode, but does not use the two icons specified for `window maximize` and `window maximize exit` specified in #1947.

mejackreed:
Fixes #2343 

The Window FullMin/Max was an easy fix. The Workspace state area may have gotten moved in some previous selector work. The second commit moves this back to the workspace state.

### DIFF
--- a/__tests__/src/selectors/config.test.js
+++ b/__tests__/src/selectors/config.test.js
@@ -3,7 +3,6 @@ import {
   getShowZoomControlsConfig,
   getTheme,
   getWorkspaceType,
-  getFullScreenEnabled,
   getContainerId,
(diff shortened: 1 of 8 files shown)


ggeisler:
👍 Looks good.

PR #2350, opened 2019-03-29
sets default thumbposition to off fixes #2346

### LINKED ISSUE #2346: Make default state of thumbnail strip OFF
It looks like the current default value for the thumbnail strip is `Bottom` (when opening a new window, the thumbnail strip is always open and placed at the bottom of the window).

We should make the default value for the thumbnail strip to be `Off`. This will make new windows look less cluttered and not give the user more information than she might want.

mejackreed:
Configures the spec file to show similar to previous setup.

### DIFF
--- a/__tests__/integration/mirador/index.html
+++ b/__tests__/integration/mirador/index.html
@@ -15,10 +15,10 @@
        windows: [{
          loadedManifest: 'https://iiif.harvardartmuseums.org/manifests/object/299843',
          canvasIndex: 2,
+         thumbnailNavigationPosition: 'far-bottom',
        },
        {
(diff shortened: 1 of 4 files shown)


camillevilla:
Doh. Sorry, @mejackreed 😥
I merged this because the initially configured window, which doesn't have a set thumbnail strip position (e.g. the Bodleian one), is responding appropriately to the `defaultPosition`.

However, it looks like new windows still open with the thumbnail strip  at `far-bottom`. I think the intention of #2364 I'm not sure if there are some changes to the actions or default prop types required to make this happen, but I would be interested in pairing so I can learn!

PR #2352, opened 2019-04-01
closes #2288: rework workspace type selection dialog

### LINKED ISSUE #2288: Update the "Select workspace type" dialog
Related to #2286

This is intended to supersede existing ticket #1838 by using the [simple dialogs](https://material-ui.com/demos/dialogs/#simple-dialogs) approach, which will enable the user to change workspace types with fewer actions.

The basic idea is to show a simple dialog with all of the workspace options visible (rather than the current select list or the form-oriented dialog in #1838). The user clicks on a workspace type option, the dialog is dismissed, and the application switches to that workspace type.

So the updated dialog would look similar to this:

workspace-type-1

Changes needed to the current implementation:
- Heading should be changed to use an `H2`
- Dialog width should probably be around 400px
- Select list and label should be replaced with a list of the workspace options (Mosaic and Elastic, though eventually we would add Grid)
- Each option consists of a thumbnail (100px wide), a heading (styled using the `H3` typography, but don't use a heading element), and description text (styled using `Body1` typography):

Elastic: `Move and size windows freely in an unlimited workspace. Windows can overlap.`

Mosaic: `Move and size windows in relation to each other, within the visible frame.`

(Check with @jvine for the SVG images; an initial implementation could omit them, if that helps getting started on it sooner.)

When the user mouses over or keyboards to one of the options, it gets the item highlighting background, as seen in the [simple dialogs demo](https://material-ui.com/demos/dialogs/#simple-dialogs):

workspace-type2

glefi:
@ggeisler The `ListItemText` component provides the possibility to define a primary and a secondary text, that I used. I'm not sure, if the resulting text formatting matches the desired one.
@jvine Could you attach the workspace type thumbs? I've added some placeholders for now.


ggeisler:
@glefi Thanks for checking. As I mentioned in #2288, the typography for those elements should be styled as an `H3` (the label of the workspace type; but leave the element as is; don't make it an `<h3 />`) and `Body1` (the text description of each type). See the typography details for those  ("Headline 3" and "Body 1") here: https://github.com/ProjectMirador/mirador/issues/2093

If you update those elements to match the typographic specifications in #2093, plus update the dialog header to be an `H2`, the result should look similar to this:

workspace-type-dialog

By the way, the mouse selection looks good (highlighting and cursor) but I notice I can't use the dialog with the keyboard. But I know this is WIP so you might not have gotten to that yet.

glefi:
@ggeisler I've changed the text formatting according to the typographic specifications and the keyboard navigation works now, too. But now the required height of the text elements is slightly higher than 100px, maybe we should increase the image height, what do you think?

src/components/WorkspaceSelectionDialog.js
```diff
@@ -2,55 +2,127 @@ import React, { Component } from 'react';
...
     } = this.props;
     return (
-      <Dialog id="workspace-settings" container={container} open={open} onClose={handleClose}>
-        <DialogTitle id="form-dialog-title">{t('workspaceSelectionTitle')}</DialogTitle>
+      <Dialog
+        aria-labelledby="workspace-selection-dialog-title"
+        container={container}
+        id="workspace-settings"
+        onClose={handleClose}
+        onKeyDown={event => this.keyDownHandler(event)}
```
glefi: I don't know if it's common, to set the keyDownListener to the hosting node, but I thought, especially when there will be more workspaceTypes perhaps in the future, we could save some locs this way.

src/components/WorkspaceSelectionDialog.js
```diff
@@ -2,55 +2,127 @@ import React, { Component } from 'react';
...
+    updateConfig({
+      workspace: {
+        type: workspaceType,
+      },
+    });
+    handleClose();
+  }
+
+  /** */
+  selectPreviousItem() {
```
glefi: A simple toggling function would have been sufficient here, but I tried to be prepared for some additional workspaceTypes that may come in the future.

ggeisler:
> But now the required height of the text elements is slightly higher than 100px, maybe we should increase the image height, what do you think?

@glefi Sure, assuming you only need to make the images a little bit taller (110 or 120px?) that sounds fine.

This is looking good but a couple things I noticed (in case you aren't aware):

- When I switch to elastic mode (using the mouse) and then choose the workspace type menu item again, the dialog shows the selected mode is Mosaic, when it should reflect that I am currently in Elastic:

Screen Shot 2019-04-02 at 9 56 41 AM

This actually works correctly when using the keyboard (nice work on getting the keyboard nav working!) but compare the difference when using the mouse. 

- The color of the text for the workspace type description `<p />`s is a gray (I think coming from `.jss381` or the simple dialog default styles) but it should be our default color (I think the color in `.jss357` that is overriden by `.jss381`). Otherwise, the color contrast when the item is selected is not accessible.

glefi:
@ggeisler Thanks for your feedback. I've added your suggestions to the pr. What do you think about it?

ggeisler:
Looks good @glefi!

src/components/WorkspaceSelectionDialog.js
```diff
@@ -28,41 +43,70 @@ export class WorkspaceSelectionDialog extends Component {
...
             >
-              <img src="/src/images/elastic.jpg" alt={t('elastic')} />
-              <ListItemText
-                primary={t('elastic')}
-                secondary={t('elasticDescription')}
-              />
+              <Card className={classes.card}>
+                <CardMedia
+                  className={classes.media}
+                  image="/src/images/icon_workspace_elastic.svg"
```
mejackreed: I'm concerned that referencing the image this way will hardcode the image path (and its not included in the build). Is there a way to reference the SVG directly?

camillevilla: @mejackreed Do we have a pattern established for this anywhere else in the repo?

mejackreed: Yeah, I think the predominate pattern is to wrap the SVG in a component like we do with the icons:

https://github.com/ProjectMirador/mirador/blob/master/src/components/icons/BookViewIcon.js

glefi: @ggeisler Could you provide the two SVGs in a scaled variant (90x120)? Because resizing the files with inkscape seems not to work properly.

ggeisler: @glefi I've updated the gists I previously gave you, so try those and see how they work:

- Mosaic: https://gist.github.com/ggeisler/21c514a51ae9eb0cf62288a1f05e6ca0
- Elastic: https://gist.github.com/ggeisler/3bacfde8f287748603a6bc33e0ef8197

src/components/WorkspaceSelectionDialog.js
```diff
@@ -28,41 +44,68 @@ export class WorkspaceSelectionDialog extends Component {
...
     const {
-      classes, container, handleClose, open, children, t,
+      classes, container, handleClose, open, children, t, workspaceType,
     } = this.props;
     return (
       <Dialog
         aria-labelledby="workspace-selection-dialog-title"
         container={container}
-        id="workspace-settings"
+        id="workspace-selection-dialog"
```
mejackreed: I like how you renamed this to be more accurate

PR #2359, opened 2019-04-01
Update gallery view styling/layout

### LINKED ISSUE #2312: Fix Gallery View Item layout styling
- [x] items should have a fixed height
- [x] item captions should have a fixed height area in which to render, providing 2 lines of text
- [x] ~~item captions should break words in order to fit more text on both lines~~
- [x] item captions should use ellipses if they are too long
- [x] it images are too skinny, there should be a minimum width of 60px for the total item layout box so that captions are still legible

jkeck:
Closes #2312 

## Chrome (before)
chrome-before

## Chrome (after)
chrome-after

---

## Firefox (before)
ff-before

## Firefox (after)
_**Note:** Firefox does not support `display: box` yet, so the text is truncated w/o an ellipsis_
ff-after

---

## Safari (before)
safari-before

## Safari (after)
safari-after

### DIFF
--- a/__tests__/src/components/GalleryView.test.js
+++ b/__tests__/src/components/GalleryView.test.js
@@ -9,7 +9,7 @@ function createWrapper(props) {
   return shallow(
     <GalleryView
       canvases={manifesto.create(manifestJson).getSequences()[0].getCanvases()}
-      classes={{ currentCanvas: 'currentCanvas' }}
+      classes={{ galleryViewItemCurrent: 'galleryViewItemCurrent' }}
       window={{
         canvasIndex: 0,
         id: '1234',
@@ -34,7 +34,7 @@ describe('GalleryView', () => {
     expect(wrapper.find('div[role="button"]').length).toBe(3);
   });
   it('sets a mirador-current-canvas class on current canvas', () => {
-    expect(wrapper.find('div[role="button"]').at(0).props().className).toEqual('currentCanvas');
+    expect(wrapper.find('div[role="button"]').at(0).props().className).toEqual('galleryViewItemCurrent');
   });
   it('renders the canvas labels for each canvas in canvas items', () => {
     expect(wrapper.find('WithStyles(Typography)').length).toBe(3);
--- a/src/components/GalleryView.js
+++ b/src/components/GalleryView.js
@@ -47,7 +47,7 @@ export class GalleryView extends Component {
                     aspectRatio={manifestoCanvas.aspectRatio}
                     style={{ margin: '0 auto' }}
                   />
-                  <Typography variant="caption">
+                  <Typography variant="caption" className={classes.galleryViewCaption}>
                     {manifestoCanvas.getLabel()}
                   </Typography>
                 </div>
--- a/src/components/GalleryView.js
+++ b/src/components/GalleryView.js
@@ -32,7 +32,7 @@ export class GalleryView extends Component {
                   className={
                     classNames(
                       classes.galleryViewItem,
-                      canvas.index === window.canvasIndex ? classes.currentCanvas : '',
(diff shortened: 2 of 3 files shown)


ggeisler:
Nice. I know the PR is still `draft` and this wasn't included in #2312 but I wonder if we should reduce the line-height of the captions?

@jvine What do you think? The current line-height matches the Caption specs in #2093 but I think in this context using a headline line-height (e.g., `1.33em`) might work better? It would help tighten up the whole layout while still be very readable, I think.

jkeck:
Yeah, in an earlier pairing session the line-height of the captions was called out as needing some tweaking.  Happy to change it (and the related heights) as needed.

jkeck:
I set the line-height for the caption to be `1.33rem` and the height of the caption area to be a fixed height of `3rem`, but it looks like the descender is getting clipped a bit.

Also, the original ticket says something about adding a word break, but wasn't sure exactly what we wanted there.  Is there a specific word break strategy that you all have in mind?

ggeisler:
Actually my suggestion was `1.33em` not `rem`, which does seem to make a difference (slightly more compact line height).

jkeck:
Okay, so should the fixed height be `2rem` then?  (`line-height: 1.33em` and `height: 2rem`)?

Forgive my ignorance 😬

ggeisler:
That seems to work, so I _think_ so...

ggeisler:
Also @jkeck,  about this question:

> Also, the original ticket says something about adding a word break, but wasn't sure exactly what we wanted there. Is there a specific word break strategy that you all have in mind?

I think @aeschylus wrote that and I'm not sure what he had in mind specifically, but I'm not sure we need to specify anything. The default seems to be what we want. The main alternative would be to add `word-break: break-all;` but I don't think that produces the result we want (does allow more letters per line in some cases, but at the cost of weird word breaks).

jkeck:
Agreed, I played w/ some of the various options, and none of them seemed to be better than the default.

jkeck:
Okay, pushed it up, let me know if that looks good and I'll make it ready for review.

ggeisler:
Hmm, the captions look good but now the rows seem to have way too much whitespace between them. The whitespace increases as the browser width increases (or hits different breakpoints, I think):

Screen Shot 2019-04-02 at 9 06 26 AM

Seems like the row height should be static across breakpoints, since the image height and caption height is fixed?

jkeck:
Okay, so what height would you like me to set then?

ggeisler:
Wait, I just adjusted it again to play with the height and it snapped back to how it was before, where the height is static and doesn't change with the breakpoints. So I guess what I was looking at was not accurate. So I don't think any change is needed, sorry.

jvine:
1. We need `word-break: break-word` to handle long words - otherwise they will break the spacing. First image has `word-break: break-word`, second does not:
    Screen Shot 2019-04-02 at 11 31 18 AM

2. As for the line-height and overall height... 
a) I feel like `1.33em` is too close for readability/accessibility for such small text. The guideline would be `1.5em`, which is just slightly roomier without being loose. 
b) I think it's potentially brittle if we use different measures for the line height (em) and overall height (rem) -- so if we use `1.5em`, the height should be `3em`. 

**Before (`1.33em` and `2rem`)**
Screen Shot 2019-04-02 at 11 44 36 AM

**After (`1.5em` and `3em`)**
Screen Shot 2019-04-02 at 11 44 57 AM

It's a tiny difference but it's still a bit closer than `1.25rem` and it's officially compliant.

ggeisler:
I guess it's important to test on more than one fixture! I used NGA and didn't see any difference with `word-break: break-word` so thought that was the default but with the other object it's clear that it does make a difference...

PR #2361, opened 2019-04-02
closes #2289 update-download-export-dialog

### LINKED ISSUE #2289: Update the "Download/Export" dialog
Related to #2286

The title of this dialog is currently "Download/Export workspace" but since it seems that the only way it can be used is to copy the configuration presented, we suggest simplifying the title to "Export workspace" Also, the dialog currently does not have any buttons. Buttons should be added, following the approach used in [form dialogs](https://material-ui.com/demos/dialogs/#form-dialogs).

- Change the label of the workspace menu item that triggers this dialog to "Export workspace"
- Change the dialog heading to "Export workspace"
- Add [text button](https://material-ui.com/demos/buttons/#text-buttons) labeled "Cancel"
- Add a [contained button](https://material-ui.com/demos/buttons/#contained-buttons) labeled "Copy"
- Use the `scroll=paper` [long content option](https://material-ui.com/demos/dialogs/#scrolling-long-content) so users don't have to scroll through entire view of the code if they just want to copy it

The updated dialog would look similar to this:

export

glefi:
closes #2289
Haven't found a way to check if the content of the textarea was copied to the clipboard actually, but I've tried it manually.


camillevilla:
(and if I don't hear back from anyone on this side of the Atlantic by EOD I'll merge this)

mejackreed:
I'm 👍 with the current way that WorkspaceMenu is working. Some changes in `WorkspaceMenu` could allow us to not have to manage this in component state enabling some future simplicity.

PR #2362, opened 2019-04-02
Create utility function for state updates

charbugs:
As preparation for refactoring of reducers. This functions would also replace immutable.js.

### DIFF
--- a/__tests__/src/reducers/utils.test.js
+++ b/__tests__/src/reducers/utils.test.js
@@ -1,0 +1,71 @@
+import { setItem, updateItem, removeItem } from '../../../src/state/reducers/utils';
+
+describe('setItem', () => {
+  it('does not change input object', () => {
+    const object = { foo: { a: 1 } };
+    setItem(object, 'foo', { b: 2 });
+    expect(object).toEqual({ foo: { a: 1 } });
+  });
+
+  it('sets a new item with the provided props', () => {
+    const object = { foo: { a: 1 } };
+    const result = setItem(object, 'bar', { b: 2 });
+    expect(result).toEqual({
+      bar: { b: 2 },
+      foo: { a: 1 },
+    });
+  });
+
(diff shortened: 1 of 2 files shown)


mejackreed:
I'm not sure I yet understand the implications here. Which refactors are you talking about?

charbugs:
It is about the immutable object creation pattern like this:

```
return {
  ...state,
  [action.windowId]: {
    ...state[action.windowId],
    view: action.viewType,
  },
};
```

Most of this can be written in one or two lines with helper functions: 

```
return update(state, action.windowId, { view: action.viewType });
```

But I just realized, that there are utility functions for immutable state update in lodash/fp.

mejackreed:
Gotcha. So the intention is to replace something like this?

https://github.com/ProjectMirador/mirador/blob/master/src/state/reducers/manifests.js#L9-L17

charbugs:
Right. Or this:

```
case ActionTypes.REMOVE_WINDOW:
      return Object.keys(state).reduce((object, key) => {
        if (key !== action.windowId) {
          object[key] = state[key]; // eslint-disable-line no-param-reassign
        }
        return object;
      }, {});
```

to:

```
case ActionTypes.REMOVE_WINDOW:
      return unset(state, action.windowId);
```

charbugs:
Technically it will be slower for sure because there are at least two more function calls involved on every state update. But I guess you wondering if the helper functions perform a deep clone on the state object. Of course that would be a peformance issue. According to [this article](https://hackernoon.com/a-small-trick-to-write-clean-reducers-a0b1b1eff3d2) the lodash/fp functions do not deep clone the object but only update the needed parts. I think this is similar to the raw pattern that we use currently.

mejackreed:
Yeah I was thinking less about function calls and more about returning new objects if there was no update. So it seems like this will work fine.

PR #2364, opened 2019-04-02
Handle annotation resources that are a single object…

### LINKED ISSUE #2329: Sometimes a lack of annotations crashes the viewer
Screenshot 2019-03-26 at 8 36 08 AM

This has happened a few times with https://gist.githubusercontent.com/jeffreycwitt/1a75fdb4a97e1c2a98bd35797aad263d/raw/859104cb6cd7bd99f3be668f064066f4b3ba2b29/manifest-with-three-level-deep-toc.json for some reason

jkeck:
…(in addition to an array of objects)

Fixes #2329 

Example Manifest: https://gist.githubusercontent.com/jeffreycwitt/1a75fdb4a97e1c2a98bd35797aad263d/raw/859104cb6cd7bd99f3be668f064066f4b3ba2b29/manifest-with-three-level-deep-toc.json

Click on the canvas labeled `216r` (which currently crashes mirador in master)

### DIFF
--- a/src/lib/Annotation.js
+++ b/src/lib/Annotation.js
@@ -1,3 +1,4 @@
+import flatten from 'lodash/flatten';
 import AnnotationResource from './AnnotationResource';
 /** */
 export default class Annotation {
@@ -21,6 +22,7 @@ export default class Annotation {
   /** */
   get resources() {
     if (!this.json || !this.json.resources) return [];
-    return this.json.resources.map(resource => new AnnotationResource(resource));
+
+    return flatten([this.json.resources]).map(resource => new AnnotationResource(resource));
   }
 }
--- a/__tests__/src/lib/Annotation.test.js
+++ b/__tests__/src/lib/Annotation.test.js
@@ -22,5 +22,11 @@ describe('Annotation', () => {
         { resources: [{ foo: 'bar' }] },
       ).resources.forEach(resource => expect(resource).toBeInstanceOf(AnnotationResource));
     });
+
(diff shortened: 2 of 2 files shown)


src/lib/Annotation.js
```diff
@@ -21,6 +21,13 @@ export default class Annotation {
   /** */
   get resources() {
     if (!this.json || !this.json.resources) return [];
-    return this.json.resources.map(resource => new AnnotationResource(resource));
+    let resourceList;
+    if (Array.isArray(this.json.resources)) {
```
jkeck: Is there a better way to handle this sort of check?

mejackreed: We could do something like this:

https://github.com/ProjectMirador/mirador/blob/edd913a3a05a79bbedf0237c9917e6818943cc77/src/lib/AnnotationResource.js#L36

jkeck: Ah, so pull those in from `lodash`?

mejackreed: Yep.. we are already using them elsewhere to do similar type things.

src/lib/Annotation.js
```diff
@@ -1,3 +1,4 @@
+import { flatten } from 'lodash';
```
mejackreed: So importing this way will just reimport the entire lodash library... /me shakes my head at JavaScript... 

a work around is this: https://github.com/ProjectMirador/mirador/blob/edd913a3a05a79bbedf0237c9917e6818943cc77/src/lib/AnnotationResource.js#L2

PR #2365, opened 2019-04-02
Pass params using DOM Apis to prevent rerender by function prop

mejackreed:
See https://reactjs.org/docs/faq-functions.html#how-do-i-pass-a-parameter-to-an-event-handler-or-callback

Before I move forward with more of this would be interested to hear others feedback


marlo-longley:
This draft explored using DOM `data-*` attributes to pass parameters and avoid unstable function prop references. React's guidance has since evolved (hooks, useCallback, useMemo) different solutions to this problem. Can ticket/investigate elsewhere.

PR #2381, opened 2019-04-03
Don't allow single windows in non-elastic view to be dragged (and style as such)

### LINKED ISSUE #2345: Suppress move/drag icon in Mosaic mode when only one window
Currently when in mosaic mode with only one open window, we show the move/drag icon when the user hovers over the drag area just as we do when there are multiple open windows. However, in mosaic mode with only one open window there is nowhere to move the window so in that case we should suppress the drag area of the top bar and not display the move/drag icon.

jkeck:
Closes #2345

### DIFF
--- a/src/components/Window.js
+++ b/src/components/Window.js
@@ -18,17 +18,20 @@ export class Window extends Component {
    * additional functionality based on workspace type
    */
   wrappedTopBar() {
-    const { manifest, window, workspaceType } = this.props;
+    const {
+      manifest, window, workspaceType, windowDraggable,
+    } = this.props;
     const { mosaicWindowActions } = this.context;
     const topBar = (
       <div>
         <WindowTopBar
           windowId={window.id}
           manifest={manifest}
+          windowDraggable={windowDraggable}
         />
       </div>
     );
-    if (workspaceType === 'mosaic' && window.maximized === false) {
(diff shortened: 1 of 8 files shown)


src/components/Window.js
```diff
@@ -18,17 +18,20 @@ export class Window extends Component {
...
       <div>
         <WindowTopBar
           windowId={window.id}
           manifest={manifest}
+          windowDraggable={windowDraggable}
         />
       </div>
     );
-    if (workspaceType === 'mosaic' && window.maximized === false) {
+    if (workspaceType === 'mosaic' && windowDraggable) {
```
jkeck: This was going to be a bit cleaner until I remembered/realized that we're using two different drag-n-drop libraries for Mosaic and Elastic. _sigh_

src/state/selectors/windows.js
```diff
@@ -112,3 +113,21 @@ export const getCompanionAreaVisibility = createSelector(
...
   },
 );
+
+/**
+ * Returns the draggability of a window
+ * @param {object} state
+ * @param {object} props
+ * @return {Boolean}
+ */
+export const getWindowDraggability = createSelector(
```
jkeck: Folks have been saying a lot of words about selectors lately.  Not sure if this is the correct usage or not.

PR #2389, opened 2019-04-04
replace size-limit with bundlesize for CI goodness

mejackreed:


cbeer:
are we using the CI part of bundlewatch too? are the necessary variables magically set in netlify, or do they need to go somewhere?

mejackreed:
I was hoping that everything just worked. I set the `BUNDLEWATCH_GITHUB_TOKEN` in Travis.. not sure what else we need to setup? but it didn't work

cbeer:
Ah; I was expecting to see it as an encrypted secret in the travis.yml. I'm happy to ship and see if it works.

PR #2392, opened 2019-04-04
[WIP] Add TOC

aeschylus:
basictoc


aeschylus:
Still writing tests, but @ggeisler and @jvine there are probably a couple pieces of feedback you'll want to put it, so check out the https://deploy-preview-2392--mirador-dev.netlify.com/__tests__/integration/mirador/ when you get a chance. 🙏

camillevilla:
@aeschylus chatted with @jvine and @ggeisler about this this morning and we agreed that styling refinements and accessibility work will get split out to other tickets. The one remediation here would be removing the `h6`s from this work.

PR #2394, opened 2019-04-04
Add distinction between highlighted/selected annotations...

### LINKED ISSUE #2383: Distinguish between highlighted and selected annotations
Related to #2264. 

We currently don't distinguish between highlighted and selected annotations, but with the "Highlight all" option in the Annotations panel, we need to. Below is the recommendation for adding the distinction.

## Terminology

**Highlighted** annotation - purpose is to indicate to the user where on the canvas the annotation is, but user has not indicated this is an annotation they care about or want to select. 

- Shown on canvas with blue border (#00BFFF - fixed color; doesn't change with theme)
- Indicated in panel with hover/focus highlighting (the lighter gray used in places like the workspace menu, for example). So when a mouse user hovers over an annotation item in the panel list, or a keyboard user moves into an annotation item (I think arrows into, though keyboard nav in the panel probably isn't working yet), the annotation item area is highlighted in gray. 

**Selected** annotation - purpose is to indicate to the user the annotation(s) they have added to the selection list (annos of interest). 

- Shown on canvas with yellow border (current color)
- Indicated in panel with selected item highlighting (the darker gray you currently use in #2257). So when a mouse user clicks on an annotation item in the panel, or a keyboard users hits Enter while focused on an annotation item in the panel, the canvas anno border changes to yellow, and the panel item area is highlighted with the darker gray.

## Examples
Hopefully these will clarify the guidelines above.

### Highlight all OFF, hover/focus one anno item
User is mousing over, or with the keyboard has arrowed into (but not pressed Enter), the sixth annotation:

OFF-hover-none-selected

### Highlight all OFF, selected one anno item
User has clicked on, or with the keyboard has pressed Enter on, the sixth annotation:

OFF-one-selected

### Highlight all OFF, selected one anno item, hover/focus one anno item
After selecting the sixth annotation, the user is mousing over, or with the keyboard has arrowed into (but not pressed Enter), the second annotation:

OFF-one-selected-one-highlighted

### Highlight all OFF, selected three anno items
Here the user has selected three different anno items, by clicking on each one, or with the keyboard arrowing into and pressing Enter on each one:

OFF-Select-three

If the user were to click, or arrow into and press Enter, on one of the selected anno items, it would deselect, removing the corresponding panel highlight and the yellow border on the canvas.

### Highlight all ON
With the Highlight all toggle set to on, all of the annotations on the canvas are highlighted. Note that none of the panel anno items are highlighted, because when turning the toggle on the user can't at the same time be hovering over, or have arrowed into, one of the panel annotations. 

ON

### Highlight all ON, hover/focus one anno item
If after turning on the Highlight all toggle, the user mouses over, or with the keyboard arrows into one of the panel annotations, the anno item gets the hover/focus highlighting (but there is no change to the corresponding canvas annotation indicator, because it is already shown in blue):

ON-highlight-one

### Highlight all ON, selected one anno item
If the user then clicks on, or presses Enter on, one of the panel annotations, it is shown with the selected highlighting and the corresponding canvas annotation border is changed to yellow:

ON-select-one

### Highlight all ON, selected three anno items
If the user goes on to click on, or press Enter on, two additional annotations in the panel, they are similarly shown as selected:

ON-select-three

If the user was to turn off the Highlight all toggle in state above, the selected annotations would remain (the selected highlighting of the three on the panel and the corresponding yellow borders on the canvas) but the blue highlighted borders on the canvas would go away.

jkeck:
... and add hover behavior to highlight individual annotations.

Closes #2383 

newAnnoBehavior


src/containers/WindowSideBarAnnotationsPanel.js
```diff
@@ -28,6 +29,7 @@ function getIdAndContentOfResources(resources) {
  * @private
  */
 const mapStateToProps = (state, { windowId }) => ({
+  allAnnotationsAreHighlighted: getWindow(state, { windowId }).displayAllAnnotations,
```
jkeck: I _could_ change this state/config attribute name do `highlightAllAnnotations`.  Not sure if it matters.

src/components/OpenSeadragonViewer.js
```diff
@@ -11,6 +11,31 @@ import CanvasWorld from '../lib/CanvasWorld';
  * and rendering OSD.
  */
 export class OpenSeadragonViewer extends Component {
+  /**
```
mejackreed: Did this method get moved around in the class? Was this due to a linting suggestion?

jkeck: Yeah, since we're passing `currentAnnotations` as a parameter to the function (instead of using the annotations props) so that it can be used for different types (selected vs. highlighted) of annotations it's no longer using `this` so it needs to be a `static` function (which apparently are required to come before the constructor).

mejackreed: oh wow.. got it thanks.

jkeck:
WRT the odd canvas behavior, I think this is a pre-existing issue that maybe is just more noticeable w/ the annotations now changing state on mouseEnter/mouseLeave of annotation elements.  Check these out (which are all happening in master right now).  I suppose we should create an issue about investigating that.

### Clicking Menus
weirdScroll

### Selecting Annotations
weirdAnnoSelection

PR #2399, opened 2019-04-05
Remove viewport debouncing; we're losing updates if the viewer re-ren…

cbeer:
…ders before the debounce.

This is most obvious if you:

- focus on a window
- initiate a pan/zoom
- switch focus to a different window

You'll see the first window revert back to the last-stored state.

### DIFF
--- a/__tests__/src/actions/canvas.test.js
+++ b/__tests__/src/actions/canvas.test.js
@@ -1,8 +1,6 @@
 import * as actions from '../../../src/state/actions';
 import ActionTypes from '../../../src/state/actions/action-types';
 
-const debounceTime = 100;
-
 describe('canvas actions', () => {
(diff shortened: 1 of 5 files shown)


mejackreed:
I'm noticing a sluggishness with zooming after removing this in Chrome. Are you not seeing that?

cbeer:
Only when I had the developer tools open? Still, sluggishness seems preferable to brokenness.

christopher-johnson:
There some issues with the OpenSeadragon Component.  Check `annotationsMatch `in `componentDidUpdate`,  canvasOverlayUpdate is called on every update, even when there are no annotations.

mejackreed:
I'm now noticing after #2394, the highlighting added there is no longer working. Happy to pair up to see if we can resolve.

Issue #2283, opened 2019-03-21
Refactoring observations about OpenSeadragonViewer.js, for fixing bugs and aiding maintainability


aeschylus:
We're already experiencing some OSD-related bugs (#2281, #2280, #2279, #2120 ), and adding features that pertain to annotation rendering have proven difficult and the solutions hard to follow/mimic for future contributors, as well as difficult for us to extend and build on. For maintainability and future development, I want to get to the root cause of some of these problems and create a clear seam between OSD and our app, with a pattern for how to interact with OSD from redux. These are some initial thoughts and questions to get the discussion started.

## Context 
The effectiveness of redux state management depends upon a functional, stateless rendering system. React provides this for most of our interface elements. In this paradigm, the rendering pipeline is typically:

**the full state is yielded from redux => react diffs the props => react makes the visual changes in elements that reference the props.**

OpenSeadragon is not a stateless rendering system like react, and we haven't been able to rely on this pattern to implement features. Therefore, currently, some state lives in openseadragon, some lives in redux, some lives in both places, some lives as properties on the component class, and some lives as code in the form of promises and sentinel statements - we are mutating state across paradigms, which is causing bugs and making it hard to reason about the implementation of new features. I would like to create a clear boundary between these paradigms, with patterns describing which things should be put on which sides of this boundary line, in such a way that it is easy to see where and how we would drop in another viewer. This mixing paradigms was a problem in Mirador 2 as well, so I want us to try to head it off at the pass if possible.

### What paradigms are we talking here?
There are 4 state-management paradigms attempting to coexist and interact here, and we need names for them.

1. **Immutable Rendering** - This is the system used in the react render function. This function is run every single time any that it depends on changes. It doesn't need to know the _type_ of change that occurred because it will render exactly what the data it receives tells it to render, and will always create the same output given the same input.
1. **OO API usage** - This is the system used in some places where we update openSeadragon (`fitBounds`, `addTileSource`). It lets openSeadragon decide how to update its internal state and what side effects to produce as a result of our method call. Calling the same method with the same inputs _will **NOT**_ produce the same result each time unless OSDs internal state is also the same when it is called. We therefore cannot guarantee the outcomes of this behaviour unless we know what the internal state is. There is another aspect of this paradigm which is using constructors to do initialisation and then accessors to make changes to properties. Because these aspects rely on the execution context of the object, in is sometimes hard to know when certain properties will or will not be defined, especially when object initialisations occur inside of react lifecycle methods.
[...]

Issue #2280, opened 2019-03-21
Zoom is incorrect (sometimes) on window instantiation.


aeschylus:
<img width="405" alt="Screenshot 2019-03-21 at 11 12 09 AM" src="https://user-images.githubusercontent.com/145874/54776541-9a436e00-4bcd-11e9-8ba7-d3fd9722f8c2.png">
<img width="415" alt="Screenshot 2019-03-21 at 11 22 47 AM" src="https://user-images.githubusercontent.com/145874/54776543-9a436e00-4bcd-11e9-8923-0297fc12cf4b.png">

This is related to these two areas of the OSD component:
https://github.com/ProjectMirador/mirador/blob/07e1544f7273cc8ace844637057b338b94e3fba2/src/components/OpenSeadragonViewer.js#L52-L55
https://github.com/ProjectMirador/mirador/blob/07e1544f7273cc8ace844637057b338b94e3fba2/src/components/OpenSeadragonViewer.js#L80-L100

Essentially we are waiting to set the zoom until the tilesources are loaded with a promise, but not managing the state properly. There is an interaction between these viewer instance existence guard conditional statements and the promise resolutions that needs to be teased apart.

PR #2417, opened 2019-04-08
[WIP] Eliminates extra OSD rendering cycles.

mejackreed:
Rather than continuously update redux state that is also passed in as a
prop, this change only makes those redux updates on demand when
explictly called. This increases performance, reduces complexity and
hopefully still accomplishes our state goals. A plugin / or other
component should also be able to explicatly call this instance method if
needed.

Related to #2283 , #2410, #2399 , #2280

### DIFF
--- a/src/components/OpenSeadragonViewer.js
+++ b/src/components/OpenSeadragonViewer.js
@@ -49,6 +49,8 @@ export class OpenSeadragonViewer extends Component {
     this.updateCanvas = () => {};
     this.ref = React.createRef();
     this.onUpdateViewport = this.onUpdateViewport.bind(this);
+    this.zoomIn = this.zoomIn.bind(this);
+    this.zoomOut = this.zoomOut.bind(this);
     this.zoomToWorld = this.zoomToWorld.bind(this);
   }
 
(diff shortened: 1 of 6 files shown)


src/components/OpenSeadragonViewer.js
```diff
@@ -73,7 +72,6 @@ export class OpenSeadragonViewer extends Component {
 
     this.osdCanvasOverlay = new OpenSeadragonCanvasOverlay(this.viewer);
     this.viewer.addHandler('update-viewport', this.onUpdateViewport);
-    this.viewer.addHandler('viewport-change', this.onViewportChange);
```
cbeer: I don't think `update-viewport` covers all the cases where we need to collect data. At least, when I toggled back and forth between two different windows, I could pretty reliably get the viewer to jump around.

mejackreed: I've pushed up an additional commit here that I think addresses the problem. Let me know what you think.

src/components/OpenSeadragonViewer.js
```diff
@@ -125,23 +123,13 @@ export class OpenSeadragonViewer extends Component {
           this.zoomToWorld();
         }
       });
-    } else if (viewer) {
```
cbeer: Without this, the zoom controls will break.

mejackreed:
Closing in favor of #2430

PR #2422, opened 2019-04-08
Improve how annotation matching in the OSDViewer component works …

jkeck:
… (by using isEqual on resource ID array comparison)

This is to address a `TODO` I left behind in some of the annotation work I did last week.  This should hopefully help cut down on some additional superfluous canvas redraws.

### DIFF
--- a/src/components/OpenSeadragonViewer.js
+++ b/src/components/OpenSeadragonViewer.js
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import Paper from '@material-ui/core/Paper';
+import isEqual from 'lodash/isEqual';
 import OpenSeadragon from 'openseadragon';
 import ns from '../config/css-ns';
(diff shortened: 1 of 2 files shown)


src/components/OpenSeadragonViewer.js
```diff
@@ -105,12 +103,8 @@ export class OpenSeadragonViewer extends Component {
         this.osdCanvasOverlay.clear();
         this.osdCanvasOverlay.resize();
         this.osdCanvasOverlay.canvasUpdate(() => {
-          if (highlightsUpdated) {
-            this.annotationsToContext(highlightedAnnotations, '#00BFFF');
-          }
-          if (selectionsUpdated) {
```
jkeck: These `if` statements were the cause of the odd behavior I was seeing before w/ the usage of `isEqual` in `annotationsMatch`.  We need to redraw all annotations every time we get here since we clear the canvas overly above.

PR #2425, opened 2019-04-09
Add a (Mirador window) responsive divider between the zoom and page controls

### LINKED ISSUE #2167: Add responsive separator behavior for the canvas/zoom controls panel
### Responsive behavior
- [ ] If the canvas is narrower than can fit the entire row of zoom/nav buttons, they should float to two lines:

responsive

- [ ] Note that when they go to two lines, we should not display the vertical divider between the two sets of buttons.

jkeck:
Closes #2167

### DIFF
--- a/__tests__/src/components/WindowCanvasNavigationControls.test.js
+++ b/__tests__/src/components/WindowCanvasNavigationControls.test.js
@@ -37,4 +37,15 @@ describe('WindowCanvasNavigationControls', () => {
     wrapper = createWrapper({ visible: false });
     expect(wrapper.matchesElement(<></>)).toBe(true);
   });
+
+  it('updates the canvasNavWidth state when the visibility changes', () => {
+    wrapper = createWrapper({ visible: false });
+    wrapper.instance().canvasNav = { current: { offsetWidth: 300 } }; // fake the ref
+
+    expect(wrapper.state('canvasNavWidth')).toEqual(null);
+
+    wrapper.setProps({ visible: true });
+
+    expect(wrapper.state('canvasNavWidth')).toEqual(300);
+  });
 });
--- a/src/components/WindowCanvasNavigationControls.js
+++ b/src/components/WindowCanvasNavigationControls.js
@@ -1,5 +1,6 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
+import classNames from 'classnames';
 import ZoomControls from '../containers/ZoomControls';
 import ViewerInfo from '../containers/ViewerInfo';
 import ViewerNavigation from '../containers/ViewerNavigation';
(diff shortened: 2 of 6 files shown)


src/components/WindowCanvasNavigationControls.js
```diff
@@ -9,6 +10,62 @@ import ns from '../config/css-ns';
...
+    });
+  }
+
+  /**
+   * Determine if canvasNavControls are stacked (based on a hard-coded width)
+  */
+  canvasNavControlsAreStacked() {
+    const { canvasNavWidth } = this.state;
+
+    return (canvasNavWidth && canvasNavWidth <= 253);
```
jkeck: I'm a bit concerned that this may be fragile.  I wanted to check the width of the container against the width of the zoom controls element (to determine if they were stacked or not) but because of how flexbox works the elements widths are not reliable (they only take up the full width when in the column direction).  I ended up needing to track the width of the container, so it didn't make sense to have the addl. code and potential for finicky behavior of also tracking the width of the child controls.  Happy to pair up if folks think we should explore alternate solutions (e.g. abandoning flexbox all-together for these components).

src/containers/ZoomControls.js
```diff
@@ -33,13 +33,20 @@ const mapDispatchToProps = { updateViewport: actions.updateViewport };
  * ListItem: {paddingBottom: number, paddingTop: number}}}
  */
 const styles = theme => ({
+  divider: {
+    borderRight: '1px solid #808080',
```
jkeck: I couldn't find a matching color in the theme and the `divider` color in the palette is way too light.  Is there something we should be using here from the theme?

src/containers/ZoomControls.js
```diff
@@ -33,13 +33,20 @@ const mapDispatchToProps = { updateViewport: actions.updateViewport };
...
+    display: 'inline-block',
+    height: '24px',
+    margin: '12px 6px',
```
jkeck: Wondering if we should be using the theme's spacing here.  It's 8px so it might be

```js
margin: `${theme.spacing.unit * 1.5}px ${theme.spacing.unit}px`
```

Which would be equivalent to 

```css
margin: '12px 8px'
```

src/components/WindowCanvasNavigationControls.js
jkeck: Okay, I took a spike at using `AutoSizer` and I couldn't get it to quite work the way that I needed.  Happy to pair up on trying to implement something (to see if my understanding of how to use `AutoSizer` is off).

src/containers/ZoomControls.js
mejackreed: I think we may just need to come back to this when we do more theme things.

__tests__/src/components/WindowCanvasNavigationControls.test.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { shallow } from 'enzyme';
+import { mount, shallow } from 'enzyme';
```
mejackreed: Is this `mount` still needed?

Issue #2355, opened 2019-04-01
Can we disable drag-workspace when a menu is open? (Elastic mode)


jvine:
When a menu is open in Elastic mode, the entire area outside the menu is draggable and moves the workspace. So it's possible to temporarily create a situation where the menu is disconnected from its parent. Once you stop the drag, the menu closes. It's not actively destructive, but it's disconcerting.

<img width="800" alt="Screen Shot 2019-04-01 at 10 43 57 AM (2)" src="https://user-images.githubusercontent.com/6945691/55348086-3fcacd00-546b-11e9-89b2-94c008d954f1.png">
<img width="800" alt="Screen Shot 2019-04-01 at 10 43 38 AM (2)" src="https://user-images.githubusercontent.com/6945691/55348088-3fcacd00-546b-11e9-93cd-9612f9262943.png">

glefi:
react-rnd provides the `draggingEnabled`property to enable or disable drag and drop. Unfortunately this property is ignored when a selection list is opened. In this case, dragging is available all opened windows are moved at the same time. I currently see no way to solve this problem.

glefi:
@aeschylus I've tried to to set `draggingEnabled` for the window as soon as the the `<Select>` component opens, but unfortunately this doesn't seem to help. Because immediatly after the selection list appears, you can drag the windows anyway.

PR #2428, opened 2019-04-09
#2355: adds toggling of the disableDragging property

glefi:
@jvine I've disabled the dragging for the canvas panel and the window top menu.


jvine:
The behaviour looks correct to me, @glefi. Can't comment on the technical side.

mejackreed:
Is there a way that the git history could be cleaned up here? Maybe just a cherry-pick of the commits needed?

Issue #2376, opened 2019-04-03
Reorganize workspace menu


jvine:
We've been struggling with the organization of the workspace menu as development has progressed with different patterns than initially planned. The Handschriften Portal has made a change that we like, in exposing the open windows list as a separate feature; 

<img width="201" alt="Screen Shot 2019-04-03 at 5 52 51 PM" src="https://user-images.githubusercontent.com/6945691/55522325-59217400-5639-11e9-94c6-244c1e603fef.png">

and we want to combine that idea with an earlier design draft to reorganize the appbar buttons and menus. We think this will improve usability overall, eliminate re-use of the hamburger menu icon, and clarify what is/is not available when the load window is open. 

The goal is to break the single menu out into three separate menus, like so:

- Jump to window (icon shows count of open windows)
  - windows list
- Settings 
  - Show zoom controls
  - Workspace type
  - Language
  - Theme
  - additional settings plugins
- Workspace options
  - Export workspace
  - Import workspace
  - additional workspace plugins
- Full screen 

- [x] **Open windows**
<img width="456" alt="appbar open windows" src="https://user-images.githubusercontent.com/6945691/55522239-ee703880-5638-11e9-95e9-5a42d8e92e51.png">

- [ ] **Settings**
<img width="265" alt="appbar settings" src="https://user-images.githubusercontent.com/6945691/55522241-ef08cf00-5638-11e9-97e9-e7d350af8720.png">

- [ ] **Workspace options**
<img width="285" alt="appbar more" src="https://user-images.githubusercontent.com/6945691/55522240-ee703880-5638-11e9-9ddf-2609f0ed175f.png">

When the load window is open, only the Settings and Full screen buttons are enabled. Within the Settings menu, Show zoom controls is disabled. 

- [ ] **Load window**
<img width="276" alt="appbar load window" src="https://user-images.githubusercontent.com/6945691/55522456-ebc21300-5639-11e9-8ad3-9ecc160483b1.png">

(There is an argument for making the currently-open windows list available when the load window is open, so you can compare what you're adding with what you already have; but its not clear what the menu behaviour would be. Selecting an open window would have no effect in the load window itself; and it shouldn't change the focus behind the load window either.)

ggeisler:
@jvine In the headings above the mockups, aren't Settings and Workspace options reversed (Settings heading shows workspace options and vice versa)?

jvine:
^ Yep, fixed.

PR #2449, opened 2019-04-10
Pull WindowList out of WorkspaceMenu and into its own button in the Workspace Sidebar

jkeck:
Part of #2376 

Submitting this partial PR because I have gotten distracted on fixing up some previous work.  Will continue to make other changes in additional PRs, but this felt like it could stand on its own (or un-assign)

WindowListMenuButton


jkeck:
I _think_ the coverage failure here is due to removed/moved code.

PR #2450, opened 2019-04-10
Extract elastic layout properties to a separate redux slice

cbeer:
Both @jkeck and I have been... surprised... about the width/height properties of window, when they only apply to elastic mode. Hopefully moving them to a new slice will make this a little more approachable.


src/state/reducers/elasticLayout.js
```diff
@@ -0,0 +1,32 @@
+import { updateIn, merge } from 'immutable';
+import ActionTypes from '../actions/action-types';
+
+/**
+ * windowsReducer
```
mejackreed: `elasticLayoutReducer`?

cbeer: 👍

__tests__/src/actions/window.test.js
```diff
@@ -12,14 +12,18 @@ describe('window actions', () => {
 
       const mockState = {
         companionWindows: {},
-        windows: {
+        elasticLayout: {
           window: {
```
mejackreed: Should there be another `windows` layer here?

cbeer: I don't think so; `window` is just a terrible name for the window id there.

cbeer:
You're right; I was missing a commit from #2454. Juggling branches gets confusing.

PR #2456, opened 2019-04-10
Move Paper wrapper to be consistent with component it is styling

mejackreed:
When working on a plugin where I wanted to wrap `WindowCanvasNavigationControls` I noticed that the `Paper` componenet styling them was actually part of another component. I wasn't sure if there was a reason for this, but I think this approach works in all cases.

### DIFF
--- a/src/components/WindowCanvasNavigationControls.js
+++ b/src/components/WindowCanvasNavigationControls.js
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
+import Paper from '@material-ui/core/Paper';
 import ZoomControls from '../containers/ZoomControls';
 import ViewerInfo from '../containers/ViewerInfo';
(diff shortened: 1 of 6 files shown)


src/components/WindowCanvasNavigationControls.js
```diff
@@ -22,35 +23,39 @@ export class WindowCanvasNavigationControls extends Component {
...
-      >
-        <ZoomControls
-          displayDivider={!this.canvasNavControlsAreStacked()}
-          windowId={window.id}
-          zoomToWorld={zoomToWorld}
-        />
-        <ViewerNavigation window={window} />
-        <ViewerInfo windowId={window.id} />
-      </div>
+      <Paper square className={classes.controls} elevation={0}>
```
cbeer: Do we still need the `div` too, or can those classes go onto `Paper`?

mejackreed: Nice catch. I think we can 🔥 .. pushed up

PR #2457, opened 2019-04-10
Updates propType expectation to handle a single node or multiples 

mejackreed:
(getting an error without this)

### DIFF
--- a/src/components/CollapsibleSection.js
+++ b/src/components/CollapsibleSection.js
@@ -62,7 +62,7 @@ export class CollapsibleSection extends Component {
 }
 
 CollapsibleSection.propTypes = {
-  children: PropTypes.element.isRequired,
+  children: PropTypes.node.isRequired,
   classes: PropTypes.object.isRequired, // eslint-disable-line react/forbid-prop-types
   id: PropTypes.string.isRequired,
   label: PropTypes.string.isRequired,


src/components/CollapsibleSection.js
```diff
@@ -62,7 +62,10 @@ export class CollapsibleSection extends Component {
 }
 
 CollapsibleSection.propTypes = {
-  children: PropTypes.element.isRequired,
+  children: PropTypes.oneOfType([
+    PropTypes.arrayOf(PropTypes.node),
```
cbeer: I forget; is this maybe equivalent to `PropTypes.element`?

jkeck:
I'm curious what error you're seeing w/ this?  I'm not seeing any errors, but perhaps I'm not looking at the right kind of thing to trigger that.

mejackreed:
@jkeck I'm seeing this on master on initial page load:

Screen Shot 2019-04-10 at 5 45 38 PM

jkeck:
:+1: Ah, I see, it's not an error, it's a warning (which is why I wasn't seeing it in production)

src/components/CollapsibleSection.js
mejackreed: Yeah I guess so.. pushed up that change.

Issue #2379, opened 2019-04-03
Enable user-controlled panel resizing [placeholder]


ggeisler:
The user should be able to resize the width (in vertical mode) or height (in horizontal mode) of companion windows, as well as the side panel, within some min and max range. Details TBD.

PR #2466, opened 2019-04-11
Make companion windows resizeable; a stab at #2379

cbeer:

### DIFF
--- a/src/components/CompanionWindow.js
+++ b/src/components/CompanionWindow.js
@@ -6,6 +6,7 @@ import MoveIcon from '@material-ui/icons/DragIndicatorSharp';
 import Paper from '@material-ui/core/Paper';
 import Typography from '@material-ui/core/Typography';
 import Toolbar from '@material-ui/core/Toolbar';
+import { Rnd } from 'react-rnd';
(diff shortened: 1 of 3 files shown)


cbeer:
Are those blockers to merging? I think setting the minimums will be easy enough, but I don't know about setting maximums (particularly across the different view modes..)

ggeisler:
I think we can live with the lack of the max sizing restriction for the time being. Panel resizing is a really nice feature so it seems worth living with the likely infrequent case of the user making a mess by sizing too large to get it.

src/components/CompanionWindow.js
```diff
@@ -34,55 +61,69 @@ export class CompanionWindow extends Component {
...
-            position === 'left'
-              ? updateCompanionWindow
-                && (
-                  <MiradorMenuButton
-                    aria-label={t('openInCompanionWindow')}
-                    onClick={() => { updateCompanionWindow(windowId, id, { position: 'right' }); }}
-                  >
-                    <OpenInNewIcon />
-                  </MiradorMenuButton>
+        <Rnd
```
mejackreed: Should we add a quick test for this thing?

PR #2468, opened 2019-04-11
Homepage is always single-valued

cbeer:


mejackreed:
Looks like there may be a multiple valued homepage coming. https://github.com/IIIF/trc/issues/4

Should we hold off?

PR #2472, opened 2019-04-11
Pull Import/Export options out to a separate WorkspaceOptionsMenu component

### LINKED ISSUE #2376: Reorganize workspace menu

jkeck:
Closes #2376 

NewMenuLayout


src/containers/WorkspaceOptionsMenu.js
```diff
@@ -0,0 +1,18 @@
...
+  containerId: getContainerId(state),
+});
+
+// containerId: getContainerId(state),/
+const enhance = compose(
+  withTranslation(),
+  connect(mapStateToProps, null),
+);
+
+export default enhance(WorkspaceOptionsMenu);
```
jkeck: One thing to call out here is that the ticket makes explicit where in the menu plugins can add options.  I'm not 100% sure what the state of our plugin system is (`withPlugins`, `PluginHook`/`PluginComponents`, etc), so any guidance on what I should be doing wrt hooking this menu (and presumably the other one) with the plugin system would be greatly appreciated.

PR #2515, opened 2019-04-15
Misc performance optimizations

cbeer:


src/components/CaptionedCanvasThumbnail.js
```diff
@@ -24,9 +24,6 @@ export class CaptionedCanvasThumbnail extends Component {
           }
           isValid={manifestoCanvas.hasValidDimensions}
           maxHeight={height}
-          style={{
-            maxWidth: `${Math.ceil(height * manifestoCanvas.aspectRatio)}px`,
```
mejackreed: I think the case here is for manifests that canvas sizes don't reflect the actual image sizes, while in the right side.

See: https://media.nga.gov/public/manifests/nga_highlights.json will produce a horizontal scroll bar while in bookview on right side. Maybe there is a better way?

mejackreed: I wonder if we could make this a constant instead?

PR #2516, opened 2019-04-15
Extract i18n lint task for translators

camillevilla:
¯\\_(ツ)_/¯ Thought it might be nice to extract this command for potential translators and add it to the i18n documentation here:
https://github.com/ProjectMirador/mirador/wiki/M3-Internationalization-(i18n)#adding-a-new-language

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,8 @@
   ],
   "scripts": {
     "clean": "rm -rf ./dist",
-    "lint": "node_modules/.bin/eslint ./ && node_modules/.bin/sass-lint -v ./src/styles/**/* && node ./scripts/i18n-lint.js",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -9,6 +9,7 @@
   "scripts": {
     "clean": "rm -rf ./dist",
     "lint": "node_modules/.bin/eslint ./ && node_modules/.bin/sass-lint -v ./src/styles/**/* && node ./scripts/i18n-lint.js",
+    "lint:translations": "node ./scripts/i18n-lint.js",
```
cbeer: Should `lint` call this instead of invoking it directly? 🤷‍♂️

Issue #2518, opened 2019-04-15
Update theming (primary/secondary colors and dark mode)


ggeisler:
Below are just some baby steps in improving the theming. There are still some challenges related to color contrast for accessibility and getting things correctly set up for institutional theming (going beyond just "light' and "dark" themes). But we can iterate after these changes get in.

### Update the theme settings for primary and secondary colors
- [x] Redefine the current `secondary.main` color (`#1967d2`) to be the `primary.main` color
- [x] Define a new color (`#ffa224`) to `secondary.main`
- [x] Update all `secondary.main` color assignments to use `primary.main` (so the end result no visible change to the UI; the FAB, current window top border, controls, etc. are still blue)
- [x] Assign the new `secondary.main` color to the notification dot currently used to indicate a canvas has annotations. So the dot will change from red to orange.

After the changes above are made, things should look similar to this:

<img width="539" alt="Screen Shot 2019-04-15 at 3 32 45 PM" src="https://user-images.githubusercontent.com/101482/56169696-d29f5780-5f93-11e9-8fa5-354ce04a7866.png">

### Give dark mode a different primary color
- [x] Define a different primary color in dark mode: `#4db6ac`

After making the changes above, when you change to the dark theme, the FAB, side panel icon menu highlight, and form controls should change to a teal color:

<img width="539" alt="Screen Shot 2019-04-15 at 3 38 30 PM" src="https://user-images.githubusercontent.com/101482/56169899-8acd0000-5f94-11e9-96d5-c99593990a7e.png">

### Update component assignments to work better in dark mode
- [x] Update the companion window panel header background so that is remains the current color in light mode, but changes to be `grey[700]` (`#616161`) in dark mode.
- [x] Update the workspace background so that is remains the current color in light mode, but changes to be `grey[700]` (`#616161`) in dark mode.

After those changes, in dark mode the workspace background and window panel header background should look like this:

<img width="539" alt="Screen Shot 2019-04-15 at 3 37 49 PM" src="https://user-images.githubusercontent.com/101482/56169878-72f57c00-5f94-11e9-84b4-6ef9a538a656.png">

PR #2520, opened 2019-04-15
Theme updates

mejackreed:
Theme updates based off of design in #2518 .

Intentionally does not "Give dark mode a different primary color". Not sure if thats possible with our approach.


src/containers/Workspace.js
```diff
@@ -20,8 +21,26 @@ const mapStateToProps = state => (
   }
 );
 
+
+/**
+ * @param theme
+ */
+const styles = theme => ({
+  workspaceViewport: {
+    background: (theme.palette.type === 'light') ? theme.palette.grey[100] : theme.palette.grey[700],
```
cbeer: Instead of calculating this in a couple places, maybe we can explicitly add the value to the theme?

mejackreed: Can you say more about that? I'm not sure how to do that without still having to check what theme is currently selected

cbeer: I'm thinking something like akin to how we're using `theme.palette.darkened` or `.lightened` currently:

- https://github.com/ProjectMirador/mirador/blob/master/src/config/settings.js
- https://github.com/ProjectMirador/mirador/blob/master/src/containers/CompanionArea.js#L41

mejackreed: Ah.. I was hoping we could eventually remove those at some point (I'm not a fan of them.. )

cbeer: Yeah, I'm not sure they're entirely idiomatic, but it has the benefit of being externally theme-able instead of hardcoding colors in our components.

mejackreed: true.. what if we added something like `hue` which would allow adopters to pass a mui color HUE that was then used? https://material-ui.com/style/color/

```json
    palette: {
      type: 'light', // dark also available
      hue: 'grey'
      primary: {
        main: '#1967d2',
      },
      secondary: {
        main: '#ffa224',
      },
      grey: {} // must be defined for `hue`
```

That way an adopter could then use any color hue w/ varying shades

ggeisler:
@mejackreed It looks like we lost the current window top border, in both light and dark mode?

src/containers/Workspace.js
mejackreed: meh.. now I don't like that proposal. I'm going to do something similar to your first suggestion :)

ggeisler:
@mejackreed It looks like you made the form controls use the new secondary (orange) color, based on how the annotation toggle looks when turned on.

It does make sense that a theme would use primary for things like the FAB and secondary for form controls, but I don't think that was our plan. We still want form controls like the toggle to be blue (or whatever the primary color is), at least for now.

mejackreed:
Akk. thats an error on my part. I'll make the change @ggeisler

src/containers/Workspace.js
mejackreed: @cbeer I've updated using your suggestion

mejackreed:
@ggeisler I've updated this to specify that component to use primary. Be default it uses secondary.

PR #2523, opened 2019-04-16
reduce plugin rerenders using memo

mejackreed:
While monitoring http://127.0.0.1:4444/__tests__/integration/mirador/plugins/add.html I'm seeing a reduction of rerenders from 1319 -> 1230

### DIFF
--- a/__tests__/src/extend/withPlugins.test.js
+++ b/__tests__/src/extend/withPlugins.test.js
@@ -19,15 +19,17 @@ function createPluginHoc(pluginMap) {
 }
 
 describe('withPlugins', () => {
-  it('should return a function (normal function call)', () => {
-    expect(withPlugins('Target', Target)).toBeInstanceOf(Function);
+  it('should return a React.memo object (normal function call)', () => {
+    expect(withPlugins('Target', Target)).toBeInstanceOf(Object);
+    expect(withPlugins('Target', Target).type).toBeInstanceOf(Function);
   });
 
-  it('should return a function (curry function call)', () => {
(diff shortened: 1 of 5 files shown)


__tests__/src/extend/withPlugins.test.js
```diff
@@ -19,15 +19,17 @@ function createPluginHoc(pluginMap) {
...
 
-  it('should return a function (curry function call)', () => {
-    expect(withPlugins('Target')(Target)).toBeInstanceOf(Function);
+  it('should return a React.memo object (curry function call)', () => {
+    expect(withPlugins('Target')(Target)).toBeInstanceOf(Object);
+    expect(withPlugins('Target')(Target).type).toBeInstanceOf(Function);
   });
 
-  it('displayName prop of returned function is based on target name argument', () => {
+  xit('displayName prop of returned function is based on target name argument', () => {
```
mejackreed: I couldn't figure out how to get this to return properly.

charbugs:
Can we hold on merging this? I'm currently about a refactoring that seems promising to me in terms of performance.

mejackreed:
Of course.

__tests__/src/extend/withPlugins.test.js
charbugs: There are also other tests failing in withPlugin. I already run into similar issues with that memo function. The problem is that the memo function does not return a component function but a react element object. And this object is somewhat opaque and not described in the docs.

What about putting the PluginHoc at top level and then doing a name export for testing purpose?

PR #2530, opened 2019-04-16
[WIP] Refactor plugin system for performance

charbugs:
The idea is to decorate components with a plugin HOC only if there extists a plugin for that component. That is, there should be no `WithPlugin(Component)` at all if there are no plugins. The only problem is: it doesn't work! 

The `withPlugins` function expects that the plugins are already processed and retrievable via `getPlugins`. But this is not the case because in `MiradowViewer` the `App` component (and therefore the entire component tree including the containers where `withPlugins` is exectuted) is imported BEFORE the plugins are processed. :disappointed: 

One idea is to import `App` dynamically. Any experience with that or other ideas?


mejackreed:
@aeschylus you mentioned some barriers you noticed the other day about something you saw w/ plugins. Was it that previously there was no way to have other entry points beyond `MiradorViewer`?

Issue #2533, opened 2019-04-16
Issues with the edge of the world in elastic mode


jvine:
**Part 1** 
It turns out the elastic mode workspace has a boundary on all sides at which point windows cannot move further, but the user can continue to drag the surface. 

The boundary is invisible on the top and left. On the right and bottom it's marked by scrollbars. The scrollbars are not functional, just present. Only visible to those of us who have scrollbars set to show always. 

- [ ] workspace is set to `overflow:scroll`; this should be `auto` or removed.

**Part 2**
It's actually possible to drag the workspace entirely outside of the viewport, and/or drag the minimap viewport outside the actual viewport. There's no user need that this supports, and once you've done it, you can only get it back using Jump to window.

<img width="993" alt="viewport edge" src="https://user-images.githubusercontent.com/6945691/56249713-9b977780-6061-11e9-9b09-7f17f28bfaa1.png">

1. Is it possible to constrain the minimap draggable viewport to the minimap area? 
2. Is it possible to prevent the user from dragging past the real edge of the workspace?

<img width="473" alt="viewport edge copy" src="https://user-images.githubusercontent.com/6945691/56249911-63446900-6062-11e9-9186-9864266dee7c.png">

cbeer:
The non-minimap parts of this ticket have been addressed, so I've edited the description to just include the minimap constraints.

marlo-longley:
Related: https://github.com/ProjectMirador/mirador/issues/2140
There is currently no minimap

marlo-longley:
Closing to reduce backlog — this issue describes a sub-feature of the elastic workspace minimap, which hasn't been implemented in v4. The parent feature is tracked in #2140. If the minimap is being built this can be reopened.

PR #2535, opened 2019-04-17
Make the elastic workspace seem infinite by continuously expanding th…

cbeer:
…e width + height when the window goes beyond the workspace border.

One approach to fixing #2533 (there's no problem at the edges if there are no edges!)


aeschylus:
Awesome. Works great.

PR #2539, opened 2019-04-17
Use the a config setting to define a default behavior for showing…

### LINKED ISSUE #2504: Add a setting to turn on zoomControls by default

jkeck:
…zoom controls.

Closes #2504

### DIFF
--- a/src/state/selectors/config.js
+++ b/src/state/selectors/config.js
@@ -20,8 +20,15 @@ export const getLanguagesFromConfigWithCurrent = createSelector(
 );
 
 export const getShowZoomControlsConfig = createSelector(
-  [state => state.workspace],
-  workspace => workspace.showZoomControls,
+  [
+    state => state.workspace,
+    getConfig,
+  ],
+  (workspace, config) => (
+    workspace.showZoomControls === undefined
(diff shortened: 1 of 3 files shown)


src/state/selectors/config.js
```diff
@@ -20,8 +20,15 @@ export const getLanguagesFromConfigWithCurrent = createSelector(
...
-  [state => state.workspace],
-  workspace => workspace.showZoomControls,
+  [
+    state => state.workspace,
+    getConfig,
+  ],
+  (workspace, config) => (
+    workspace.showZoomControls === undefined
+      ? (config.workspace.showZoomControls)
+      : workspace.showZoomControls
```
jkeck: Not really pleased about doing this here, however our current method of setting defaults for the Workspace doesn't really lend itself to making use of config settings.

https://github.com/ProjectMirador/mirador/blob/8b56e610bfb241f6d774299eab25857dfdcba9ef/src/state/reducers/workspace.js#L8-L18

I could make this ticket/PR about fixing that, or I can create another issue for that and reference this code for when that refactoring happens.  Happy to go either way.

PR #2540, opened 2019-04-17
Disconnect plugin provider from redux context for performance reasons

charbugs:
@mejackreed Does this reduce rerenderings?


mejackreed:
I'm not seeing rerender reductions on the `plugins/add.html` page with this change.

I'm cherry-picking 37703ca90140212ceb82948efb2329873c0be405 commit on top of this and then uncommenting this line to include all the components: https://github.com/ProjectMirador/mirador/compare/why-did?expand=1#diff-1fdf421c05c1140f6d71444ea2b27638R12

cbeer:
Effectively done with #3061

PR #2541, opened 2019-04-17
closes #2404 Disable "Move to bottom" button on companion windows when canvas height is less than 300px

### LINKED ISSUE #2404: Disable "Move to bottom" button on companion windows when canvas height is less than 300px
We want to prevent a situation where the user cannot interact with the window sidebar because she has created more horizontal companion windows than can fit in the window while still allowing enough visibility of the sidebar to be able to interact with it. For example, in the scenario below the user can't see any of the sidebar panel content, nor can she scroll the side panel content:

Screen Shot 2019-04-05 at 11 59 57 AM

---

To prevent this, we should disable the "Move to bottom" button in all vertical companion windows when the canvas height is less than 300px.

Screen Shot 2019-04-05 at 11 52 39 AM

---

If the user closes a companion window, moves a horizontal companion window back to the right (vertical), or widens the browsers window vertically such that the canvas height becomes 300px or taller, then the "Move to bottom" button in all vertical companion windows should be re-enabled.

I don't know that we've disabled buttons anywhere in the app yet so there might not be an established pattern for this. But this [Material Design page](https://material.io/develop/web/components/buttons/icon-buttons/) does have a section called "Disabled" where they indicate you can add a `disabled` attribute to a button to prevent the user from interacting with it. Hopefully that will work for us.

glefi:


jkeck (dismissed):
Thanks for tackling this @glefi.  I'm not able to get the desired result in the Netlify deploy.  If I'm understanding the goal of #2404 correctly, then I should not be able to get into this scenario:

Short Window

The window itself here is shorter than 200px so the canvas area is definitely smaller than that, so I believe this button should be either disabled or not present (but I can move the companion window from that height).

glefi:
Thanks for your comment @jkeck The latest commit fixes the disabling of the button, in case the canvas's size is lower than 300px.

mejackreed (changes requested):
This is working nicely when the companion window is on the right side. I did however notice a crash when the companion window is on the bottom and < 300 px.

`Uncaught Invariant Violation: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.`

glefi:
@mejackreed thanks for pointing that out. the latest commit should fix the problem.

src/containers/CompanionWindow.js
```diff
@@ -14,13 +17,16 @@ import { CompanionWindow } from '../components/CompanionWindow';
  * @private
  */
 const mapStateToProps = (state, { id, windowId }) => {
+  const window = getWindow(state, { windowId });
   const companionWindow = getCompanionWindow(state, { companionWindowId: id });
+  const showMoveToBottom = !window.height || window.height > 300 || companionWindow.position !== 'right';
```
jkeck: @glefi I believe the desired behavior here is to `disable` the button, not necessarily remove it all-together.  I think this should be a value that would be passed to the appropriate `MiradorMenuButton` component as the `disabled` prop (which would also indicate a potential name/logic change).

jkeck (dismissed):

marlo-longley:
Stale now, but a good reference for https://github.com/ProjectMirador/mirador/issues/2404 in 2026. Thanks for the work here.

PR #2543, opened 2019-04-17
Use consistent styles for section borders

cbeer:


src/containers/WindowSideBarInfoPanel.js
```diff
@@ -34,7 +34,7 @@ const mapDispatchToProps = (dispatch, { windowId, id }) => ({
  */
 const styles = theme => ({
   section: {
-    borderBottom: '.5px solid rgba(0,0,0,0.25)',
+    borderBottom: `.5px solid ${theme.palette.divider}`,
```
jkeck: I asked @ggeisler very specifically about this when implementing this feature and he said to not use the divider color.  I'll leave it up to him to 👍/ :-1: because this is what I did in the first place.

cbeer: 🆒; it strikes me as a little strange that some of the borders were that and some weren't. I don't care what they are, but consistency would be nice.

ggeisler: Yeah, I think that was because @jvine and I thought the default wasn't dark enough to delineate the sections as strongly as we thought they should be delineated. The default uses 0.12 versus our recommendation of 0.25.

I personally prefer our custom version but also see the value of using the default and being consistent with the un-customized divider being used elsewhere in the app. So I'd be amenable to changing it. @jvine do you have thoughts on sticking with our customized version versus going with the divider default?

### Customized at 0.25



### Default at 0.12

ggeisler: I can see how from a code point of view it might seem strange to use different colors for the divider in different places, but from a visual design point of view the dividers are used for different purposes/contexts around the app, so I don't think it's necessarily unreasonable for the color to vary with the context.

jvine: I don't remember that discussion but prefer the lighter (default) colour.

PR #2546, opened 2019-04-17
Sets paged view by default for manifests and sequences

### LINKED ISSUE #1440: No recognition of "paged" viewingHint
When loading a manifest it should read the manifest's viewingHint and do "book" display if it sees "paged"

### LINKED ISSUE #1443: Individuals viewingHint when in BookView
Though it displays the `individuals` `viewingHint` correctly in BookView, a user can't annotate or use image manipulation tools unless in ImageView.  Until ImageView and BookView are merged into a single widget, `individuals` should really switch to ImageView rather that persist in BookView.

mejackreed:
For IIIF manifests and sequence that set a viewingHint to paged,
Mirador3 will now automatically put the view into "book" view.
Fixes #1443
Fixes #1440

A user action here will always override the manifest behavior.

paged by default

### DIFF
--- a/__tests__/fixtures/version-2/015.json
+++ b/__tests__/fixtures/version-2/015.json
@@ -1,0 +1,57 @@
+{
+  "@context": "http://iiif.io/api/presentation/2/context.json", 
+  "@id": "http://iiif.io/api/presentation/2.0/example/fixtures/15/manifest.json", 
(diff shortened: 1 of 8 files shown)


mejackreed:
Ah... nice catch on that. I've updated that with an additional commit.

PR #2555, opened 2019-04-17
Responsive updates to the Workspace Sidebar/Control Panel

### LINKED ISSUE #2528: Responsive updates to workspace app bar
At the `xs`(600px, I believe) breakpoint and below, we need to customize the workspace app bar to prevent this sort of ugliness:

#### Normal workspace
Screen Shot 2019-04-16 at 11 21 48 AM

---
#### Empty workspace
Screen Shot 2019-04-16 at 11 22 03 AM

### Responsive updates
In the `xs` breakpoint, we should:

- [x] Hide the Mirador logo

- [x] Hide the "Mirador" text (empty workspace version)

- [x] Use the normal version of the FAB, rather than the version with the "Start here" label, when using the empty workspace version of the app bar

- [x] Keep FAB aligned-left, but align the other buttons right

This means the workspace app bar should look the same in both normal and empty workspace modes while in the `xs` breakpoint. End result would be similar to this:

Screen Shot 2019-04-16 at 11 22 27 AM

jkeck:
Closes #2528

### DIFF
--- a/__tests__/src/components/WorkspaceAddButton.test.js
+++ b/__tests__/src/components/WorkspaceAddButton.test.js
@@ -13,6 +13,8 @@ function createWrapper(props) {
       classes={{}}
       setWorkspaceAddVisibility={() => {}}
       t={str => str}
+      useExtendedFab
+
       {...props}
     />,
   );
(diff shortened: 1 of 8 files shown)


ggeisler:
@jkeck Did you look at trying to do the last item in the #2528 list:

> Keep FAB aligned-left, but align the other buttons right

I guess it is not critical if it poses technical challenges or complications. I was able to do it (as in the screenshots in the ticket) by changing the `div` containing the buttons to `flex-end` (or whatever the exact thing is to say you want the flex row right-aligned), and then making the FAB button itself left-aligned (again, I forget exactly how, but I think using the selector for aligning a specific flex item).

jkeck:
Yeah, that's why I ended up marking this as WIP.

ggeisler:
Oh, great, sorry, I glossed over that!

jkeck:
@ggeisler no worries.  I had put it up w/o the WIP before and when I went to go make all the boxes I realized I had missed the alignment issue.

I've pushed that up now, please take a look and let me know if it looks 👌

ggeisler:
@jkeck Nice, looks good to me.

PR #2557, opened 2019-04-17
 Add configuration for the default sidebar panel and open state

### LINKED ISSUE #2402: Enable configuration of default state of window side panel
Related to #2372.

Currently, the About/Metadata panel is the panel that is shown when a window side panel is initially opened. 

However, an implementor might instead want the Rights panel to be the default panel displayed, because they are especially concerned that users see the rights/attribution info for the object. 

Issues to resolve:
- It is possible that an object manifest not include any rights/attribution info, in which case we intend that the Rights panel not be represented in the sidebar nav. But if an implementor configures the Rights panel to be the default, what happens when a user opens an object with no rights/attribution info (and thus there is no Rights panel to display)?

- For the use case where the implementor wants to configure the Rights panel as the default because they are especially concerned that users see the rights/attribution info for the object, is it enough to make the Rights panel the default when the sidebar is explicitly opened by the user, or do we also need to enable the implementor to indicate that the sidebar for a newly-opened window should be **open** by default?

Windows currently open with the sidebar closed, so even if the Rights panel is the default, there is no guarantee that the user will see the rights/attribution info, because they have to deliberately choose to open the sidebar to see it.

cbeer:
Fixes #2402

### DIFF
--- a/src/config/settings.js
+++ b/src/config/settings.js
@@ -180,8 +180,10 @@ export default {
   window: {
     allowClose: true, // Configure if windows can be closed or not
     allowMaximize: true, // Configure if windows can be maximized or not
+    defaultSideBarPanel: 'info', // Configure which sidebar is selected by default
     defaultView: 'single',  // Configure which viewing mode (e.g. single, book, gallery) for windows to be opened in
     hideAnnotationsPanel: false, // Configure to hide the annotations panel in the WindowSideBarButtons
+    sideBarOpenByDefault: false, // Configure if the sidebar (and its content panel) is open by default
   },
   windows: [], // Array of windows to be open when mirador initializes (each object should at least provide a loadedManifest key with the value of the IIIF presentation manifest to load)
   thumbnailNavigation: {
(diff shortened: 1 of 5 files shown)


src/config/settings.js
```diff
@@ -180,8 +180,10 @@ export default {
   window: {
     allowClose: true, // Configure if windows can be closed or not
     allowMaximize: true, // Configure if windows can be maximized or not
+    defaultSideBarPanel: 'info', // Configure which sidebar is selected by default
```
camillevilla:
```suggestion
    defaultSideBarPanel: 'info', // Configure which sidebar is selected by default. Options: info, rights, canvas, annotations
```

It would be nice to list the available panel names here. Another idea would be linking to [CompanionWindowFactory](https://github.com/ProjectMirador/mirador/blob/287e2ce1938cda8430a4b887c191066fbe5f50a1/src/components/CompanionWindowFactory.js#L17-L29)

Useful follow-up action would be renaming AttributionPanel.js so that it matches the `WindowSidebar{__}Panel` naming convention.

cbeer: I'm trying to move us away from the `WindowSidebar{__}Panel` convention, it's way too wordy. I think we're now 1/2 and 1/2.

Issue #2479, opened 2019-04-12
Consistency of hover/focus/selected indicators


jvine:
## Definitions

Mouse hover indicates "this is an active thing you can click". Keyboard focus sometimes indicates "this is an active thing you can click"; sometimes actually triggers an action. We need to treat these separately.

**Hover**
- mouse is hovering over a control, _or keyboard focus is on the control but no action has been triggered_. e.g. icon buttons, menu options, index panel lists
- expressed as an altered background ("light")

**Focus/Select**
- mouse click or keyboard focus triggers a change in the canvas or panels. e.g. thumbnail nav, gallery view
- expressed as a different altered background ("dark"), or a coloured border, depending on context

**Active/Pinned**
- item is marked as selected to provide information/context, independently of mouse/keyboard activity. It remains in this state until something else is selected.
- expressed as a different altered background ("dark"), or a coloured border (sometimes combined with coloured icon/text)
- does not move when "hover" changes

In the light theme these visual states are currently rendered as:
- Light background:  `rgba(0,0,0,0.08)`
- Dark background:  `rgba(0,0,0,0.14)`
- Border:  theme secondary colour `#1967d2`

## Gallery view

**Hover**
- [x] remove the current border and use the hover (light) background colour
Keyboard & click behaviour are correct.

Additional theme change:
- [x] set background for gallery view to white (light theme)

<img width="358" alt="galleryhover" src="https://user-images.githubusercontent.com/6945691/55925645-0dd00e00-5bc3-11e9-9f82-4eb99c259424.png">

**Active/pinned**
Current behaviour is correct.

## Thumbnail nav

Because the layout is so pixel-fussy, we propose using the outline property to provide both the hover and focus/selected states.

On the `<div role="button">`
- [x] set a default of `outline:0`
- [x] remove the current selected border
- [x] for hover add
```
    outline: 9px solid rgba(0, 0, 0, 0.08);
    outline-offset: -2px;
```
- [x] for active add
```
    outline: 2px solid #1967d2;
    outline-offset: 3px;
```

Additional theme change:
- [x] set background for thumbnail nav to white (light theme)

## Load window 

Currently has no hover/focus indication.

**Hover**
- [x] add the hover background to the div that contains the button (i.e. not the button itself, which looks weird).

<img width="782" alt="load window hover-focus" src="https://user-images.githubusercontent.com/6945691/55925578-c8134580-5bc2-11e9-9145-be2de952c3c7.png">

[...]

PR #2564, opened 2019-04-18
Add keyboard tab navigation to the annotation companion window

jkeck:
Part of #2479


jkeck:
@jvine should tab focus also highlight the annotation as it does when mouse hovering?  (which is what I've done currently in this PR so you can check it out in the Netlify)

jvine:
Yes. Netlify looks good.

PR #2565, opened 2019-04-18
closes #2534 Update behavior of the Export workspace dialog

### LINKED ISSUE #2534: Update behavior of the Export workspace dialog
This is not critical, in that the Export workspace dialog currently works, but the interaction is a bit awkward and I think could be improved.

The primary action in the dialog is "Copy," which copies the displayed code to the clipboard. The dialog doesn't currently close after you select "Copy" so the user needs to subsequently use "Cancel" or click outside of dialog area to dismiss the dialog.

To make the interaction more efficient:

- [x] Close the dialog when the user selects "Copy"

- [ ] Display a snackbar message to confirm to the user that the code was copied to her clipboard. 

I don't think we've used the [Snackbar](https://material-ui.com/demos/snackbars/) component yet so this might be a new pattern. For this context, I'd suggest using the `top-center` position, the default styling, and the default time (I think) before it automatically disappears if the user doesn't close it manually (I'm not sure what the timing options are, but the [main example](https://material-ui.com/demos/snackbars/#simple) seems to be 5 - 7 seconds, which feels right).

Suggested text: `The workspace configuration was copied to your clipboard.`

End result, immediately after the user selects "Copy," is that the Export dialog is closed and the user sees the snackbar at the top of the window for five or so seconds (or until manually closed), similar to this:

snackbar-export-workspace

glefi:


ggeisler:
@glefi This is great, except it looks like the Cancel button no longer works?

ggeisler:
Also, @glefi since you will have to update the PR anyway, can you update the snackbar text to be:

> The workspace configuration has been copied to your clipboard.

glefi:
@ggeisler The button should now work again and I also replaced the snackbar's message.

ggeisler:
@glefi Yep, looks good now. 👍

cbeer:
Fixed by #3106

PR #2566, opened 2019-04-18
Handle initial focus/selection of the WindowList component.

jkeck:
- DisableAutoFocus on menu
- Remove selected prop from first ListItem
- Add custom onEntering function to focus the 2nd ListItem

Part of #2479 

## Before
Notice that the darker, active color is applied to the Van Gogh self portrait. However, if you look at the workspace, the focused window is actually the illustration of Krishna and Radha. 
confusing-selection

### Menu navigated by mouse
fixed-selection

### Menu navigated by keyboard
keyboard-selection

### DIFF
--- a/src/components/WindowList.js
+++ b/src/components/WindowList.js
@@ -10,6 +10,16 @@ import ns from '../config/css-ns';
  */
 export class WindowList extends Component {
   /**
+   * Given the menuElement passed in by the onEntering callback,
+   * find the 2nd ListItem element (avoiding the header) and focus it
+  */
+  static focus2ndListIitem(menuElement) {
+    if (!menuElement.querySelectorAll('li') || menuElement.querySelectorAll('li').length < 2) return;
+
+    menuElement.querySelectorAll('li')[1].focus(); // The 2nd LI
+  }
+
+  /**
    * Get the title for a window from its manifest title
    * @private
    */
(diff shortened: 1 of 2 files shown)


src/components/WindowList.js
```diff
@@ -9,6 +9,17 @@ import ns from '../config/css-ns';
...
 export class WindowList extends Component {
+  /**
+   * Given the menuElment passed in by the onEntering callback,
+   * find the 2nd ListItem element (avoiding the header) and focus it
+  */
+  static focus2ndListIitem(menuElement) {
+    if (menuElement.firstChild.tagName !== 'UL' || menuElement.firstChild.children.length < 2) return;
+    if (menuElement.firstChild.children[1].tagName !== 'LI') return;
+
+    menuElement.firstChild.children[1].focus(); // The 2nd LI
```
jkeck: Really not happy about this, but not particularly sure what to do.  The `menuElement` that we're getting is a native HTML element and it is what is passed from MUI in the `onEntering` callback prop. https://github.com/mui-org/material-ui/blob/92d6e987a43041fde63fd02cad61277d04193c29/packages/material-ui/src/Menu/Menu.js#L60-L62

Another approach that I could imagine would be to conditionally wrap the first window `ListItem` w/ a `RootRef` and use that ref instead (but this felt like how MUI is pushing us toward interacting w/ the element).  Happy to explore refs further if this seems like a "Terrible No Good" idea.

jkeck: Using querySelectorAll instead of native element traversal.  Thanks @camillevilla!

PR #2571, opened 2019-04-18
Update palette-to-component mappings

### LINKED ISSUE #2563: Update theme settings and assignments related to secondary color
Below are some recommendations for updating the theme settings. These are not urgent from the standpoint of improving the way the app looks by default, because they won't cause any visible change to our current UI. However, if our thinking is correct they will provide a more logical set of theme settings and make it a bit easier for implementors to customize things, so the sooner we can make these updates the sooner implementors can take advantage of them.

We currently use the secondary color just for the annotation notification indicator. Even if we eventually use that for other notifications (e.g., layers), this feels like an unconventional use of `secondary` in that it is so limited and specialized.

Also, even though we aren't doing it in the default implementation, it seems reasonable that other implementors might want to use different colors for the FAB and the selection/form controls. If I understand things correctly, currently an implementor who wants to do that would have to override the color assignments in multiple components (either the FAB or the various selection/form controls) to achieve that.

Here are my suggestions to fix this situation, although I think both @cbeer and @mejackreed  have a better technical understanding of the theming stuff so feel free to object if any of this doesn't make sense to you.

- [x] Update the color property used by the selection/form controls to `secondary.main` from the current `primary.main` color.

- [x] Change the `secondary.main` color in `settings.js` to `#1967d2`. (This is the same as the `primary.main` color so might seem redundant but the point is that this now enables an implementor to make the FAB and selection/form controls different colors just by updating the `primary.main` and `secondary.main` assignments in `settings.js`.)

- [x] Add a `secondary.main` color for the dark theme and set it to `#4db6ac`.

- [x] Make sure the current window top bar border still uses `primary.main`.

The end result of the above changes shouldn't visibly affect our current UI, other than the annotation indicator dot, since it is currently assigned `secondary.main` color. To fix that:

- [x] Add `notification` as a new theme setting and assign it the orange color:

```
notification: {
  main: '#ffa224',
},
```

- [x] Update the color property assigned to the annotation notification indicator to `notification.main` from the current `secondary.main`

Those changes should result in the UI looking no different than it currently does, but also should make it more clear and easier for implementors to customize the FAB, form controls, and notification indicator colors.

camillevilla:
Closes #2563

This PR:
- Sets the same hex value for `primary.main` and `secondary.main` in Mirador's default settings. Sets the Selects and FormControls to `secondary.main`. This will give implementers a little more flexibility in customizing their Mirador instances (see description in #2563). 
- Adds a `notification.main` color to the palette and wires it up to the annotations icon

To test out the new mapping,  change the secondary color in your browser console:

```javascript
var action = miradorInstance.actions.updateConfig({
  theme: {
    palette: {
      secondary: {
        main: '#ba55d3',
      },
    },
  },
});
miradorInstance.store.dispatch(action);
```
### Example with custom `secondary.main` value
Screenshot illustrating palette mappings

### DIFF
--- a/__tests__/src/components/WindowThumbnailSettings.test.js
+++ b/__tests__/src/components/WindowThumbnailSettings.test.js
@@ -29,13 +29,13 @@ describe('WindowThumbnailSettings', () => {
     expect(labels.at(2).props().value).toBe('far-right');
   });
 
-  it('should set the correct label active (by setting the primary color)', () => {
+  it('should set the correct label active (by setting the secondary color)', () => {
     let wrapper = createWrapper({ thumbnailNavigationPosition: 'far-bottom' });
-    expect(wrapper.find(FormControlLabel).at(1).props().control.props.color).toEqual('primary');
-    expect(wrapper.find(FormControlLabel).at(2).props().control.props.color).not.toEqual('primary');
+    expect(wrapper.find(FormControlLabel).at(1).props().control.props.color).toEqual('secondary');
+    expect(wrapper.find(FormControlLabel).at(2).props().control.props.color).not.toEqual('secondary');
 
     wrapper = createWrapper({ thumbnailNavigationPosition: 'far-right' });
-    expect(wrapper.find(FormControlLabel).at(2).props().control.props.color).toEqual('primary');
+    expect(wrapper.find(FormControlLabel).at(2).props().control.props.color).toEqual('secondary');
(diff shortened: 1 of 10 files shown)


camillevilla:
Great idea, @ggeisler! I've popped another commit with that documentation onto this PR. I'll update the wiki pages once this gets approved by a dev.

PR #2575, opened 2019-04-19
sets preventDefaultAction on canvas click

### LINKED ISSUE #2573: Prevent default zoom action on canvas click
OSD has a default zoom action when a user clicks on the canvas.  This is disruptive for users who may have a specific viewport previously targeted for comparison with another window.

christopher-johnson:
closes #2573


christopher-johnson:
will resubmit on alpha.6 when #2593 is fixed.

PR #2583, opened 2019-04-19
Make MuiListSubheaders that have role=presentation have no focus …

jkeck:
…outline.

Part of #2479 

## Before
beforeFocus

## After
afterFocus

### DIFF
--- a/src/config/settings.js
+++ b/src/config/settings.js
@@ -146,6 +146,13 @@ export default {
       useNextVariants: true // set so that console deprecation warning is removed
     },
     overrides: {
+      MuiListSubheader: {
+        root: {
+          '&[role="presentation"]:focus': {
+            outline: 0,
+          },
+        },
+      },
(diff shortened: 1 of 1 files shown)


src/config/settings.js
```diff
@@ -146,6 +146,13 @@ export default {
       useNextVariants: true // set so that console deprecation warning is removed
     },
     overrides: {
+      MuiListSubheader: {
+        root: {
+          '&[role="presentation"]:focus': {
```
cbeer: This still allows the components to have focus though, right? Could we do something dumb like add `tabindex="-1"` to keep it out of tabbing entirely?

jkeck: This is keyboard arrow navigable, not tab navigable, so tabindex does us no good here.  MUI is handling this [keyboard/focus implementation](https://github.com/mui-org/material-ui/blob/f132c3c8a44e119c893877f9b1804d820a02b59f/packages/material-ui/src/MenuList/MenuList.js#L99-L138).

PR #2584, opened 2019-04-19
Prevent OSD from performing click-to-zoom when the window just received focus.

cbeer:
(Note: this is targeting `2573-preventDefaultAction)

I don't know if this fixes the most annoying part of #2573 (where clicking a window to get focus also ends up zooming the canvas) without disabling click-to-zoom for all other uses.

If it's acceptable, I can fix up the tests as needed.

### DIFF
--- a/src/components/OpenSeadragonViewer.js
+++ b/src/components/OpenSeadragonViewer.js
@@ -35,14 +35,6 @@ export class OpenSeadragonViewer extends Component {
   }
 
   /**
-   *
-   * @param e
-   */
-  static onCanvasClick(e) {
-    e.preventDefaultAction = true;
-  }
-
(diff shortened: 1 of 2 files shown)


christopher-johnson (commented):
this "enable click-to-zoom after focus timeout" idea seems to serve a different user preference than #2575 (i.e. "disable click-to-zoom always".  both solutions are valid, though objects are perhaps disjoint.  either option could be provided in settings and opt-in would be configurable.

marlo-longley:
This targeted a feature branch and was ultimately fixed by https://github.com/ProjectMirador/mirador/pull/3987 which switched OpenSeadragon to use double-click for zooming instead of single-click

Issue #2395, opened 2019-04-05
Design for plugin areas 


charbugs:
We said there should be special places where plugins can add an icon. Currently there are two such places (WindowTopMenu and WorkspaceControlPanel) but the design is arbitrary at this point. What's the desired design for this places?

jvine:
There is design for the WindowTopMenu hiding over in another plugin issue: #1985.  The intent is to add items to the menu as icon+text or text-only menu options that can launch a dialog, open a toolbar, toggle a setting.

<img width="741" alt="Screen Shot 2019-04-08 at 2 38 12 PM" src="https://user-images.githubusercontent.com/6945691/55758995-2b5b7700-5a0d-11e9-9316-87c42fd2ac32.png">

For the WorkspaceControlPanel menu, #2376 might provide more context about where plugins can be added - basically, to either the Settings or Workspace options menu, depending on the character of the plugin. Again, the intent is that items will be added as icon+text or text-only options, not as icons alone.

charbugs:
Oh, I missed that. Thank you for the link.

mejackreed:
@charbugs Do you envision every component that needs to display plugins to separately implement `PluginHook`? I've started to look into #1985 and it seems like I will need that hook in `OpenSeadragonViewer`.

Should the HOC know how to do that in each place?

charbugs:
> Do you envision every component that needs to display plugins to separately implement PluginHook?

Yes (when it comes to add-plugins with UI). See https://github.com/ProjectMirador/mirador/pull/2299#issuecomment-475768564

> Should the HOC know how to do that in each place?

Not sure if I understand right, but the HOC (`PluginHoc` in `withStyle.js`) is generic. It simply passes the add-plugins to the target and then it's up to the target the render or to ignore the plugin.

PR #2590, opened 2019-04-23
Create plugin list item component. Part of #2395.

charbugs:
Create a component that can be used by adopters to match mirador style guides while adding a plugin to a menu.

pluginlistitem


charbugs:
Travis build failed. The error messages does not make sense for me. Any ideas?

christopher-johnson:
see #2593 for why build fails

PR #2597, opened 2019-04-30
[WIP] 1890 app crashes when manifest from config cant be loaded

glefi:


cbeer:
Fixed with sagas, and errors bubble up to the window with #3092.

PR #2615, opened 2019-05-10
Add Japanese locale

knagasaki:
Please add a Japanese translation


src/config/settings.js
```diff
@@ -195,6 +195,7 @@ export default {
     de: 'Deutsch',
     en: 'English',
     fr: 'Français',
+    ja: 'Japanese',
```
mejackreed: @knagasaki should this be something like `日本人` instead? For the other languages we use that language's representation in the native language.

Either way would be ok for me.

knagasaki: @mejackreed Thank you for your advice. I changed it into the word in Japanese.

PR #2618, opened 2019-05-16
Check whether pluginMap is empty

broadwell:
Using Mirador v3.0.0-alpha.7 as an embedded component in a React app, based on the instructions in [this](https://github.com/ProjectMirador/mirador/wiki/M3-Embedding-in-Another-Environment) doc from the wiki, the following error is generated in the browser, breaking the app:

`TypeError: pluginMap is undefined
    PluginHoc webpack:///./node_modules/mirador/dist/es/src/extend/withPlugins.js?:29`

Note that the example from the documentation does not deal with plugins. Adding an `isEmpty` check for the `pluginMap` in `withPlugins.js` makes this error go away, and the viewer loads properly. The effect on plugin functionality is undetermined.

### DIFF
--- a/src/extend/withPlugins.js
+++ b/src/extend/withPlugins.js
@@ -9,11 +9,6 @@ function _withPlugins(targetName, TargetComponent) { // eslint-disable-line no-u
   /** */
   function PluginHoc(props) {
     const pluginMap = useContext(PluginContext);
-
-    if (isEmpty(pluginMap)) {
-      return <TargetComponent {...props} />;
-    }
-
     const plugins = pluginMap[targetName];
 
     if (isEmpty(plugins)) {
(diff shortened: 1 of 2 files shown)


src/extend/withPlugins.js
```diff
@@ -9,6 +9,11 @@ function _withPlugins(targetName, TargetComponent) { // eslint-disable-line no-u
...
   function PluginHoc(props) {
     const pluginMap = useContext(PluginContext);
+
+    if (isEmpty(pluginMap)) {
+      return <TargetComponent {...props} />;
+    }
+
     const plugins = pluginMap[targetName];
 
     if (isEmpty(plugins)) {
```
mejackreed: I wonder if we can delete this now?

broadwell: I think we definitely should keep that check -- lots of test cases break if it's removed after adding the `pluginMap` check above it. Presumably this is because the downstream checks of the properties of `plugins` don't cover all the bases.

I've submitted the PR as a single commit here:
https://github.com/ProjectMirador/mirador/pull/2620

broadwell:
Closed in favor of https://github.com/ProjectMirador/mirador/pull/2620

PR #2620, opened 2019-05-16
Add isEmpty() check for pluginMap

broadwell:
Using Mirador v3.0.0-alpha.7 as an embedded component in a React app, based on the instructions in [this](https://github.com/ProjectMirador/mirador/wiki/M3-Embedding-in-Another-Environment) doc from the wiki, the following error is generated in the browser, breaking the app:

`TypeError: pluginMap is undefined PluginHoc webpack:///./node_modules/mirador/dist/es/src/extend/withPlugins.js?:29`

Note that the example from the documentation does not deal with plugins. Adding an `isEmpty()` check for the `pluginMap` in `withPlugins.js` makes this error go away, and the viewer loads properly in this case.

See also the discussion around closed PR https://github.com/ProjectMirador/mirador/pull/2618

### DIFF
--- a/src/extend/withPlugins.js
+++ b/src/extend/withPlugins.js
@@ -9,6 +9,11 @@ function _withPlugins(targetName, TargetComponent) { // eslint-disable-line no-u
   /** */
   function PluginHoc(props) {
     const pluginMap = useContext(PluginContext);
+
+    if (isEmpty(pluginMap)) {
+      return <TargetComponent {...props} />;
+    }
+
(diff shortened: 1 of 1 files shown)


charbugs:
Sorry for the late answer. The problem is that the instructions in the wiki page are not up to date. The App component needs to be wrapped in a PluginProvider component to work correctly. See the initialisation code: https://github.com/ProjectMirador/mirador/blob/master/src/lib/MiradorViewer.js#L29. 

Please, see also #2627

PR #2630, opened 2019-05-29
Adds a configuration setting for hiding the window top bar title

### LINKED ISSUE #2623: Add configuration setting to hide window title
A common scenario is to embed the viewer in a larger application, such as a library catalog show page. In those contexts, the title of the object in the viewer is usually prominently displayed elsewhere on the page so showing it again the M3 viewer title bar is redundant. As an example (this is M2 but same scenario would currently occur with M3), here's a show page from a Spotlight exhibit:

Screen Shot 2019-05-17 at 12 39 45 PM

Adding a `window` setting to control the display of the window title would help the implementor control embed scenarios better. Something like this:

```
 window: {
   hideWindowTitle: false, // Configure if the window title is shown in the window title bar or not
}
```

mejackreed:
Fixes #2623 

With `hideWindowTitle: true`

Screen Shot 2019-05-30 at 7 13 41 AM

A few questions come up:

 - Should we even render the component if `hideWindowTitle: true`? ~I don't think so now because of the spacing~ ✅ 
 - Should the buttons be pulled right? ~YES~  ✅

### DIFF
--- a/__tests__/src/components/WindowTopBar.test.js
+++ b/__tests__/src/components/WindowTopBar.test.js
@@ -1,12 +1,12 @@
 import React from 'react';
 import { shallow } from 'enzyme';
 
-import Typography from '@material-ui/core/Typography';
 import Toolbar from '@material-ui/core/Toolbar';
 import AppBar from '@material-ui/core/AppBar';
 
 import WindowTopMenuButton from '../../../src/containers/WindowTopMenuButton';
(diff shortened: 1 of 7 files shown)


ggeisler:
@mejackreed Same answer as in #2623 but in this context I understand your question more clearly: yes, definitely pull the buttons to right-align in the window (other than the sidebar panel button, which should remain left-aligned).

mejackreed:
Great.. I've updated it to do what we need.

PR #2640, opened 2019-05-31
Update authentication-related UI elements

### LINKED ISSUE #2636: Update authentication-related UI elements
The intention (based on discussion with @cbeer and @jvine) is to switch the authentication UI elements [currently a snackbar and a modal dialog](https://github.com/ProjectMirador/mirador/pull/2367) to a single, expandable div that is displayed at the top of the object window, just below the window top bar (example on left below).

This `div` does not overlay the window canvas but is inserted between the window top bar and the canvas/side panel/companion window content. This means when the user clicks "Continue", the `div` expands and pushes the canvas/sidebar panel/companion windows down (example on right below).

expandable-div

Here are same examples but in a scenario where the user has the window side panel open:

expandable-div-sidepanel

The same `div` should be used to show the failure message when necessary:

general-failure

### Initial collapsed state
Details of the initial, collapsed `div` (or a variation of a companion window; whatever makes sense technically):

collapsed-details

- [x] `div` is full width of the window, inserted directly after the window top bar. Background color is the current palette's secondary color. (Presumably the `onText` color is calculated automatically based on the background color? If not, we might have to discuss expanding the theme configuration to give the auth div explicit background and text colors.)
- [x] Lock icon: `lock` icon (@material-ui/icons/LockSharp). Ensure it has something like `margin-right: 12px` to separate it from the label text.
- [x] Value of the 'label' property is the text next to the lock icon.
- [x] Make 'label' an `h3` but style it using the `body1` [typography](https://github.com/ProjectMirador/mirador/issues/2093).
- [x] The text "CONTINUE", in `body1` typography, with something like `margin-left: 20px` to offset it from the `label` text.
- [x] Although the "CONTINUE" text looks like a text button, we'd rather make the whole _lock icon/label text/CONTINUE_ block a single button, which when selected would toggle the `div` open (showing the full auth message and action buttons) and closed (the initial state).

### Expanded state
When the user toggles the auth `div` open, it expands to show the "recommended" elements `header` and `description` and the action buttons:

expanded-details

- [x] If provided, display the 'header' value, styled using `body1` typography.
- [x] If provided, display the 'description' value.
- [x] If both header' and 'description' are provided, insert a colon after `value` so it reads as in the mockup above.
- [x] Add a text button labelled "Cancel", as in the mockup.
- [x] When selected, the "Cancel" button should collapse the `div` back to its initial state.
- [x] Add a primary action button next to the "Cancel" button, using the filled button style. If provided, use the `confirmLabel` value as the button label. If `confirmLabel` is not provided, use the fallback text in  #2637 and #2638 for the button label.
- [x] When selected, the primary action button should trigger the opening of the authentication service, as it does in the current implementation.
- [x] The "CONTINUE" text should be hidden when the `div` is in the expanded state (forgot to remove it from the mockup).

### Failure state
If the user succeeds in authenticating (or agrees to the terms of service), the auth `div` should be removed. If the user's authentication/terms of service action results in failure, however, we need to show a failure message(s):

failure-details

- [x] Maintain the lock icon and the action buttons from the expanded state of the `div`, but replace the text as described next:
- [x] If provided, show the 'failureHeader' value next to the lock icon. If not provided, use the fallback text in #2637, #2638, and #2639 instead.
- [x] If provided, show the 'failureDescription' value below. If not provided, omit the description.
- [x] If the user selects the "Cancel" button in the failure state, the `div` should return to the initial state (collapsed, showing the label as described in the "Initial collapsed state" section above).
- [x] The "Login" button should trigger opening the authentication service, same as when in the expanded state.

### Example manifest
Mockups above based on:
- Manifest: https://iiifauth.digtest.co.uk/manifest/11_kitty_joyner
- Auth section: https://iiifauth.digtest.co.uk/img/11_kitty_joyner_degraded.jpg/info.json

### Fallback text
Since only one of the six auth text properties is **required**, we should in some cases provide fallback text when a given property doesn't have a value. The fallback text depends on context (which authentication pattern is in play), so I'm creating separate tickets with the details for each auth pattern:

#2637 - Customize auth details for Login pattern
#2638 - Customize auth details for Clickthrough pattern
#2639 - Customize auth details for Kiosk and External patterns

### Theme examples
If we base the background color of the auth `div` on the secondary color of the theme as suggested earlier, the mockups above are what we'd get from the current Mirador 3 default light theme. Here's how it would look when switching to the default dark theme:

dark-theme

And a custom theme:

sul-theme

### LINKED ISSUE #2637: Customize auth details for Login pattern
Related to #2636.

If the manifest auth `profile` value indicates the object should use the [`login` interaction pattern](https://iiif.io/api/auth/1.0/#login-interaction-pattern), the primary action button should open the authentication service. Use the following fallback strings when not values are not provided in the manifest:

- [x] `label`: "Authentication required for full access" (I'm not sure we need fallback text for this one, since `label` is "required" per the [auth spec](https://iiif.io/api/auth/1.0/#service-description), but perhaps we should add it just in case?).
- [ ] `header`: "Log in to access all available features."
- [x] `description` : omit if not provided.
- [x] `confirmLabel` : "Log in" (note space between words, which differs from the auth spec examples)
- [x] `failureHeader`: "Authentication failed"
- [x] `failureDescription`: omit if not provided

Here's a mockup of this scenario, where the manifest included the required 'label' value ("Login to Example Institution") but no other auth text property values:

login-example

login-failed

### LINKED ISSUE #2639: Customize auth details for Kiosk and External patterns
Related to #2636.

If the manifest auth `profile` value indicates the object should use either the [`kiosk`](https://iiif.io/api/auth/1.0/#kiosk-interaction-pattern) or [`external`](https://iiif.io/api/auth/1.0/#external-interaction-pattern) interaction patterns, there will be no auth-related user interaction, but a failure message might need to be displayed.

The differences from the `login` pattern are the same for these two interaction patterns, I believe:

- [x] If after processing the manifest the auth service successfully obtains the access cookie/token, the object is displayed normally and no auth `div` should be displayed.
- [x] If the auth service fails to obtain the access cookie or token, display the auth `div` in the failure state.
- [x] Keep the lock icon but omit the action buttons.
- [x] If provided, display the `failureHeader` and `failureDescription` values as in #2636.
- [x] If a `failureHeader` is not provided, use the fallback string `Authentication failed`.
- [x] If a `failureDescription` is not provided, omit it.

Here's an example of a failure message for the `kiosk` interaction pattern, where the `failureHeader` and `failureDescription` values are provided (using the [spec example](https://iiif.io/api/auth/1.0/#kiosk-interaction-pattern)):

kiosk

And here's an example of a failure message for the `external` interaction pattern, where the `failureHeader` and `failureDescription` values are provided (using the [spec example](https://iiif.io/api/auth/1.0/#external-interaction-pattern)):

external

cbeer:
Fixes #2636 
Fixes #2637 
Fixes #2639


ggeisler:
@cbeer I can't seem to load any new manifests into the Netlify instance for this PR. Can you suggest a manifest that you know works so I can see your revision in action?

From your description, I think this part makes sense:

> I think we shouldn't display a fallback header if a description is provided (and vice versa), to avoid weird cases where our fallback wording conflicts with what we get from the provider.

But I'm unsure of this part:

> this PR omits the collapsed section (and controls) entirely, and always displays the login button instead when that content is absent

because if the user knows they can't auth or just doesn't want to, the collapsed version is more compact and enables the user to work with the un-authed object with less visual clutter than the expanded version. Perhaps this isn't relevant in your revision, but I'm not sure without seeing it in action. I guess a screenshot would be enough for me to review if for some reason it isn't feasible to see the interaction via Netlify.

PR #2649, opened 2019-06-06
Sanitize IIIF metadata links to use target=_blank; fixes #2594

### LINKED ISSUE #1944: Override link behaviour of html anchors included in manifests
Sometimes links provided in the html content of manifest metadata open in the current tab, and sometimes they open in another tab. We should override this behaviour so all supplemental information opens in a new browser tab.

### LINKED ISSUE #2594: Links in Rights panel should open in new window
The License & other links in the Rights panel currently open in the same window and lose all your lovely Mirador work. They should open in a new window, like the links in the Metadata panel.

### LINKED ISSUE #2624: Update link color in dark theme
When in dark theme, the color of links remains the same blue color it is in the light theme:

Screen Shot 2019-05-17 at 4 44 11 PM

We need to make sure that when in dark theme, the `color` property of links is something with better contrast. I suggest using `rgba(255,255,255,0.7)` which is far enough from the white used in the headings and body text in the panel to stand out a bit, while having sufficient contrast for accessibility (it's also the same color used for the icons in the window sidepanel menu):

Screen Shot 2019-05-17 at 4 45 01 PM

cbeer:
Fixes #2594 
Fixes #2624 
Fixes #1944

### DIFF
--- a/__tests__/src/components/CanvasInfo.test.js
+++ b/__tests__/src/components/CanvasInfo.test.js
@@ -4,7 +4,7 @@ import Typography from '@material-ui/core/Typography';
 import { CanvasInfo } from '../../../src/components/CanvasInfo';
 import { LabelValueMetadata } from '../../../src/components/LabelValueMetadata';
 import CollapsibleSection from '../../../src/containers/CollapsibleSection';
-import { SanitizedHtml } from '../../../src/components/SanitizedHtml';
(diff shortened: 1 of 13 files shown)


mejackreed:
Also #1944?

cbeer:
I guess so -- didn't know about that one.

PR #2651, opened 2019-06-07
Sets up consistent sizing for import/export dialogs + enhancements

### LINKED ISSUE #2567: Update styling of the Import workspace dialog
To better match the styling of the Export workspace dialog and make the Import dialog a bit more user-friendly:

- [x] **Put the cursor in the `textarea` when the dialog is displayed.** There are no other things for the user to do here, so I don't see a downside and it would eliminate one click to accomplish the task.

- [x] **Update the styling of the `textarea` font.** Since the user is basically working with the same text for similar purposes as in the Export workspace dialog, I suggest using the same typography used there (smaller font size, monospace font, and no line-wrapping).

- [x] **Widen the dialog.** Again, the Export workspace dialog seems like a good model to copy, since it feels like the right width and seems to have appropriate responsiveness.

- [x] **Update how the hint text ("Paste a Mirador 3 configuration to be imported") is implemented.** It currently crowds the bottom of the `textarea` pretty tightly. Rather than give  it some top margin directly, I'd suggest looking at how the [hint text in the Add Resource form](https://github.com/ProjectMirador/mirador/blob/d6b306aff4300b2f5b14a9c6c0b197372e141f01/src/components/ManifestForm.js#L111) in the load window was done. That uses a built-in `helperText` property that provides the right margin and text color and is probably a more appropriate way to do the hint text than the current implementation.

mejackreed:
fixes #2567

Screen Shot 2019-06-07 at 7 03 55 AM
Screen Shot 2019-06-07 at 7 03 44 AM

### DIFF
--- a/__tests__/src/components/WorkspaceExport.test.js
+++ b/__tests__/src/components/WorkspaceExport.test.js
@@ -31,6 +31,13 @@ describe('WorkspaceExport', () => {
     expect(wrapper.find('WithStyles(Dialog)').length).toBe(1);
   });
 
+  it('renders sizing props', () => {
+    expect(wrapper.find('WithStyles(Dialog)').props()).toEqual(expect.objectContaining({
+      fullWidth: true,
+      maxWidth: 'sm',
+    }));
+  });
+
   it('renders an exportable version of state', () => {
     expect(wrapper.find('pre').length).toBe(1);
     expect(wrapper.find('pre').text()).toMatch('"companionWindows":');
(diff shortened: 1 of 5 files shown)


ggeisler:
@mejackreed Fixing the styling of the Import hint text is checked off in ticket #2567 but from the screenshot it doesn't look like it was improved as intended in my ticket. Compare to the Add resource hint text, which has both left and top margin. The Import text hint appears to have no top or left margin?

I tried figuring out what was going on via Web Inspector in the Netlify instance on this PR, but I'm getting some weird behavior when I try to inspect the hint text (it disappears), which is probably just something going on with my browser, so I'm not 100% sure how the Import text hint is actually styled in this PR.

mejackreed:
Good catch on that @ggeisler , that was previously checked so I didn't look at that. I've added on a follow on commit that changes that.

PR #2680, opened 2019-06-11
Add boilerplate search state management

mejackreed:


src/state/actions/search.js
```diff
@@ -0,0 +1,67 @@
+import fetch from 'node-fetch';
+import ActionTypes from './action-types';
+
+/**
+ * requestSearch - action creator
+ *
+ * @param  {String} targetId
```
cbeer: what is targetId here?

mejackreed: `targetId` is likely just the `manifestId` most of the time. If we extend it further it could be other things I think

src/state/reducers/search.js
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * searchReducer
+ */
+export const searchesReducer = (state = {}, action) => {
+  switch (action.type) {
+    case ActionTypes.REQUEST_SEARCH:
+      return {
+        ...state,
+        [action.targetId]: {
+          [action.searchId]: {
```
cbeer: We probably want to include `...state[action.targetId]` in all these reducers, right (for either when I have multiple search companion windows open, or two windows with the same manifest)?

mejackreed: I'm not sure I follow here. Maybe we can chat quickly?

mejackreed: 👍 added

PR #2681, opened 2019-06-11
Hook up search state to form

mejackreed:
Not necessarily ticketed but a prereq to #2670 

hookupsearch

### DIFF
--- a/src/components/SearchPanelControls.js
+++ b/src/components/SearchPanelControls.js
@@ -10,31 +10,66 @@ import InputAdornment from '@material-ui/core/InputAdornment';
 /** */
 export class SearchPanelControls extends Component {
   /** */
+  constructor(props) {
+    super(props);
+
+    this.state = { search: '' };
+    this.handleChange = this.handleChange.bind(this);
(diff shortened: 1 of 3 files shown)


src/components/SearchPanelControls.js
```diff
@@ -9,32 +9,64 @@ import InputAdornment from '@material-ui/core/InputAdornment';
...
-      </div>
+      </form>
     );
   }
 }
 
 SearchPanelControls.propTypes = {
   cwId: PropTypes.string.isRequired,
+  fetchSearch: PropTypes.func.isRequired,
+  searchService: PropTypes.object.isRequired, // eslint-disable-line react/forbid-prop-types
```
jkeck: It looks like we're starting to get an idea of the interface of the `searchService` we're depending on (`id` & `options`).  Just thinking if this is a good time to assert that's what we expect here (happy to defer until we know more)

mejackreed: Good point. I've updated the proptypes definition for `id` and `options`

PR #2682, opened 2019-06-11
Add basic search results list fixes #2670

### LINKED ISSUE #2670: Hook up content search input to render simple annotation list

mejackreed:
https://purl.stanford.edu/fg165hz3589/iiif/manifest

searchresults

### DIFF
--- a/src/components/SearchPanel.js
+++ b/src/components/SearchPanel.js
@@ -2,6 +2,7 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import CompanionWindow from '../containers/CompanionWindow';
 import SearchPanelControls from '../containers/SearchPanelControls';
+import SearchResults from '../containers/SearchResults';
 /** */
 export class SearchPanel extends Component {
   /** */
@@ -18,7 +19,12 @@ export class SearchPanel extends Component {
         windowId={windowId}
         id={id}
         titleControls={<SearchPanelControls cwId={id} windowId={windowId} />}
-      />
+      >
+        <SearchResults
+          cwId={id}
+          windowId={windowId}
+        />
+      </CompanionWindow>
     );
   }
 }
--- a/src/components/SearchResults.js
+++ b/src/components/SearchResults.js
@@ -1,0 +1,49 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import List from '@material-ui/core/List';
+import ListItem from '@material-ui/core/ListItem';
+import ListItemText from '@material-ui/core/ListItemText';
+import SanitizedHtml from '../containers/SanitizedHtml';
+
+/** */
+export class SearchResults extends Component {
+  /** */
+  render() {
+    const {
+      cwId,
+      searchHits,
+    } = this.props;
+
+    return (
+      <>
+        <List>
+          {
+            searchHits.map(hit => (
+              <ListItem
+                button
+                key={`${cwId}-${hit.annotations[0]}`}
+                component="li"
+              >
+                <ListItemText primaryTypographyProps={{ variant: 'body2' }}>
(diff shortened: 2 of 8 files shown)


src/components/SearchPanel.js
```diff
@@ -18,7 +19,11 @@ export class SearchPanel extends Component {
         windowId={windowId}
         id={id}
         titleControls={<SearchPanelControls cwId={id} windowId={windowId} />}
-      />
+      >
+        <SearchResults
```
cbeer: Don't we need to use the companion window id here too, in case I pop two search windows out?

mejackreed: Nice catch. I've pushed that up.

src/containers/SearchResults.js
```diff
@@ -0,0 +1,27 @@
...
+  getSearchHitsForManifest,
+} from '../state/selectors';
+
+/**
+ * mapStateToProps - used to hook up connect to state
+ * @memberof SearchResult
+ * @private
+ */
+const mapStateToProps = (state, { windowId }) => ({
+  searchHits: getSearchHitsForManifest(state, { windowId }),
```
cbeer: Sorry, I may have missed the boat on the data shape, but `getSearchHitsForManifest` probably needs to get keyed on the companion window id too, and that'll need to trickle down all the way to the redux state?

mejackreed: Possibly? Is that for garbage collection? I was kind of just kicking the can down until the `clear` button got implemented.

mejackreed: Happy to chat about this.

src/components/SearchResults.js
```diff
@@ -0,0 +1,49 @@
...
+export class SearchResults extends Component {
+  /** */
+  render() {
+    const {
+      cwId,
+      searchHits,
+    } = this.props;
+
+    return (
+      <>
```
jkeck: I believe this Fragment is redundant since we have a top-level `<List>`

src/containers/SearchResults.js
jkeck: Yeah, I guess that's a good question.  Not sure if you have chatted about this already, but this may be a question for @jvine / @ggeisler (e.g. when I pop out the search sidebar to a companion window on the right, then open up the search again in the sidebar, should my previous search be initialized, or should I be getting a blank search box to initialize a new search)

src/components/SearchResults.js
mejackreed: Good catch, I was thinking that there would be additional components coming in for future tickets and hoping that this might be a placeholder. Do you think it's worth keeping in for that?

src/containers/SearchResults.js
mejackreed: Spinning this out to a new issue https://github.com/ProjectMirador/mirador/issues/2685 . I'm 👍 to this, but think it could happen in concert w/ other issues

mejackreed: And I think it would mean that the CS behavior would be different than the annotations panel behavior, no?

jkeck: Yeah, it has pretty large implications for differing behavior.  I think we can ship this as is assuming that is the behavior we want (and seeing that #2687 is ready).  I _assume_ this is ultimately what we want.

src/components/SearchResults.js
jkeck: Yeah, that's fine.  No harm, no foul. 😸

src/containers/SearchResults.js
mejackreed: :shipit:

ggeisler:
@mejackreed @jkeck @cbeer: JV and I discussed this and think we should open the new search panel (when the user has made the initial search panel a companion window) in the blank, new search state.

We foresee many related complications to work out (especially when the search service returns results in pages) but we can work them out as we go.

PR #2687, opened 2019-06-12
Key searches on the companion window id so multiple search windows can be shown at a time

### LINKED ISSUE #2685: Enhance search state to take a companionwindowId for uniqueness across companion windows
Spun out of #2682

cbeer:
Fixes #2685 

Includes #2682


mejackreed:
I'm 👌 with this approach.

PR #2689, opened 2019-06-12
Display search result annotations on single and book view

### LINKED ISSUE #2673: Render content search result on single page view

### LINKED ISSUE #2674: Render content search results in book view

mejackreed:
Fixes #2674
Fixes #2673

One major caveat here is that this only works for the search CW in the "left" position. I wrestled around with this and ended up at the point where we need more information. I think implementing the selected/highlighted will maybe give us this information, but I didn't want to refactor the world at the moment. Open to a different approach though.

show the annos


mejackreed:
Thanks @jkeck in discussion w/ @cbeer working on a selector/state change that keys off window instead of manifest. I think this will give us more of what we want.

PR #2690, opened 2019-06-12
Spike of contentsearch autocomplete using downshift

cbeer:


mejackreed:
I'm not able to see results pop up for me, but I am seeing the autocomplete requests fire off and return.

PR #2691, opened 2019-06-12
Create a configuration settings for where to show the authentication …

cbeer:
…popup

### DIFF
--- a/src/config/settings.js
+++ b/src/config/settings.js
@@ -206,6 +206,7 @@ export default {
     allowClose: true, // Configure if windows can be closed or not
     allowFullscreen: false, // Configure to show a "fullscreen" button in the WindowTopBar
     allowMaximize: true, // Configure if windows can be maximized or not
+    authNewWindowCenter: 'parent', // Configure how to center a new window created by the authentication flow. Options: parent, screen
     defaultSideBarPanel: 'info', // Configure which sidebar is selected by default. Options: info, attribution, canvas, annotations
     defaultView: 'single',  // Configure which viewing mode (e.g. single, book, gallery) for windows to be opened in
     hideAnnotationsPanel: false, // Configure to hide the annotations panel in the WindowSideBarButtons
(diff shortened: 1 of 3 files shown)


src/config/settings.js
```diff
@@ -206,6 +206,7 @@ export default {
     allowClose: true, // Configure if windows can be closed or not
     allowFullscreen: false, // Configure to show a "fullscreen" button in the WindowTopBar
     allowMaximize: true, // Configure if windows can be maximized or not
+    authNewWindowCenter: 'parent', // Configure how to center a new window created by the authentication flow. Options: parent, screen
```
jkeck: Would this work as a top-level config option or would this need to potentially change on a window-per-window basis?

cbeer: I think it's just a top-level option.

jkeck: Okay, so should it not be in the window config and instead go up one level?  It seems like all the other config options w/i window are specific to windows (e.g. windows defaultView is "single") where the `authNewWindowCenter` config isn't intended to be configurable on a per-window basis and instead will affect the entire app.

It's not super important, just struck me as odd that it would be w/i `window` (then again, the config is kinda all over the place so 🤷‍♂)

cbeer: Is `window` window-specific? I thought all of those `allow*`configs were global.

jkeck: They are loaded into the global config (because the entire config object is).

My point is that the configs in window seem to be meant to apply to behaviors/features/UI elements specific to windows (e.g. I would set `window.sideBarOpenByDefault` to true to make all windows sidebars open by default but may override that on a specific window that I add on initialization), where this particular feature doesn't appear to be tied to windows, and is more about application specific behavior.

jkeck: Not to say we have a hard/fast rule, it was just the pattern I saw emerging here.  This doesn't need to hold up the PR.

PR #2694, opened 2019-06-13
Clean up search state when windows or companion windows are removed

### LINKED ISSUE #2672: Improve State Garbage Collection
Content search will add a lot more data to state and we'll want to ensure that we're not bloating it even further unnecessarily.

We're currently leaving some content in redux after windows are closed.

 - [x] When we close a companion window, we need to write the reducer to throw away the search
 - [x] When we close a window, we need to write the reducer to throw away all the searches

cbeer:
Fixes #2672

### DIFF
--- a/__tests__/src/reducers/search.test.js
+++ b/__tests__/src/reducers/search.test.js
@@ -68,9 +68,49 @@ describe('search reducer', () => {
     });
   });
   it('should handle IMPORT_MIRADOR_STATE setting to clean state', () => {
-    expect(searchesReducer({}, {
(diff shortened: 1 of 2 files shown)


__tests__/src/reducers/search.test.js
```diff
@@ -67,10 +67,35 @@ describe('search reducer', () => {
       },
     });
   });
-  it('should handle IMPORT_MIRADOR_STATE setting to clean state', () => {
```
mejackreed: Should we still keep this test?

Issue #2683, opened 2019-06-11
Enhance details of search results panel


ggeisler:
After the work to display basic content search results has been done via #2670 has been completed, here are some additional details.

Assuming the case where the search service returns a list of complete list of `hits` (e.g. https://contentsearch.stanford.edu) and not paged results:

<img width="703" alt="Screen Shot 2019-06-11 at 12 05 23 PM" src="https://user-images.githubusercontent.com/101482/59306140-7fbee500-8c50-11e9-804f-b733bc36ed58.png">

(For reference, I used [these results](https://contentsearch.stanford.edu/ct571zc0811/search?q=armed) for the mockups, although note that sometimes I adjusted the result text displayed for space reasons, so things don't always map precisely.)

### Note A: Hit navigation

- [x] Use the `chevronLeftSharp` and `chevronRightSharp` icons, with `padding: 12px;` for next and previous hit navigation.
- [x] Icon tooltips: "Previous result" and "Next result".
- [x] Between the next and previous icons, the text `x of y`, where x = currently selected result item and y = total number of result items/hits.
- [x] Center the hit navigation with the panel header.
- [x] Probably obvious, but the next and previous navigation is intended to enable the user to move through the results without having to select items in the panel body directly. So as the user uses the previous and next buttons, the panel body should update (changing the selected result item highlighting) and the page shown on the canvas should update (to show appropriate page and hit highlighting on canvas) accordingly.

### Note B: Hit counter (#2699)

- [x] Result items displayed in the panel body are numbered consecutively, where the number of the result item matches the hit navigation counter described above.
- [x] Use `headline 6` [typography](https://github.com/ProjectMirador/mirador/issues/2093) for the number.
- [x] For the background radius, I think `radius: 50%;` might do it. If not, try something like `16px`.
- [x] Use `#bdbdbd` or Gray400 for the background color of the counter.
- [x] When the result item is the currently selected result item, change the background color of the counter to the same yellow we use for outlining the selected annotation on the Annotations panel. (The color is intended to match the hit highlighting color on the canvas.)
- [x] When the result item is on the same page/canvas as the currently selected result item, but is not the currently select item, change the background color of the counter to the same blue we use for outlining the non-selected annotations on the current page when on the Annotations panel. (The color is intended to match the hit highlighting color on the canvas.) See the mockup in the "Example: Two hits on same page" section below.

### Note C: Canvas label (#2702)

[...]

PR #2699, opened 2019-06-13
Add a hit counter to content search results

camillevilla:
Part of #2683

- [x] Selected results change the hit counter background color to yellow
- [ ] ~Indicate results that are on the same page w/ a blue hit counter~ (will be split out to a different PR)
- [x] tests!

### DIFF
--- a/__tests__/src/components/SearchHit.test.js
+++ b/__tests__/src/components/SearchHit.test.js
@@ -1,0 +1,38 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import { SearchHit } from '../../../src/components/SearchHit';
+
+
+/**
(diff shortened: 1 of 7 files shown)


ggeisler:
@camillevilla Tying the colors to the theme seems like a smart idea but just to double-check: if I was to change the yellow to pink, say, that would change both the counter background color on the panel **and** the border color of the selected hit on the canvas to pink, correct?

The point of coloring the hit counters on the panel is to visually tie them to their associated canvas hits, so I just want to make sure that connection survives this enhancement.

cbeer:
@ggeisler not right now, but this will set us up to use the same variables consistently.

PR #2703, opened 2019-06-13
Add next/prev navigation for hits

mejackreed:
Part A of #2683

next:prev


src/components/SearchPanelControls.js
```diff
@@ -52,46 +85,80 @@ export class SearchPanelControls extends Component {
...
+        {(searchHits.length > 0) && (
+          <Typography variant="body2" align="center" classes={classes}>
+            <MiradorMenuButton
+              aria-label={t('searchPreviousResult')}
+              disabled={!this.hasPreviousResult(currentHitIndex)}
+              onClick={() => this.previousSearchResult(currentHitIndex)}
+            >
+              <ChevronLeftIcon />
+            </MiradorMenuButton>
+              {currentHitIndex + 1}
```
cbeer: This string should be i18n'ed, yes?

mejackreed: Most definitely. I've pushed this up with a rebase from autocomplete.

PR #2714, opened 2019-06-14
 Truncates plain text for a hit, to provide a more balanced before/after match

### LINKED ISSUE #2671: Add before/after content and truncation behavior
The "truncation" renders the full snippet (before/after) in the panel by itself w/ a link that will go back to the Annotation list.

(Edit by @ggeisler below with some details)
We'll have to pick some character or word limit to trigger truncation. This will probably be easy to adjust so as a starting point I'll suggest `200 characters`, which would produce something like this result:

Screen Shot 2019-06-12 at 11 43 42 AM

~When the user selects the "more ..." link, that should take them to the detail result view panel described in #2688.~

### LINKED ISSUE #2698: Should / can we balance the before/after truncation?
I’ve been thinking about the very long “before” “after” that we provide, which is not always relevant. Could we provide a truncation algorithm to make that more digestible like this:

*Pseudo code not sure if it works*
```
TOTAL_CHARS = 200

resultingChars = (TOTAL_CHARS - match.length) / 2;
measured = [before.length, after.length].filter(length => length > 0)
charsOnSide = Math.min(resultingChars, ...measured)
display = before.substring(before.length - charsOnSide, before.length);
display += match
display += after.substring(0, Math.min(after.length, charsOnSide))
```

Would turn this:

Screen Shot 2019-06-13 at 11 07 23 AM

Into this:

Screen Shot 2019-06-13 at 11 08 01 AM

mejackreed:
Fixes #2698 
Fixes #2671 

Screen Shot 2019-06-14 at 10 50 10 AM

### DIFF
--- a/__tests__/src/components/SearchHit.test.js
+++ b/__tests__/src/components/SearchHit.test.js
@@ -29,7 +29,7 @@ describe('SearchHit', () => {
     const wrapper = createWrapper({ selectContentSearchAnnotation });
     expect(wrapper.find('WithStyles(ForwardRef(ListItem))').length).toEqual(1);
     expect(wrapper.find('WithStyles(ForwardRef(ListItem))').prop('selected')).toEqual(true);
-    expect(wrapper.find('WithStyles(ForwardRef(ListItemText))').render().text()).toEqual('1Light up the moose , and start  more');
+    expect(wrapper.find('WithStyles(ForwardRef(ListItemText))').render().text()).toEqual('1Light up the moose , and start the chai more');
     expect(wrapper.find('strong').length).toEqual(1);
 
     wrapper.find('WithStyles(ForwardRef(ListItem))').simulate('click');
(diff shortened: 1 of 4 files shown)


jkeck:
In the screenshot it looks like `A meal will` is truncated very short.  What's going on there?  (or could you provide the manifest you're using for this to test out?)

mejackreed:
Sure manifest here: https://purl.stanford.edu/fg165hz3589/iiif/manifest @jkeck 

I think what is going on is that `charsOnSide` is returning a "balanced" hit of `3`. We could increase this perhaps by increasing the filter length in that function.

It seems like a design choice / tweak that we may want to play with.

cbeer:
I wonder if we want to set a minimum length on either side too, to account for times the hit is at the beginning / end of a sentence? I think that'd fix things for @jkeck too.

mejackreed:
I've pushed a commit that adds a minimum setting, hopefully accomplishing a better result for us 

https://purl.stanford.edu/fg165hz3589/iiif/manifest

Screen Shot 2019-06-14 at 4 41 40 PM

PR #2716, opened 2019-06-14
Handle more downshift state management in our component

cbeer:


cbeer:
Thanks @jkeck, I think I've fixed that.

PR #2718, opened 2019-06-14
Add the annotation resources label to the SearchHit component.

jkeck:
Part of #2683

### DIFF
--- a/src/containers/SearchHit.js
+++ b/src/containers/SearchHit.js
@@ -7,8 +7,9 @@ import { SearchHit } from '../components/SearchHit';
 import * as actions from '../state/actions';
 import {
   getCanvasLabel,
-  getSearchAnnotationForCompanionWindow,
   getSelectedCanvases,
+  getResourceAnnotationForSearchHit,
+  getResourceAnnotationLabel,
   getSelectedContentSearchAnnotationIds,
 } from '../state/selectors';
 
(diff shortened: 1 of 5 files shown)


src/containers/SearchHit.js
```diff
@@ -25,6 +25,7 @@ const mapStateToProps = (state, { hit, companionWindowId, windowId }) => {
   const hitAnnotation = resourceAnnotations.find(r => r.id === hit.annotations[0]);
 
   return {
+    annotationLabel: hitAnnotation && hitAnnotation.resource && hitAnnotation.resource.label,
```
jkeck: I don't think we can just use this `label` directly.  Just wanted to push this up, and happy to 🍐 up on how we should handle this properly (which maybe means pushing some of this to selectors)

src/state/selectors/searches.js
```diff
@@ -83,3 +85,25 @@ export const getSelectedContentSearchAnnotations = createSelector(
     ),
   })).filter(val => val.resources.length > 0),
 );
+
+export const getResourceAnnotationForSearchHit = createSelector(
+  [
+    getSearchAnnotationForCompanionWindow,
+    (state, { hitId }) => hitId,
```
cbeer: Is `hitId` the right name? Isn't this an annotation uri?

jkeck: It's the it of the annotation in the hit (from https://github.com/ProjectMirador/mirador/blob/0a375b6e4bf5bc4c91976785fd7337ab21f88a05/src/containers/SearchHit.js#L26).  My intention was to differentiate it from the id in the resource, but agree that `annotationUri` may be less ambiguous.

cbeer: `hitAnnotationUriThisIsTheOne`? :trollface:

PR #2728, opened 2019-06-16
adds sideEffects rule to webpack config

christopher-johnson:
(enables [tree shaking](https://webpack.js.org/configuration/module/#rulesideeffects)) and reduces bundle size by ~8K


christopher-johnson:
this change does not work with the current CI / deployment setup.  (several puppeteer tests time out and imported styles are missing from netlify).  closing.

PR #2729, opened 2019-06-17
optimize css loading and split from js dist

christopher-johnson:
this reduces the js bundle size by ~4KB by extracting the css into a separate file and removing comments.
loading the node_module sourced css using sass import partials is a micro-optimization that puts these styles in a global css scope instead of using css-loader.  It could be better to determine what specific styles are required for that component (e.g mosaic) and reimplement them with jss or a styled component rather than using the entire vendor supplied css.

this potentially allows implementations to override default styling using domain sourced css without having to touch the source code.

### DIFF
--- a/__tests__/integration/mirador/embedding.html
+++ b/__tests__/integration/mirador/embedding.html
@@ -26,6 +26,7 @@
     </style>
     <title>Mirador</title>
     <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:300,400,500">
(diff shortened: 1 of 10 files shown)


christopher-johnson:
no hurry.  I agree that a no css solution would clean things up.  The vendor css would need to be forked into mirador JSS I guess, but the mirador css could definitely be eliminated.   This may pose some issues for ppl who want to use customize the dist package with their own styles.  Not sure what happens when css and jss collide, but if the css classname remains referenceable in the component, then an implementation might still use their own css.

PR #2734, opened 2019-06-17
Adjust styling for 'more' link in content search results

### LINKED ISSUE #2726: "More" links should give specific context (plus styling)
1.  Concern: The "more..." option is not technically a button-within-a-button in the HTML, but it is still reading as such to a screenreader, because the parent item has `role="button"`. I don't know if this is illegal, but it's disconcerting. I don't know what to do about this. 

2. Actionable: When the "more..." link exists, it should include some context for the screenreader. Can do this with `<button aria-describedby="resultLabelID">` which should read something like "more [pause] Page 6". _(edited)_

3. Actionable: The mockups #2683 show lower-case "m" and an underline (I think the expectation was that this would be rendered as a link, not a button). The underline with the expanded letter-spacing of the button looks awkward to me. @ggeisler?

**Current**
Screen Shot 2019-06-14 at 3 55 50 PM
**Proposed**
Screen Shot 2019-06-14 at 3 50 41 PM
  - remove the underline
  - force the height of `<span class="MuiButton-label">` to `1.5em` so it doesn't make the last line of text hang lower than it's meant to
  - lower-case the "m"

camillevilla:
Closes #2726
This PR makes slight adjustments to the styling of the "More" link in content search results. @jvine, I wasn't sure if you meant a `height` or `line-height` of `1.5em`. Take a look in netlify and let me know if this works. 

## Before
links underlined

## After
linked lower cased with line removed

### DIFF
--- a/src/components/SearchHit.js
+++ b/src/components/SearchHit.js
@@ -36,6 +36,7 @@ export class SearchHit extends Component {
       annotationLabel,
       canvasLabel,
       classes,
+      companionWindowId,
       containerRef,
       hit,
       focused,
@@ -49,6 +50,7 @@ export class SearchHit extends Component {
 
     const truncatedHit = focused ? hit : hit && new TruncatedHit(hit);
     const truncated = hit && truncatedHit.before !== hit.before && truncatedHit.after !== hit.after;
+    const canvasLabelHtmlId = `${companionWindowId}-${index}`;
 
     return (
       <ScrollTo
@@ -73,7 +75,9 @@ export class SearchHit extends Component {
           <ListItemText primaryTypographyProps={{ variant: 'body1' }}>
             <Typography variant="subtitle2" className={classes.subtitle}>
               <Chip component="span" label={index + 1} className={classes.hitCounter} />
-              {canvasLabel}
+              <span id={canvasLabelHtmlId}>
+                {canvasLabel}
+              </span>
             </Typography>
             {annotationLabel && (
               <Typography variant="subtitle2">{annotationLabel}</Typography>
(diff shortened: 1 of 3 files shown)


jvine:
Huh. Turns out setting the height screws up the focus bubble thingie and makes it hard to see when the button has focus. Setting line-height keeps the bubble.

camillevilla:
Thanks! I've corrected this PR. For anyone else who's curious, the difference between `height` and `line-height` is visible when using keyboard navigation. The GIFs below show how the Material UI ripple animation is cut off by `height` but remains the same with `line-height`. 

#### height: 1.5m
height-example

#### line-height 1.5em
line-height-example

src/components/SearchHit.js
```diff
@@ -78,7 +78,7 @@ export class SearchHit extends Component {
           <SanitizedHtml ruleSet="iiif" htmlString={truncatedHit.after} />
           {' '}
           { truncated && !focused && (
-            <Button className={classes.inlineButton} onClick={showDetails} color="secondary" size="small">
+            <Button className={classes.inlineButton} onClick={showDetails} color="secondary" size="small" aria-describedby={canvasLabel}>
```
mejackreed: Should the `aria-describedby` reference an `ID` here?

https://www.w3.org/TR/wai-aria/#aria-describedby

camillevilla: 😅 alrighty @mejackreed, let me know what you think about this approach. @cbeer suggested I use companion window IDs to deal with the case where there are multiple content search windows open.

mejackreed: Looks good!

PR #2735, opened 2019-06-18
Add support for content search pagination

### LINKED ISSUE #2684: Add search result page navigation
My first designs actually just assumed access to the next page, so instead of pagination controls there was a "More results" action that would trigger fetching the next page of results:

Screen Shot 2019-06-14 at 9 42 37 AM

---

# Superseded information

This ticket assumes work outlined in #2683 has been completed.

If the search service returns results using a paged structure (example: https://exist.scta.info/exist/apps/scta-app/iiif2/lon/search?q=contradictio), not all result items are necessarily immediately available. But we do know how many pages the results have been split into, so we can provide paging navigation for the user.

Screen Shot 2019-06-11 at 4 16 02 PM

In the mockup above, the highlighted part in the search panel header has been added to the `hits` version of results described in #2683, because this search service returns [paged results](https://exist.scta.info/exist/apps/scta-app/iiif2/lon/search?q=contradictio). 

- [ ] Use the `firstPageSharp` and `lastPageSharp` icons for the outermost icons of the page navigation element. Use `padding: 12px;` for the icons.
- [ ] Icon tooltips for above: "First page" and "Last page".
- [ ] For the inner icons, use the `chevronLeftSharp` and `chevronRightSharp` icons, with `padding: 12px;` for previous and next page navigation.
- [ ] Icon tooltips: "Previous page" and "Next page".
- [ ] Between the next and previous icons, the text "Page x of y", where x = currently selected page and y = total number of pages.
- [ ] Center the page navigation with the panel header. Note that in the mockup it appears that the outermost icons are aligned with the left and right edges of the panel header content, but that's just the way the spacing worked out. The page nav element should be centered and the icons spaced outward based on icon width + the `12px` padding.

### Page navigation interaction
This probably doesn't need much explanation, but when the user navigates to a new page, the page navigation element, the hit result navigation element, the panel body content, and the canvas should all be updated accordingly.

For example, in the mockup above the user is on page 2 of the results, which this search service returns in pages of 5. If the user then selects the next page icon:

- the page navigation changes to "Page 3 of 7" 
- the hit result navigation changes to "11 of 33"
- the panel body content changes to show hit content of result items 11 through 15
- the canvas changes to show the canvas on which result item 11 appears

### LINKED ISSUE #2731: Add notification dot to Search icon when panel has results
It could be helpful to the user if we add a notification dot to the window search icon when the window's object has active search results (or other content in the search panel body, such as the suggested query link as we are planning in an [embedded use of the viewer](https://github.com/sul-dlss/sul-embed/issues/1019)). This would provide a clue, or reminder, to the user that there is something active in the search panel.

I believe this could be pretty straightforward:

- [ ] If the search panel body is not empty, show the notification dot. (It's a borderline case, but I would consider the "No results found" situation as panel not empty, thus still requires the dot.)

- [ ] If the search panel body is or becomes empty (initial view of window when no search has yet been performed; user clears the search using the clear action or removing search terms from search input), remove the notification dot.

For illustration, imagine the user opens an object in a window and navigates to the search panel. No notification dot since search panel body is empty:

Screen Shot 2019-06-17 at 3 57 42 PM

---

User then conducts a search, results come back, notification dot is thus displayed:

Screen Shot 2019-06-17 at 3 50 55 PM

---

User leaves search query as-is, navigates to the metadata panel, notification dot still displayed since, although it's not visible, there is still content in the search panel body:

Screen Shot 2019-06-17 at 3 52 18 PM

---

User then returns to search panel, clears the search, notification dot is removed:

Screen Shot 2019-06-17 at 4 02 14 PM

### LINKED ISSUE #2739: Support IIIF Search implementations that only provide annotation lists

cbeer:
Fixes #2684 
Fixes #2739
Fixes #2731

Screen Shot 2019-06-18 at 13 25 19

### DIFF
--- a/src/state/reducers/search.js
+++ b/src/state/reducers/search.js
@@ -11,10 +11,16 @@ export const searchesReducer = (state = {}, action) => {
     case ActionTypes.REQUEST_SEARCH:
       return {
         ...state,
-        [action.targetId]: {
-          ...state[action.targetId],
+        [action.windowId]: {
+          ...state[action.windowId],
           [action.companionWindowId]: {
-            isFetching: true,
+            ...(state[action.windowId] || {})[action.companionWindowId],
+            data: {
+              ...((state[action.windowId] || {})[action.companionWindowId] || {}).data,
+              [action.searchId]: {
+                isFetching: true,
+              },
+            },
             query: action.query,
           },
         },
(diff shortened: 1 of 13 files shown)


src/state/reducers/search.js
```diff
@@ -11,44 +11,60 @@ export const searchesReducer = (state = {}, action) => {
...
       return {
         ...state,
-        [action.targetId]: {
-          ...state[action.targetId],
+        [action.windowId]: {
+          ...state[action.windowId],
           [action.companionWindowId]: {
-            isFetching: true,
+            ...(state[action.windowId] || {})[action.companionWindowId],
+            data: {
```
mejackreed: I'm noticing that the state change here (while useful for pagination) tanks Mirador when multiple queries are made within the same companion window.

Not only does the memory increase (and hold on until companion window is removed), also the search hits hang on indefinitely and are not updated.

memory

The mirador instance has crashed a few times on me.

PR #2746, opened 2019-06-18
Add aria-live="assertive" for SearchHits

### LINKED ISSUE #2717: Screen-reader accessibility of search results navigation
Currently, the Next/Previous navigation is keyboard accessible, but not for screen readers. The SR hears the label, and knows it's a button, and can click it; but no change is perceived. I think the correct way to address this is with an `aria-live` region. 

It seems like it should work to dynamically attach `<aria-live="assertive">` to the `<li>` of the current result on button press. User tabs to button, clicks it, SR reads the contents of the list item. It does work in Chrome on VO, but only the first time the button is pressed (maybe the SR doesn't perceive it as a "change" if you go back). Having trouble with other browsers. 

Anyone have other ideas on this? Goal is to have the content of the result read when the user clicks the button, without changing the focus to that content (and away from the button).

mejackreed:
Fixes #2717 

@jvine is this doing the correct thing?

### DIFF
--- a/__tests__/src/components/SearchHit.test.js
+++ b/__tests__/src/components/SearchHit.test.js
@@ -37,9 +37,15 @@ describe('SearchHit', () => {
     expect(selectContentSearchAnnotation).toHaveBeenCalledWith('window', ['foo']);
   });
 
+
   it('renders the annotation char if the hit is not available', () => {
     const wrapper = createWrapper({ annotation: { chars: 'xyz' }, hit: undefined });
     expect(wrapper.find('WithStyles(ForwardRef(ListItemText))').render().text()).toEqual('1xyz');
+  });
+
+  it('has aria-live="assertive"', () => {
+    const wrapper = createWrapper();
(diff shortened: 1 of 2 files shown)


jvine:
Something weird is happening with VoiceOver. It _mostly_ works (click next, it reads the content of the selected result) but it fairly often reads the wrong entry - typically repeats one then skips one to make up for it, or stays one behind the current highlight for a few clicks. I can't tell if it's a problem with the approach, or with VO.

jvine:
Okay, this is not working at all in NVDA on Windows, and does weird sputtery things in Chromevox. Occasionally CV picks up the "more..." or only the words following the bolded term - as if it's struggling with the structure inside the content. But mostly reads nothing. NVDA reads nothing consistently. I don't know how the code is indicating which of the aria-live regions should be read, so I can't tell which part of that might not be working. I would have expected to assign aria-live to one result at a time, but I don't honestly know if that would work either. All the examples I've seen have a single aria-live region and dynamically change the content inside it. This is currently mostly working in one known screenreader. I am at a loss.

jkeck:
Closed by #2829

PR #2748, opened 2019-06-18
Support IIIF Search endpoints that only return annotation lists

### LINKED ISSUE #2739: Support IIIF Search implementations that only provide annotation lists

cbeer:
Fixes #2739

### DIFF
--- a/src/containers/SearchResults.js
+++ b/src/containers/SearchResults.js
@@ -6,6 +6,7 @@ import { withPlugins } from '../extend/withPlugins';
 import { SearchResults } from '../components/SearchResults';
 import * as actions from '../state/actions';
 import {
+  getSearchAnnotationForCompanionWindow,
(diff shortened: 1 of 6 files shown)


src/containers/SearchResults.js
```diff
@@ -17,6 +18,8 @@ import {
  * @private
  */
 const mapStateToProps = (state, { companionWindowId, windowId }) => ({
+  searchAnnotations:
+    getSearchAnnotationForCompanionWindow(state, { companionWindowId, windowId }).resources,
```
mejackreed: Seems like that `getSearchAnnotationForCompanionWindow` can return `undefined` which will cause an issue with opening the panel.

cbeer:
#2735 doesn't have that problem, so let's just go with the combined PR instead of trying to do it piecemeal.

PR #2749, opened 2019-06-18
Update styling of content search result items; fixes #2736

### LINKED ISSUE #2736: Update styling of content search result items
Here are a handful of styling suggestions for the search panel body when there are search results, based on the current layout, which looks like this:

Screen Shot 2019-06-17 at 5 42 54 PM

---

- [x] Remove the `padding-top: 8px;` on `.MuiList-padding` so that we don't get that gap of white between the top of the first result item and the bottom of the panel header (callout A in screenshot above).

- [x] If possible, reduce the `height` of the hit counter background circle (`MuiChip-label`) to `30px` (callout B).

- [x] Change the typography for the hit counter and canvas label to `Subtitle 2` (callouts B and C).

- [x] Add bottom margin to the `<p />` that holds the hit counter and canvas label. I'd suggest `10px` (or `12px` if it's easier/more consistent to stick to multiples of 4) (callout D).

- [x] Change the typography of the hit result text to `body 1` (I originally suggested the smaller `body 2` but we want to be consistent with font size of text on other panels) (callout E).

- [x] Reduce the `padding-right` on the `.MuiListItem-gutters` to `8px` (callout F).

- [x] Add divider between result items (callout G). I showed this in my mockups in #2683 but I think I neglected to include it as a task item on that ticket. Basically just the same approach as we use in the list of annotation items on the annotations panel. Whatever we do there should work here.

Those updates should result in the panel looking something like this:

Screen Shot 2019-06-17 at 5 38 07 PM

cbeer:


src/containers/SearchHit.js
```diff
@@ -45,13 +45,11 @@ const mapDispatchToProps = {
...
   adjacent: {},
-  canvasLabel: {
-    ...theme.typography.h6,
-  },
   focused: {},
   hitCounter: {
-    ...theme.typography.h6,
+    ...theme.typography.subtitle2,
     backgroundColor: theme.palette.hitCounter.default,
+    height: '30',
```
mejackreed: I'm seeing an issue with this not getting applied... maybe it needs to be an Integer or `'30px'`?

Screen Shot 2019-06-18 at 5 07 29 PM

Issue #2754, opened 2019-06-19
Allow a mirador window to be initialized with a content search query


jkeck:
This is in order to support the scenarios laid out in:
- [x] sul-dlss/sul-embed#1019
- [x] sul-dlss/sul-embed#1022
- [ ] sul-dlss/sul-embed#1023

Issue #2664, opened 2019-06-10
Initialize a Mirador window with a Canvas URI


regisrob:
There should be a way to initialize Mirador with a given Canvas URI, in addition to the existing `CanvasIndex` config option (maybe it is already there but I did not find how to do this, or maybe it is already planned in the dev roadmap...)

camillevilla:
@regisrob We haven't implemented this, and I don't think we have specific plans for this feature. Can you tell me more about how you would like to use this feature? 

Were you thinking of something like:
```javascript
windows: [{
           loadedManifest: 'http://example.org/iiif/book1',
           canvasURI: https://example.org/iiif/book1/canvas/p1,
         },
		 {
           loadedManifest: 'http://example.org/iiif/book2',
           canvasIndex: 2,
         }],
```

regisrob:
@camillevilla yes exactly. 

I understand the Canvas URI as the reference identifier of a page/view within the IIIF world, so I think there are many situations where it would be expected to initialize a viewer with this URI (given that, from a Mirador implementer's perspective, it might be the only information you have in hand to point to a given page/view within a Manifest). This approach is used in various implementations:
- e.g. in the Mirador instance at Durham: https://iiif.durham.ac.uk/index.html?manifest=t1m4q77fr328&canvas=t1txd07gs99c
- or in this Biblissima demo https://demos.biblissima.fr/ovide-moralise/, where the whole interface is built upon IIIF annotations stored in a triplestore: in this case the Manifest and Canvas URIs that come with every annotation are used to initialize a Mirador "Compare view" like this: https://demos.biblissima.fr/ovide-moralise/mirador/?m0=btv1b8626777x&c0=f95&m1=btv1b8626780d&c1=f89
- another use case in the Biblissima portal, where we have a web pages for each illumination described in manuscripts http://beta.biblissima.fr/en/ark:/43093/ifdata51e80d55581c119cee1834e7af836366ced5f049. Here the information we store in the database is the Canvas URI of the folio bearing the illumination, and this is what we use to refer to this folio in the drag'n drop URL behind the IIIF logo, and also in the basket functionality which let the user select the illuminated folio and open it in Mirador in another tab to be able to compare with other illuminations

I am sure there are quite a lot of other use cases like this in the Mirador and IIIF community at large.

This kind of requirement might also come up in the future when implementing the upcoming Content State API: https://preview.iiif.io/api/import-to-viewers/api/content-state/0.2/#21-annotation-model-for-content-state

PR #2756, opened 2019-06-19
Make the window configuration open for extension

cbeer:
Loosely connected to #2754

Also, allow the window settings to include additional companion windows.


mejackreed:
I wonder where something like this could fit in https://github.com/ProjectMirador/mirador/issues/2664

Seems similar to the "search" API problem. But not sure how it could/would look. Obviously not the problem of this PR, but just for added context.

jkeck:
Included in #2772

PR #2760, opened 2019-06-20
Add config option to initially show manifest list rather than windows. Closes #2759

### LINKED ISSUE #2759: Add config option to initially show manifest list at rather then windows (or blank workspace)

charbugs:
Example usage:

```javascript
const config = {
  id: 'mirador',
  workspace: {
    isWorkspaceAddVisible: true,  // false by default
  },
  manifests: {
    "https://data.ucd.ie/api/img/manifests/ucdlib:33064": { provider: "Irish Architectural Archive"},
    "https://wellcomelibrary.org/iiif/b18035723/manifest": { provider: "Wellcome Library"},
  },
}
```

### DIFF
--- a/__tests__/integration/mirador/index.html
+++ b/__tests__/integration/mirador/index.html
@@ -13,14 +13,9 @@
     <script type="text/javascript">
      var miradorInstance = Mirador.viewer({
        id: 'mirador',
-       windows: [{
-         loadedManifest: 'https://iiif.harvardartmuseums.org/manifests/object/299843',
-         canvasIndex: 2,
-         thumbnailNavigationPosition: 'far-bottom',
+       workspace: {
+         isWorkspaceAddVisible: true,
        },
-       {
-         loadedManifest: 'https://iiif.bodleian.ox.ac.uk/iiif/manifest/e32a277e-91e2-4a6d-8ba6-cc4bad230410.json',
-       }],
        window: {
          hideSearchPanel: false,
        },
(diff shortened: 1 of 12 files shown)


cbeer:
Instead of all this shuffling and refactoring, would it be better to just have the workspace reducer handle the `SET_CONFIG` action too, say something like:

```
--- a/src/state/reducers/workspace.js
+++ b/src/state/reducers/workspace.js
@@ -94,6 +94,8 @@ export const workspaceReducer = (
       return { ...state, exposeModeOn: !state.exposeModeOn };
       return action.state.workspace;
+    case ActionTypes.SET_CONFIG:
+      return action.config.workspace;
     case ActionTypes.TOGGLE_DRAGGING:
       return { ...state, draggingEnabled: !state.draggingEnabled };
     default:
```

I could see this approach working for some of the other reducers too (thinking `elasticLayout` and `viewers` in particular)

charbugs:
This is an approach worth thinking about, but currently no reducer follows this, so I oriented myself on the way the config.workspace.type is managed.

cbeer:
It's definitely confusing how the different properties are spread across state. Maybe this is a good opportunity to develop a general pattern for it.

marlo-longley:
The architecture proposal here was superseded by https://github.com/ProjectMirador/mirador/pull/2827
The question of how to initially show manifest list rather than windows was answered here: https://github.com/ProjectMirador/mirador/issues/2759

PR #2767, opened 2019-06-20
Add a top bar menu that only appears when plugins have been …

### LINKED ISSUE #2757: Break out Window options menu into two separate menus
Based on usability feedback. Goal is to separate the current Window options menu into two menus:  

### 1. Window views & thumbnail display
   - always present
   - contains the View and Thumbnails sections only
   - tooltip is "Window views & thumbnail display"

view menu

Bonus points in this menu:
  - [ ] shorten the text of the headers as shown in the screenshot
  - [ ] apply body1 typography to the headers
  - [ ] apply body2 typography to the icon text
(Not just for aesthetics, but based on user response re what is perceived first.)

### 2. Window options
   - present if there are plugins to populate it
   - simple icon + text menu layout
   - tooltip remains "Window options"
   - no longer needs the "Options" header within the menu

options menu

### Icon svg
```
<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24">
    <path d="M0,0H24V24H0Z" fill="none"/>
    <path d="M24.852,17.981,3,18V15.945l21.852-.019Z" transform="translate(-2 5)"/>
    <path d="M1,4V21H22.853V4.1ZM3,19V6h8V19Zm18,0H13V6h8Z" transform="translate(0 -2)"/>
</svg>
```

jkeck:
…registered to it.

Closes #2757 

This isn't possible to see in action unless you have a plugin installed to this component (which I've done locally)
Screen Shot 2019-06-21 at 12 21 17 PM


mejackreed:
Should we drop something into `__tests__/integration/mirador/plugins/add.html`?

jkeck:
Ah, there wasn't anything analogous for the WindowTopBarMenu so I never came across that.  I can drop something in there for testing.

jkeck:
Which would be more useful if it weren't completely broken.  Added anyway.

PR #2778, opened 2019-06-22
Add a CircularProgress indicator around the Search button while the …

jkeck:
…search is active fetching.

Just a rando idea from looking through the MUI documentation.


mejackreed:
I like it!

PR #2781, opened 2019-06-24
Update settings documentation for window

mejackreed:
There seems to be some complexity about how all of this happens, just documenting it for now

### DIFF
--- a/src/config/settings.js
+++ b/src/config/settings.js
@@ -221,7 +221,18 @@ export default {
     hideWindowTitle: false, // Configure if the window title is shown in the window title bar or not
     sideBarOpenByDefault: false, // Configure if the sidebar (and its content panel) is open by default
   },
-  windows: [], // Array of windows to be open when mirador initializes (each object should at least provide a loadedManifest key with the value of the IIIF presentation manifest to load)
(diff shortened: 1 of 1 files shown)


src/config/settings.js
```diff
@@ -221,7 +221,18 @@ export default {
...
+  windows: [ // Array of windows to be open when mirador initializes (each object should at least provide a manifestId key with the value of the IIIF presentation manifest to load)
+    /**
+    Example Window:
+    {
+      manifestId: 'https://iiif.harvardartmuseums.org/manifests/object/299843',
+      canvasIndex: 2,
+      thumbnailNavigationPosition: 'far-bottom',
+    }
+    // https://github.com/ProjectMirador/mirador/blob/master/src/state/actions/window.js#L48-L65
+    // https://github.com/ProjectMirador/mirador/blob/master/src/lib/MiradorViewer.js#L53-L62
```
jkeck: These links are bound to change.  Should we use the permalink syntax instead?

camillevilla: Er. Well... I guess the commit that gets permalinked will also get out of date. Maybe it would be better to just link to the file and explicitly call out `defaultOptions` and `windowAction`?

mejackreed: Thanks both of you for the suggestions. I've updated to use relative references to the path of the file and explicit variable names.

PR #2784, opened 2019-06-24
Add a little padding to content search input

### LINKED ISSUE #2738: Add padding to search input in Search panel header
In its initial, empty state, the search input is flush against both the bottom of the search panel header and the right edge of the panel header. 

Screen Shot 2019-06-18 at 9 07 34 AM

It would look better with some whitespace between the edges of the input and the panel:

Screen Shot 2019-06-18 at 9 07 18 AM

To get the layout in the screenshot above, I updated the `MuiFormControl-root` div at the top of the `form` element holding the input like this:

- [ ] `padding-right: 12px;`
- [ ] `padding-bottom: 8px;`

Although someone more familiar with the code might see better/more consistent ways to achieve the same effect.

camillevilla:
Closes #2738
This PR just adds a little bit of padding to the content search text input.

### DIFF
--- a/src/containers/SearchPanelControls.js
+++ b/src/containers/SearchPanelControls.js
@@ -33,6 +33,10 @@ const mapDispatchToProps = {
 
 /** */
 const styles = theme => ({
+  searchInput: {
+    paddingBottom: theme.spacing(1),
+    paddingRight: theme.spacing(1.5),
+  },
   suggestions: {
     position: 'absolute',
     zIndex: theme.zIndex.modal - 1,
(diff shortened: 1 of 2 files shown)


src/containers/SearchPanelControls.js
```diff
@@ -33,6 +33,10 @@ const mapDispatchToProps = {
 
 /** */
 const styles = theme => ({
+  searchInput: {
+    paddingBottom: '8px',
```
cbeer: Should we use `theme.spacing`?

PR #2789, opened 2019-06-25
Use linear and radial gradients to try to indicate there is scrollable...

### LINKED ISSUE #2774: Invisible scrollbars causing usability issues
Some things just need to scroll, like a complex dialog in a very small window. 

If a user has their scrollbars set to show only when scrolling... they don't see any affordance for scrolling in the dialog and don't realize there are more options available. 

Is there a way to force the display of scrollbars inside dialogs? Found this but it doesn't seem robust:
https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar

jkeck:
...content in dialogs

Closes #2774

scrollShadows

Using the component in a plugin:

Screen Shot 2019-06-25 at 8 20 59 PM

### DIFF
--- a/__tests__/src/components/ScrollIndicatedDialogContent.test.js
+++ b/__tests__/src/components/ScrollIndicatedDialogContent.test.js
@@ -1,0 +1,37 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import DialogContent from '@material-ui/core/DialogContent';
+import { ScrollIndicatedDialogContent } from '../../../src/components/ScrollIndicatedDialogContent';
+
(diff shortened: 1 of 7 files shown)


jkeck:
@jvine this is a bit of a different approach taking advantage of the `background-attachment: local` property and linear + radial gradients, let me know what you think.

More details about approach: http://lea.verou.me/2012/04/background-attachment-local/

jkeck:
I believe the reduced coverage notice is bogus.  The only uncovered line it is reporting is because I removed a `className` from a button...

PR #2794, opened 2019-06-25
Separate selected search state per search

### LINKED ISSUE #2786: Maintain separate content search pagination state for each companion window
Currently, pagination state uses the single selected annotation. In a multiple-search scenario, the pagination of one search gets reset when pagination occurs in another companion window.

cbeer:
Fixes #2786 

In talking with @jvine, we agreed there should only be one selected search annotation per window (highlighted in yellow in the companion window and on the canvas). To support consistent pagination with multiple searches open, we also need to track the annotation at the search level.

### DIFF
--- a/src/state/reducers/search.js
+++ b/src/state/reducers/search.js
@@ -7,16 +7,17 @@ import ActionTypes from '../actions/action-types';
  * searchReducer
  */
 export const searchesReducer = (state = {}, action) => {
+  const searchStruct = (state[action.windowId] || {})[action.companionWindowId] || {};
   switch (action.type) {
     case ActionTypes.REQUEST_SEARCH:
-      if (((state[action.windowId] || {})[action.companionWindowId] || {}) !== action.query) {
+      if (searchStruct.query !== action.query) {
         // new query
         return {
           ...state,
           [action.windowId]: {
             ...state[action.windowId],
             [action.companionWindowId]: {
-              ...(state[action.windowId] || {})[action.companionWindowId],
+              ...searchStruct,
               data: {
                 [action.searchId]: {
                   isFetching: true,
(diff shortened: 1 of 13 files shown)


src/state/selectors/searches.js
```diff
@@ -122,11 +122,13 @@ export const getSearchAnnotationsForWindow = createSelector(
 export const getSelectedContentSearchAnnotationIds = createSelector(
   [
     getWindow,
+    getSearchForCompanionWindow,
   ],
-  window => (window && window.selectedContentSearchAnnotation) || [],
+  (window, search) => (search && search.selectedContentSearchAnnotation)
```
cbeer: FWIW, I'd be easily persuaded that selected-for-search and selected-for-window are different and should have different selectors. I went back-and-forth on that and decided not to care (yet)

jkeck: 👍 I think it's fine to make that distinction when it become necessary.

src/state/reducers/search.js
```diff
@@ -9,25 +9,39 @@ import ActionTypes from '../actions/action-types';
 export const searchesReducer = (state = {}, action) => {
   switch (action.type) {
     case ActionTypes.REQUEST_SEARCH:
+      if (((state[action.windowId] || {})[action.companionWindowId] || {}) !== action.query) {
```
jkeck: What would you think about pulling `(state[action.windowId] || {})[action.companionWindowId] || {}` this up into a named constant (since it appears to be in at least 8 places in this file).

Not considering this a blocking comment, it was already duplicated a bit, but it might benefit from giving it a name to help follow along w/ the code.

cbeer: Done.

PR #2807, opened 2019-06-26
Set stripAuthentication to false to avoid a SecurityError in microsoft browsers

cbeer:


jkeck:
Travis was 🍏, just not reporting.

PR #2827, opened 2019-06-28
Refactor how we're handling workspace configuration/state

cbeer:
An alternative approach to #2760.

This opens up the workspace config for easy extension and predictable actions for reducers to handle.  The `updateWorkspace` action can consolidate several other action types.

By having the workspace reducer handle `UPDATE_CONFIG`, we allow implementations to configure the default workspace behavior as part of the viewer initialization.


charbugs:
I think there shouldn't be a config reducer at all. There should be a `defaultConfig` object and a `customConfig` object that will initially merged into a `mergedConfig` object. Next a `setInitialState` function should distribute the values of `mergedConfig` to the reducers.

cbeer:
That sounds pretty similar to what we already have happening in `MiradorViewer`, yes?

src/config/settings.js
```diff
@@ -234,8 +234,15 @@ export default {
...
   workspace: {
+    exposeModeOn: false,
+    height: 5000,
     showZoomControls: false, // Configure if zoom controls should be displayed by default
     type: 'mosaic', // Which workspace type to load by default. Other possible values are "elastic"
+    viewportPosition: {
+      x: 0,
+      y: 0,
+    },
+    width: 5000,
```
jkeck: Would you mind adding few comments here to what these configuration options do?

charbugs:
Yes, with the difference that currently the mergedConfig object will be written to a config reducer. The problem with that is that the config reducer semantically overlaps with other reducers, so that one responsibility (e.g. workspace) is backed by two reducers (e.g. workspace and config reducer). I think it is cleaner to spread out the config data to the respective reducers.

PR #2828, opened 2019-06-28
Use separate i18n instances for each app; fixes #2782

### LINKED ISSUE #2782: You cannot have Mirador with multiple languages in the same page
I noticed while setting up https://mirador-start.glitch.me that you cannot have two Mirador instances with different languages in the same page. Maybe this isn't a problem?

cbeer:
I also updated the `embedding.html` page to demonstrate this behavior.

### DIFF
--- a/__tests__/src/components/App.test.js
+++ b/__tests__/src/components/App.test.js
@@ -7,7 +7,6 @@ import AuthenticationSender from '../../../src/containers/AuthenticationSender';
 import WorkspaceArea from '../../../src/containers/WorkspaceArea';
 import { App } from '../../../src/components/App';
 import settings from '../../../src/config/settings';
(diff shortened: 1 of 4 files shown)


__tests__/src/components/App.test.js
```diff
@@ -45,8 +44,9 @@ describe('App', () => {
   });
 
   it('sets up translations based on the config passed in', () => {
-    createWrapper({ translations: { en: { off: 'on' } } });
-    expect(i18n.t('off')).toEqual('on');
+    const wrapper = createWrapper({ translations: { en: { off: 'on' } } });
+    console.log(wrapper.instance);
```
jkeck: Debug cruft?

cbeer: 👍

PR #2829, opened 2019-06-28
Add react-aria-live and use it to broadcast search result selec…

jkeck:


jvine:
Tested the aria-live in NVDA / Chrome on Windows, and it worked as expected.

PR #2832, opened 2019-07-01
[GAIN CONSENSUS] Removes sass in favor of global jss solution

mejackreed:
The intent here was to reduce the overall bundle size by moving everything to JSS.

At this point, the size increases :(

The first commit drops the size by about 500 bytes... the additional commits increase it by about 61KB. (uncompressed)

However! Compressed (gzip) this does reduce the bundle size by about ~1%

413KB -> 409KB

Should we do this? Not really sure. There may be other opportunities to tune down some of the styles used throughout also.

## Pros 
- Slight reduction in compressed size
- Consistent approach to stylin' 💅 
- Reduction of build complexity
- Unrealized future pro: move "class" styling to more component based
- Allows us to strip out some of the unused theme styles from react-mosaic

## Cons
- Potential maintenance of jss vendor styles
- Slight increase of uncompressed size


src/styles/vendor/react-mosaic-jss.js
```diff
@@ -0,0 +1,295 @@
...
+    '.mosaic-drop-target .drop-target-container .drop-target.drop-target-hover.bottom': {
+      top: 'calc(100% -  50% )',
+    },
+    '.mosaic-drop-target .drop-target-container .drop-target.drop-target-hover.top': {
+      bottom: 'calc(100% -  50% )',
+    },
+    '.mosaic-window, .mosaic-preview': {
+      position: 'relative',
+      display: 'flex',
+      fallbacks: [
```
christopher-johnson: this fallbacks array does not serialize with JSS.

src/styles/vendor/react-mosaic-jss.js
```diff
@@ -0,0 +1,295 @@
...
+      ],
+      W: 'normal',
+      flexDirection: 'column',
+      overflow: 'hidden',
+      boxShadow: '0 0 1px rgba(0, 0, 0, 0.2)',
+    },
+    '.mosaic-window .mosaic-window-toolbar, .mosaic-preview .mosaic-window-toolbar': {
+      zIndex: '4',
+      display: 'flex',
+      fallbacks: [
```
christopher-johnson: same as above

src/containers/WorkspaceArea.js
```diff
@@ -24,6 +26,103 @@ const mapStateToProps = state => (
...
+    },
+    '.mirador-workspace-with-control-panel': {
+      paddingTop: 74,
+      [theme.breakpoints.up('sm')]: {
+        paddingLeft: 68,
+        paddingTop: 0,
+      },
+    },
+    '.mosaic-tile': {
+      boxShadow: [
```
christopher-johnson: does not serialize

src/styles/vendor/react-mosaic-jss.js
mejackreed: I've updated this in latest commit

src/styles/vendor/react-mosaic-jss.js
mejackreed: I've updated this in latest commit

src/containers/WorkspaceArea.js
mejackreed: Resolved in latest commit. Seemed to be a case of merge ordering.

christopher-johnson:
The deploy preview is broken, so it seems that mosaic needs these fallbacks.  I guess that you need to add the [jss expand plugin](https://cssinjs.org/jss-plugin-expand?v=v10.0.0-alpha.22)  which can compile object arrays.  It seems that this is not included in the default material-ui styles presets.  Check [here](https://material-ui.com/styles/advanced/#jss-plugins) for example of how to do this.

PR #2852, opened 2019-07-05
Remove react-svg-unique-id and convert svgs back to original format

mejackreed:
This removes a dependency that I'm wasn't able to determine that we
needed. It also aims to make the svgs simpler. The only modification
done here was to set the width and height.

Saves 35KB

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,6 @@
     "react-resize-observer": "^1.1.1",
     "react-rnd": "^9.2.0",
     "react-sizeme": "^2.6.7",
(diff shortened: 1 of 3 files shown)


cbeer:
I'm late to  the party -- is this going to cause any validation problems with multiple miradors on the same page?

mejackreed:
Maybe? I think this would only come into play if both modals are rendered at the same time on the same page (multiple id's in DOM at same time). But I'm not sure if that is possible?

PR #2855, opened 2019-07-07
Add support for multiple image layers

### LINKED ISSUE #2607: Add layers panel to window (story)
This is a story ticket that provides some general background. The specific relevant work tickets are:

- [x] #2608 Add layers menu icon
- [x] #2609 Enable drag and drop of layers
- [x] #2610 Add layer thumbnail and label
- [x] #2611 Enable toggling layer visibility
- [x] #2612 Add move-to-top action for layers
- [x] #2613 Add layer opacity controls
- [x] #2614 Add layer section divider

### Background
Layers is an existing feature of Mirador 2. The layer features we suggest should be core in Mirador 3 are very similar, but for Mirador 3 there are a couple of significant updates:

- The list of layers is stacked, as with layers in an application like Photoshop, such that from a z-index perspective the layer at the top of the list in the layer panel has the highest z-index value. The last layer in the list has the lowest z-index value of all layers in the list. This means that, unlike in Mirador 2, if all layers are visible and at 100% opacity, you will only see the image in the topmost layer.

- To give users necessary flexibility for image analysis, layers should be moveable, so the relative z-index of a given layer can effectively be changed by the user.

Of the existing panel types, the layers panel is most similar to the annotations panel. Just as each annotation gets its own divided section in the annotations panel, each layer of an object gets its own divided layer section. For example, here's an object with three layers:

layers-panel

---

The work tickets linked above include adding the layers panel and the following layers features:

- List all layers by thumbnail and label
- Toggle visibility of layer
- Adjust layer opacity
- Reorder layers
- Move layer to top

Here's an [interactive mockup](https://xd.adobe.com/view/c0b13056-4bea-4164-78a6-c716d635089a-b680/) of those features.

### Potentially useful sample manifests with layers defined
Note that at the time of writing this ticket, only the first of the manifests below loads images in Mirador 3; all but the last work, including displaying layers, in Mirador 2.

- [Reconstructed manifest (partial): Grandes Chroniques de France](https://demos.biblissima.fr/iiif/metadata/BVMM/chateauroux/manifest.json)(Multiple canvases, each with two layers)
- [Letter from Alexander Hamilton Papers](https://prtd.app/hamilton/manifest.json) (One canvas with many multi-spectral layers)
- [Ars memorandi per figuras Evangelistarum](https://prtd.app/aom/manifest.json) (Several canvases, each with many multi-spectral layers)
- [Flute Vault](https://prtd.app/fv/manifest.json) (One canvas, two layers)
- [Osborn fa1 IIIF Manifest created for the Digitally Enabled Scholarship with Medieval Manuscripts project at Yale University](https://manifests.britishart.yale.edu/Osbornfa1) (Multiple canvases, one of which has many layers, but doesn't currently load in either Mirador 2 or Mirador 3)

cbeer:
Fixes #2607

This PR implements basic support for canvases with multiple images. It supports two different variants found in the wild without distinction: a) canvases with multiple image resources, and b) canvases with an `oa:Choice` annotation referencing multiple image resources.

The state is modeled such that, without interacting with the layers controls, we can avoid storing any layer state at all, but once interacted with, we persist the layers for the lifetime of the window.

Potential future work includes:
- handling images annotated onto fragments of the canvas (like https://prtd.app/hamilton/manifest.json)
- creating special behavior or additional structure for `oa:Choice` annotations
- optimizing the first-load experience for canvases with lots of images (`WindowViewer` currently blocks until all the `info.json` documents are retrieved..)


package.json
```diff
@@ -42,13 +42,14 @@
...
     "lodash": "^4.17.11",
-    "manifesto.js": "^3.0.9",
+    "manifesto.js": "^3.0.11",
     "node-fetch": "^2.6.0",
     "node-sass": "^4.12.0",
     "normalize-url": "^4.2.0",
     "openseadragon": "^2.4.1",
     "prop-types": "^15.6.2",
     "react-aria-live": "^2.0.5",
+    "react-beautiful-dnd": "^11.0.4",
```
mejackreed: Would it be possible to integrate with existing DnD libraries we are using? `react-rnd` or `react-dnd` (dependency of react-mosaic)

Just curious since we are using a few different DnD libraries now. However, react-beautiful-dnd seems really nice.

cbeer: Not really. `react-rnd` and `react-dnd` are pretty low-level and generic. `react-beautiful-dnd` is tailored for sorting lists of things, and so it does a bunch of the high-level functionality and provides a bunch of accessibility features for us... and, likewise, we can't replace those other libraries with it, because `react-beautiful-dnd` is pretty much exclusively about sorting lists of things.

ggeisler:
@cbeer Is there a specific manifest I can use to see what's been implemented? 

Perhaps there's something wrong at my end, but no matter which of the manifests I listed in the layers design ticket I load, I don't ever see the layers icon in the window sidebar.

mejackreed:
@ggeisler https://deploy-preview-2855--mirador-dev.netlify.com/layers.html

ggeisler:
@mejackreed Thanks but that URL gives me this:

Screen Shot 2019-07-08 at 10 26 02 AM

I think there is something I'm missing?

ggeisler:
@mejackreed Wait, tried it again and I get the full URL, which appears to work. Thanks!

ggeisler:
@cbeer This looks really great. The a11y is great, too (keyboard; we haven't tested with a screenreader yet). Below are a handful of styling recommendations, which hopefully will be relatively minor to add. 

Here's a screenshot with callouts that I refer to in my recommendations:

Screen Shot 2019-07-08 at 1 35 26 PM

- A: There's a bit of unwanted whitespace between the bottom of the Layers panel header and the top of the first layer, which I think might be fixed by removing the `padding-top: 8px` applied to the `<ul>` that contains the layers.
- B: The sidepanel feels a bit too dense. We can give it some breathing room by adding `8px` top and bottom padding to the contents of each layer section. Currently the `<div>` that holds the two inner `<div>`s with the layer content doesn't have a class, but I think if you add a class to it, or otherwise target that containing div, and the additional top and bottom padding there, that should improve things.
- C: Add a `border: 1px solid rgba(0, 0, 0, 0.12);` (same color as the layer divider border) to the layer thumbnails. For consistency, we probably should do the same for the thumbnails shown in the Index sidepanel; I can create a separate ticket for that but mention it here in case it affects how you add the styling. 
- D: The bottom border that separates each layer does not seem to extend to the gray draggable div associated with the layer. It would be better if each draggable div also had a visible bottom border.
- E: The length of the opacity slider appears to be dependent on the length of the layer title. The opacity slider for all layers should be the same length. Ideally that length would be a minimum of about `90px` (so it still fits when the sidepanel is at its minimum width) but would increase in width as the sidepanel width increases (to stay near the edge of the sidepanel), but only up to about `150px` max width (because it looks weird to keep getting wider when the sidepanel gets beyond a certain width).

src/lib/CanvasWorld.js
```diff
@@ -50,6 +55,60 @@ export default class CanvasWorld {
     return this.canvases.map(canvas => canvas.id).indexOf(canvasTarget);
   }
 
+  /** @private */
+  indexOfImageResource(imageId) {
```
mejackreed, line 59: I think I'm seeing an issue where `imageId` can be undefined in the console logs. Maybe its where I added the level 0 support in adding the non-tiled images resources here: https://github.com/ProjectMirador/mirador/blob/master/src/components/OpenSeadragonViewer.js#L223

Trying to think out loud, maybe we need to change how we do the level 0 support?

src/lib/CanvasWorld.js
```diff
@@ -20,8 +24,9 @@ export default class CanvasWorld {
    * canvasToWorldCoordinates - calculates the canvas coordinates respective to
    * the world.
    */
-  canvasToWorldCoordinates(i) {
+  canvasToWorldCoordinates(tileSource) {
     const wholeBounds = this.worldBounds();
+    const i = this.indexOfImageResource(tileSource['@id']);
     const canvas = this.canvases[i];
```
mejackreed, line 30: If there is no `i`, do we need to update how we book keep the canvas here also?

src/lib/CanvasWorld.js
cbeer: Do you happen to have an example manifest?

mejackreed: This one (level 0) https://github.com/IIIF/api/blob/master/source/api/presentation/2.1/example/fixtures/1/manifest.json 

But i'm seeing it also in https://prtd.app/hamilton/manifest.json

PR #2861, opened 2019-07-09
removes react-iframe-comm dependency

christopher-johnson:
`react-iframe-comm` bundles `react` and `react-dom`.   A [PR](https://github.com/pbojinov/react-iframe-comm/pull/14) towards resolving issue this upstream have had no response in several months, and it appears that this library is unmaintained.  this solves the problem (and consequently reduces Mirador bundle size by ~7k).  It removes the deprecated lifecycle `componentWillReceiveProps` from the component.  Also, this resolves a console warning about the inline attribute on Typography. in `WindowAuthenticationControl`


src/lib/IFrameComm.js
```diff
@@ -0,0 +1,201 @@
+import React, { Component } from 'react';
```
cbeer: This is code copied from `react-iframe-comm`? We probably need to preserve the upstream license information in this file (and any others we're copying over.).

src/lib/IFrameComm.js
```diff
@@ -0,0 +1,225 @@
...
+  SOFTWARE.
+ */
+
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+
+/**
+ *
+ */
+export class IframeComm extends Component {
```
mejackreed: Is the reason for this living in `src/lib` because it is code copied from upstream? I would expect it to be in `src/components`

src/lib/index.js
```diff
@@ -0,0 +1 @@
+export * from './IFrameComm';
```
mejackreed: Is this a new pattern that is needed?

__tests__/integration/mirador/window_actions.test.js
```diff
@@ -19,5 +19,5 @@ describe('Window actions', () => {
     )); // only default configed windows found
     await page.waitFor(1000);
     await expect(numWindows).toBe(2);
-  });
+  }, 10000);
```
mejackreed: Was this needed for the replacement of the iframe component? Were the tests failing?

src/lib/IFrameComm.js
christopher-johnson: lib is usually where compiled code goes so storing source in this directory is odd anyway, but in this case, since it is a "third party library component" (note also that it is a `derivation`, not a copy), it could/should reside in a directory that is distinct from your homegrown code (e.g. `vendors`) or even better, make a monorepo and build it as a separate package.

__tests__/integration/mirador/window_actions.test.js
christopher-johnson: no, nothing to do with the component.  these puppeteer tests consistently fail for me locally and also sometimes on CI.  increasing the timeout solves it.

christopher-johnson:
FYI:  I do not care whether this PR gets merged, but to resolve the issue, I suggest that you make whatever tangential changes you feel are appropriate.   I do not spend anymore time on this.

mejackreed:
Ok thanks @christopher-johnson . We will close this for now and revisit if we need to.

PR #2870, opened 2019-08-10
Update material-ui and remove react-placeholder dependency

mejackreed:
This PR updates Material-UI and adds Material-Labs. Also removes react-placeholder.

Reduces the minified gzipped bundle by another 2KB. 

The color might be a little bit different than the implementation from react-placeholder, so just want to get a reaction from @jvine or @ggeisler about that. Currently they use the hover color it seems to do the animation.

skeleton

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -4,7 +4,7 @@
     "@babel/plugin-transform-runtime",
     "transform-react-remove-prop-types",
     ["lodash", { "id": [
-      "lodash", "react-placeholder",
+      "lodash",
       ]}
     ]],
 }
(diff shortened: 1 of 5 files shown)


ggeisler:
@mejackreed Can you be more specific about what color you think is different in this PR?

From your description it sounds to me that you are saying you think the hover color (when you hover over an item in the list) is different, but if so, I believe they are the same (`rgba(0,0,0,0.08)` in both cases):

Screen Shot 2019-08-12 at 8 32 24 AM

But maybe I'm misunderstanding and you are talking about some other object's color being different?

mejackreed:
@ggeisler The color of the transition placeholder animation stuff. This is only present when manifests are loading still.

ggeisler:
Oh okay. The placeholder shapes that get replaced by real content. 

Yeah, I see the current color is a darker gray than in this PR. It doesn't bother me, especially since I think the actual permanent placeholders (e.g. the institution thumbnail placeholder if there isn't one available) remain the same gray. So I think we're just talking about the color that a user will only briefly see. And it is still gray, so yeah, okay with me.

PR #2875, opened 2019-09-29
Update plugin utilities

mejackreed:
Adds OSDReferences so that multiple plugins can use this.


src/components/PluginListItem.js
```diff
@@ -1,32 +0,0 @@
-import React from 'react';
```
jkeck: Was this file deleted intentionally?

mejackreed: Yep intentional in the first commit. I think this was added in the idea that it could be reusued, but I _think_ the list item implementation has changed underneath now and its no longer usable.

jkeck: Ah, yeah, I remember coming to a similar conclusion when building out some of our local plugins.

PR #2906, opened 2019-12-10
Add "sideEffects": false

mejackreed:
Since we are not remodifying imported packages, specify sideEffects false to improve build sizes downstream

https://webpack.js.org/configuration/optimization/#optimizationsideeffects

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
   "files": [
     "dist"
   ],
+  "sideEffects": false,
   "scripts": {
     "clean": "rm -rf ./dist",
     "lint": "node_modules/.bin/eslint ./ && npm run lint:translations",


cbeer:
From the docs:

> By default optimization.sideEffects is enabled in production mode and disabled elsewise.

So.. does this actually change the artifact sizes in a meaningful way? Or is something else fishy?

mejackreed:
My understanding is this.

> By default optimization.sideEffects is enabled in production mode and disabled elsewise.

Is only for a webpack config that is consuming other packages. The change intended in this PR is so that consumers of the Mirador package can get that optimization. Without specifying this in our `package.json`, Mirador will be skipped when trying to look for these optimizations.

PR #2910, opened 2019-12-11
Remove "core" module fixes #2905

### LINKED ISSUE #2905: Proposal: Remove m3core umd build
Ideally designed to be able to use state management without the UI, this can now be achieved by just using the ES modules from Mirador.

This proposal is to remove the extra published build so that we don't need to maintain future compatibility.

mejackreed:
This PR aims to simplify Mirador 3, by removing the extra core build. This was originally intended so that adopters could use the state management piece of Mirador without a UI. While this works, I'm not sure how useful this is for people over just using the ES builds of Mirador instead.


mejackreed:
There weren't any vocal objections to this on the 12/12/2019 community call. I'd say lets move forward with this.

charbugs:
Agree, I think the state logic is thightly bound to the overall implementation, so I see no use cases for the core module so far. And redoing that seems to be not that tough.

PR #2914, opened 2019-12-12
Infer sc:AnnotationList from otherContent strings

rwd:
From [IIIF Presentation API 2.1.1](https://iiif.io/api/presentation/2.1/#canvas):
> Note that the items in the otherContent list may be either objects with an @id property or strings. In the case of a string, this is the URI of the annotation list and the type of “sc:AnnotationList” can be inferred.


rwd:
@mejackreed sure thing, that's done.

PR #2916, opened 2019-12-19
Implementation of Table of Contents view

mejackreed:
A few things to call out. This PR implements a basic piece of table of contents view in Mirador 3. It uses the manifesto designed tree structure to recursively build a Material UI `TreeView`. 

I've found some serious performance issues when opening multiple windows at the same time. This isn't related necessarily to this work, but is already existing.

Todos (many now completed thanks to @lutzhelm )
 - [x] Open TOC to canvas where you are
 - [x] Save your place when you reopen or move a companion window
 - [x] Support for Presi v3 (not tested at least)
 - [x] Figure out how to use TOC when there are structures available #2948
 - [x] Better styling of this

@charbugs what are your plans on table of contents?


lutzhelm:
Sorry for the lack of progress and feedback on that issue, it took me longer than expected to familiarize myself with the general Mirador code. For the moment, I managed to mark the ranges that contain the currently visible canvas(es) (but without opening them yet).

Some thoughts and questions about the TOC:
* Currently, clicking on / selecting a tree node does only set the current canvas to a new value if that node is at the deepest level in the tree. This makes it easy to access the first of the contained canvases of said range but there might be situations where a range (e.g. a chapter of a book) contains canvases as well as another range that appears somewhere in the middle of the former range (e.g. an illustration). I guess it would be desirable to jump to the start of the former range.
* Something that probably concerns newspapers: if a range does consist of non consecutive canvases, would canvas nodes be a thing that could be necessary? Maybe as a configuration option?

lutzhelm:
I still have some changes hover [here](https://github.com/lutzhelm/mirador/tree/toc-attempt) regarding this PR. 

What works:
* Display all ranges that contain currently visible canvases, as well as their parents
* Opening the tree to any node that contains a currently visible Canvas (or part of a Canvas)
* Keeping the tree state when moving the companion window
* Scrolling to the first node that contains a visible canvas
* Keyboard navigation (up/down handled by Material-UI TreeView, left/right/enter/space handled by Mirador). This should be replaced by using `onNodeSelect` once mui-org/material-ui#18357 is merged.
* Prezi 3 did already work, although I didn't check yet whether it still does if an item within a Range is a SpecificResource instead of a Canvas

Styling and dealing with alternative structures is still open.

mejackreed:
Thanks so much @lutzhelm 👏 ! Would you mind making a pull request to this branch `toc`?

jvine:
Some comments on the behaviour:

1. The tree seems to retain a memory of the range that was selected when a parent node was first opened. I'm not sure of the rationale behind this; I'm finding that it can be quite disorienting. After navigating around, opening and selecting other ranges, each time I click Comedies I'm jumping to The Winters Tale. Is this something defined in the manifest? or? Even when I refresh the browser window, the same thing happens.

2. I think we need to be able to open/close a parent _without_ selecting it (or whatever range within it is in memory). Once I've moved down in the hierarchy (say in Histories) - if I want to close Comedies to get it out of my way... I don't actually want to jump back to it. I just want to close it. IMO the view should change only when the user expressly clicks on a child. (I imagine there's some condition in the manifest that will make this impossible...)

3. Whenever I select "Comedies - The Tempest - Actus primus, Scena prima" - I also get "[Front matter] - [Preliminaries]". I think this is because one image includes the back of the last page in this section... (but why does it select the first item in the section, rather than the last?)

But it points out another issue, which is that the tree will dramatically scroll when I click on a range that has a non-contiguous selection, and the thing I clicked on is now outside my viewport. This is disorienting, and requires the user to literally reorient themselves to continue navigating forward. IMO the viewport needs to remain where the user has clicked, even if an item that was outside of view is now selected.

mejackreed:
@jvine just following up here on the things you mentioned about the overall experience.

1. 
> Is this something defined in the manifest?

Yes, that is how the manifest is setup, so that "Comedies" goes immediately to "fol. Z6v" which is the "Winter's Tale"

3.  I think this is just really because of the edge case example of how this manifest sets up its ranges. The "you can do anything" approach here can cause usability issues for sure.

4. This is also maybe related to `#1`? Since the range creation allows them to point to this thing (not ideal) Maybe a simpler manifest doesn't display this behavior? https://iiif.bodleian.ox.ac.uk/iiif/manifest/ca3dc326-4a7b-479f-a754-5aed9d2f2cb4.json

lutzhelm:
1. Cause is indeed that "[Comedies]" contains a single Canvas for a blank page (fol. Z6v) that does not belong to any other `Range`. Selecting "Comedies" will use this `Canvas` to set the current `Canvas` as it is the first within `canvases`.

2. Opening / closing a parent without selecting is helpful in some cases, in others the user might actually want to select the parent, for example because a leaf item is only an illustration within a section, and you might want to select tha latter. Maybe we can change the default `TreeView` / `TreeItem` behavior to treat clicks on the icons different from clicks on the label (with corresponding keyboard interaction left/right arrows for opening / closing and space / enter for selecting). Or we go back to selecting leafs only.

3. It is indeed a matter with the book view. But it does not include any child nodes in "[Preliminaries]", it just looks like the first item due to styling.

4. Yes, this is irritating. The behavior of `getNodeIdToScrollTo` in `src/state/selectors/ranges.js` is currently to select the first visible leaf node that contains the Canvas, or the first branch node that contains the Canvas if no leaf node contains it, or their deepest non collapse-hidden parent. I'm not sure how to solve that. Maybe no auto-scroll-to at all or temporarily remember the node that was selected.

Some general notes:

The current selection behavior is that a node can be selected if it directly contains a Canvas. Selecting a node will set the first Canvas that is contained in the Range. It should actually check for the `start` / `startCanvas` property which would solve the problem 1. in this specific case.

For a more general solution, there are several possibilities:
  * Go back to only allow selection of leaf nodes. Problem is that branch node selection might be wanted, see 2. above.
  * Way more complex: Allow the selection of every node and detect the `Canvas` to select in the following way:
    * If present in the `Range`, use `start` / `startCanvas`.
    * If `Range` contains `items` / `members`, parse those and use the first `Canvas` found, either directly within the range or within its descendants, whatever comes first in `items` / `members`.
    * If only `ranges` is present / has any content, get the first Canvas of any child range recursively.
    * If `ranges` is present alongside `canvases`, still use the first Canvas in `canvases`.
      * Alternatively, the order in the default sequence could be used to determine which is the first `Canvas` in a branch `Range`. On the other hand, differences of order in a `Range` and in the default sequence might be intentional.

lutzhelm:
Just noticed about 3.: Nodes should only expand automatically if they contain descendants that contain currently visible canvases, not if only they themselves contain visible canvases.

mejackreed:
The discussion here has concluded (moved to other PR's now merged or on community call) and this pull request is ready for review. I have also rebased on latest `master` and added a commit to accommodate the latest manifesto changes.

PR #2920, opened 2019-12-19
Setup cached manifest selector that is shared between component insta…

### LINKED ISSUE #2917: Performance issue, when multiple windows are open, selectors are not memoized
Demo of this problem recreated forthcoming

### Steps to recreate
1. Drop a logger into commonly used, process intensive method, https://github.com/ProjectMirador/mirador/blob/master/src/state/selectors/manifests.js#L8
1. Setup a Mirador instance to load with 1 window
1. Verify that this is only called once 
1. Setup a Mirador instance to load with 2 windows
1. Notice that this will be called more than 30 times

### Broader impacts

This has broader impacts and has affects on other compute or data traversal operations, as it will trigger cascading calls.

mejackreed:
…nces

Fixes #2917 

Open to other solutions also.

This pr reduces 3 calls to this expensive call vs previously 70 😿 

## Before
beforecached

## After
setupcachedmanifests

### DIFF
--- a/__tests__/src/selectors/manifests.test.js
+++ b/__tests__/src/selectors/manifests.test.js
@@ -80,6 +80,12 @@ describe('getManifestoInstance', () => {
     const received = getManifestoInstance(state, { manifestId: 'x' });
     expect(received.id).toEqual('http://iiif.io/api/presentation/2.1/example/fixtures/19/manifest.json');
   });
+  it('is cached based off of input props', () => {
+    const state = { manifests: { x: { json: manifestFixture019 } } };
+    const received = getManifestoInstance(state, { manifestId: 'x' });
+    expect(getManifestoInstance(state, { manifestId: 'x' })).toBe(received);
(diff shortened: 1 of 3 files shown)


mejackreed:
I'm looking into seeing if this presents problems with multi-lingual manifests.

__tests__/src/selectors/manifests.test.js
```diff
@@ -80,6 +80,12 @@ describe('getManifestoInstance', () => {
     const received = getManifestoInstance(state, { manifestId: 'x' });
     expect(received.id).toEqual('http://iiif.io/api/presentation/2.1/example/fixtures/19/manifest.json');
   });
+  it('is cached based off of input props', () => {
```
mejackreed: I'm not sure how to better test this. This isn't a failing test to start with unfortunately, but I'm unsure how to actually test the recompution based off of different component instances calling our selector.

PR #2924, opened 2020-01-07
Enable both select add components to also be wrappable

### LINKED ISSUE #2919: WindowTopBarPluginMenu: Allow the developer to specify the icon used, as we all control what happens on click
Being able to change the icon would be nice.

Currently, the `WindowTopBarPluginMenu` returns a `Menu` when the button is clicked. Allowing the developer to control what happens when that button is clicked would allow more user experiences. An example would be to show a `CompanionWindow` on the right side of the viewer window, so that the user could view information about the canvas on both the left and right sides of the window. They could see all the regular info in the sidebar on the left, and external resources about the canvas on the right.

mejackreed:
Fixes #2919 

Aims to address the user facing concerns in https://github.com/sul-dlss/sul-embed/issues/1082 . This will allow adopters to customize the menu button in the `<WindowTopBarPluginMenu>` as well as turn on adding and wrapping of a component.

Checkout http://127.0.0.1:4444/__tests__/integration/mirador/plugins/add.html

## ✨ Wrapped & Added ✨ 
Screen Shot 2020-01-06 at 6 16 20 PM

The second commit here also supports additional design for users wanting to add new things to the WindowTopBar.

Screen Shot 2020-01-08 at 2 03 04 PM

### DIFF
--- a/src/extend/withPlugins.js
+++ b/src/extend/withPlugins.js
@@ -26,6 +26,19 @@ function _withPlugins(targetName, TargetComponent) { // eslint-disable-line no-u
       return <TargetComponent {...passDownProps} />;
     }
 
+    if (!isEmpty(plugins.wrap) && !isEmpty(plugins.add)) {
+      const WrapPluginComponent = plugins.wrap[0].component;
+      const AddPluginComponents = plugins.add.map(plugin => plugin.component);
+      return (
+        <WrapPluginComponent
+          targetProps={passDownProps}
+          {...passDownProps}
+          PluginComponents={AddPluginComponents}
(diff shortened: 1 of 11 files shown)


src/extend/withPlugins.js
```diff
@@ -26,6 +26,19 @@ function _withPlugins(targetName, TargetComponent) { // eslint-disable-line no-u
       return <TargetComponent {...passDownProps} />;
     }
 
+    if (!isEmpty(plugins.wrap) && !isEmpty(plugins.add)) {
```
mejackreed: @charbugs any reason not to do this?

Issue #2931, opened 2020-01-23
3 issues with integrating mirador in redux app


charbugs:
Here are 3 issues with integrating mirador in a react/redux application.

1) [x] Mirador action types should be prefixed to avoid name conflicts with actions from the host app.

2) Having two redux stores in an app is problematic and unhandy. So you want to hold both mirador's reducers and the host's app redcuer in a single store side by side. This is possible in general but the problem is that you can not organize the reducers as you want to as Mirador expects it's state slices to be at the root. And this in turn may result in reducer name conflicts.

It would be disirable to shape your state like that:
```
{
  todos: { ... }
  mirador: { ... } 
}
```
That would require to let Mirador know where to find it's state.

3) [x] During initialisation the plugin algorithm changes the entire reducer setup to inject the plugin reducers. If you set up the store like above all host reducers will be thrown away.

charbugs:
to be clear: none of the points above stops you from integrating mirador in a react/redux app. but it can be cumbersome to work around this problems. so this issue is meant to document some dx problems.

PR #2932, opened 2020-01-23
prefix action types

charbugs:
see #2931


mejackreed:
I'm good with this. Is there anything else that needs to be done to account for point 3 in #2931 ?

Is there anything within [createSlice](https://redux-starter-kit.js.org/api/createslice/) that we could use? I haven't looked into that too much, but curious if you had.

mejackreed:
One more minor thought.. is there anyway that this should be configurable as a setting? Would someone have multiple Mirador stores in the same application?

charbugs:
> Is there anything else that needs to be done to account for point 3 in #2931

It's possible to work around this by providing a `createRootReducer` function similar to the one of Mirador, so it's not a blocker for integration with React/Redux. On the other hand it's not nice to do so, it could be more developer friendly. I think this is a good resource for this issue: https://redux.js.org/recipes/code-splitting

> One more minor thought.. is there anyway that this should be configurable as a setting? Would someone have multiple Mirador stores in the same application?

Good point. I think this shouldn't be so hard. I will check what's necessary for that

charbugs:
Ok, it seems to be not that simple: 

If you run two instances of Mirador each with its own redux store than the action types will not conflict., so that's save I guess. Running two instances of Mirador with a shared store is not possible at the moment for at least two reasons:

1) The action types will conflict because of the same name (that's what was statet above)
2) You need different state slices for each Mirador instance, but Mirador expects its state at the root level (see point 3 of #2931)

Enabeling two Mirador instances in a React/Redux app requires some more work than this PR can do. So, this PR only solves the problem of name conflicts with the action types of the host application.

PR #2933, opened 2020-01-23
WIP: Use a getter function to access Mirador state

charbugs:
This shows a possible approach to solve the second point in #2931 (Mirador should not know about the global state, but only the possition of it's own state slices)

The idea is that: 

* There is a `getMiradorState(globalState: any) => MiradorState` selector that defaults to return the global state. Any time you want to read from the state you first have to get the Mirador state via this function

* There is a `setMirdorStateGetter(customGetter: (globalState: any) => MiradorState) => void` function that can be used by host applications to direct Mirador to it's state. This enables the host application to place the root reducer of Mirador to any state slice it wants to.

### DIFF
--- a/src/containers/App.js
+++ b/src/containers/App.js
@@ -2,23 +2,23 @@ import { compose } from 'redux';
 import { connect } from 'react-redux';
 import { withPlugins } from '../extend/withPlugins';
 import * as actions from '../state/actions';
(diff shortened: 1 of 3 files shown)


marlo-longley:
The `miradorSlice` function fully implements what this is proposing

PR #2944, opened 2020-02-20
Additional TOC implementations

lutzhelm:
Changes / Additions:
* Set the window's current Canvas not only for leaf nodes but for any nodes that contain canvases
* Track open TOC nodes in the application state in order to be able to reopen them if companion window is moved
* Visibly mark all ranges that contain currently visible canvases, as well as their parents
* Open any branch nodes that (or whose children) contain a currently visible Canvas (or part of a Canvas), except if a node has been manually closed
* Scroll to the first node that contains a currently visible canvas
* Keyboard navigation (up/down handled by Material-UI TreeView, left/right/enter/space handled by Mirador). This should be replaced by using `onNodeSelect` once mui-org/material-ui#18357 is merged.

To be done:
* Check if Prezi 3 `Range` items of type `SpecificResource` (instead of `Canvas` with or without fragment) do work; support those if that's not the case
* Styling
* Tests

I temporarily stored the Range ids of open nodes instead of the node ids themselves but encountered manifests where a Range would appear on different branches of the tree, so I dismissed that idea. Any thoughts on that?


src/state/actions/companionWindow.js
```diff
@@ -57,3 +57,18 @@ export function removeCompanionWindow(windowId, id) {
     windowId,
   };
 }
+
+/** */
+export function toggleNode(windowId, id, nodeId) {
```
mejackreed: Can we add a test for this?

src/state/reducers/companionWindows.js
```diff
@@ -25,6 +25,14 @@ export function companionWindowsReducer(state = {}, action) {
       return removeIn(state, [action.id]);
     case ActionTypes.IMPORT_MIRADOR_STATE:
       return action.state.companionWindows;
+    case ActionTypes.TOGGLE_TOC_NODE:
```
mejackreed: Is it possible to add a test for this?

lutzhelm: I vastly simplified the reducer and added a test.

src/state/actions/companionWindow.js
lutzhelm: I hope to get it done by tomorrow.

lutzhelm:
> I noticed one other odd thing happening, have you seen this yet? It seems like with a long list and then switching to compact view that list is not scrollable. Other than that this is looking really good!

I have noticed, but you did already fix this last year in #2921. The fix is just not in the branch.

mejackreed:
Ah maybe I did. Thanks for pointing that :)

PR #2948, opened 2020-03-05
Enable default toc selection based off of range existence

mejackreed:
@lutzhelm what do you think?

This selector now sets the default view as `tableOfContents` only if ranges are present.


mejackreed:
Thanks @lutzhelm , would you mind merging the pull request? I gave you access to the repo now

PR #2963, opened 2020-03-26
Enable basic code chunking by using React.Lazy

mejackreed:
This PR adds React.lazy and Suspence for loading the `App` component. This has the impact of enabling basic code splitting (using Webpacks SplitChunksPlugin). The result here is a quicker First Contentful Paint and First Meaningful Paint.

## Before
All JavaScript is loaded up in a single umd js file ~1.74MB. The browser does not start doing anything until this is fully loaded and parsed.
Screen Shot 2020-03-26 at 2 11 08 PM

Screen Shot 2020-03-26 at 2 14 48 PM

## After
The JavaScript is now split into 3 different chunks, equaling approximately the same size (but a bit smaller 1.72MB). The big difference here is that the browser can start loading Mirador after the first chunk is downloaded and parsed. This reduces our paint times.
Screen Shot 2020-03-26 at 2 10 01 PM
Screen Shot 2020-03-26 at 2 19 23 PM

## Browser compatibility
This does seem to work fine with Chrome, Firefox, and even IE 11.
Screen Shot 2020-03-26 at 2 23 46 PM

## Future implications
There maybe future useful implications of this functionality. Where plugins that may show other types of content might also be able to chunk out JS dependencies and only show them when needed. There are also other places where things could be chunked.

## Other Considerations
Should we modify the chunk naming scheme?
Does this have any impact who may want to use the `mirador.min.js` from unpkg.com?

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -1,13 +1,14 @@
-import React, { Component } from 'react';
+import React, { Component, lazy, Suspense } from 'react';
 import PropTypes from 'prop-types';
 import { MuiThemeProvider, createMuiTheme } from '@material-ui/core/styles';
 import Fullscreen from 'react-full-screen';
 import { I18nextProvider } from 'react-i18next';
 import { LiveAnnouncer } from 'react-aria-live';
 import createI18nInstance from '../i18n';
-import WorkspaceArea from '../containers/WorkspaceArea';
 import AuthenticationSender from '../containers/AuthenticationSender';
 import AccessTokenSender from '../containers/AccessTokenSender';
+
+const WorkspaceArea = lazy(() => import('../containers/WorkspaceArea'));
 
 /**
  * This is the top level Mirador component.
@@ -64,7 +65,11 @@ export class App extends Component {
             <MuiThemeProvider theme={createMuiTheme(theme)}>
               <AuthenticationSender />
               <AccessTokenSender />
-              <WorkspaceArea />
+              <Suspense
+                fallback={<div />}
+              >
+                <WorkspaceArea />
+              </Suspense>
             </MuiThemeProvider>
           </LiveAnnouncer>
         </I18nextProvider>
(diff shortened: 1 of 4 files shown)


src/components/App.js
```diff
@@ -64,7 +65,11 @@ export class App extends Component {
             <MuiThemeProvider theme={createMuiTheme(theme)}>
               <AuthenticationSender />
               <AccessTokenSender />
-              <WorkspaceArea />
+              <Suspense
+                fallback={<div />}
```
mejackreed: Right now we just show an empty div when loading. But we could put a spinner or other type of stuff here.

src/containers/WorkspaceMosaic.js
```diff
@@ -37,7 +37,7 @@ const styles = {
       boxShadow: 'none',
     },
     '& .mosaic-window-toolbar': {
-      display: 'none',
+      display: 'none !important',
```
mejackreed: Since the css potentially loads differently now we need to specify important.

charbugs:
I did some experiments and I think the problem is how weback locates the chunks files on runtime. From what I know so far this is done by specifying the `publicPath` (defaults to`/`) in the webpack config. But that means that this path is fixed after compilation. So Mirador would stipulate where consumers have to store the bundles.

This is probably not a problem if you use the es modules because in this case you control the compilation process. But I think it would not work with CDNs and may cause problems is you use the prebuild bundles via NPM.

I did not dig deep in this issue so there may exist solutions.

mejackreed:
Good point @charbugs . I've added 0949ce49e1c854ad384ab82099c8bde9b5b85a6f which hopefully helps in this regard. For the built Mirador distributed umd file, chunking is turned off. However, for development it is turned on so we can test how this works. If you think this will work for your uses feel free to approve and merge this PR.

Thanks!

charbugs:
I tested the integration in another react app using the es modules and it seems to work as usual. But the code splitting does not work as described above. This is probably because we use a slightly different approach for code splitting. 

> However, for development it is turned on so we can test how this works.

It does not break anything, so I'm going to merge this.

(By the way, what tool gave you this performance figures above?)

PR #2969, opened 2020-04-15
Add logout button from an auth service

mejackreed:
This PR adds a logout button that can be used to logout of a IIIF Auth service.

https://iiif.io/api/auth/1.0/#logout-service

Screen Shot 2020-04-15 at 1 43 16 PM

Kapture 2020-04-15 at 13 43 36


src/state/reducers/infoResponses.js
```diff
@@ -42,6 +44,13 @@ export const infoResponsesReducer = (state = {}, action) => {
       }, {});
     case ActionTypes.IMPORT_MIRADOR_STATE:
       return {};
+    case ActionTypes.RESET_AUTHENTICATION_STATE:
```
mejackreed, line 47: The info response for the given tokenService is now removed. A user canvas navigation will retrigger this to be requested. We could do other things, but maybe this is ok?

Issue #2946, opened 2020-02-24
Implement basic filtering annotations (by motivation)


mejackreed:
Seen in mocks https://xd.adobe.com/view/f2440f29-045b-4d8d-7b61-0d5c8ece8104-8475/screen/f5e9cc3a-11f7-4269-9291-98d48d54cbd8/Annotation-example-2-annotations-page-3r-motivatio

![Screen Shot 2020-02-24 at 4 04 16 PM](https://user-images.githubusercontent.com/1656824/75199079-65199700-571f-11ea-932d-0b90a26bf6ee.png)

GroovinChip:
Use cases:
- Annotations are being retrieved from multiple sources. We need to be able to choose which annotation source can be viewed and distinguish them
- There are different categories of data (transcription, translation, geo-spatial, etc) that annotations can contain. We need to be able to label the annotations based on the category and be able to filter annotations by category

mejackreed:
@GroovinChip For your different categories, are you using motivations to distinguish them? If so, what motivations are you using? If not, how are you distinguishing them?

GroovinChip:
@mejackreed we are currently using the annotation labels to distinguish data categories, but we could potentially use something else.

GroovinChip:
@mejackreed see https://rosetest.library.jhu.edu/rosademo/wa/rose/SeldenSupra57/1r/canvas

ggeisler:
Below are updated mockups for adding annotation filtering. 

### Filter icon and expand/collapse control
<img width="477" alt="Screen Shot 2020-05-12 at 10 31 34 AM" src="https://user-images.githubusercontent.com/101482/81726312-07ead300-943c-11ea-8063-6adb88c390a1.png">

Filter icon: `filter_list` at https://material.io/resources/icons/?search=filter&icon=filter_list&style=sharp

Expand/collapse control: `expand_more` at https://material.io/resources/icons/?search=expand&icon=expand_more&style=sharp
- Rotated to point right when closed

### Filter section expanded
When the user selects the filter's expand/collapse control, the annotation header section should expand to reveal the filter options:

<img width="477" alt="Screen Shot 2020-05-12 at 10 39 01 AM" src="https://user-images.githubusercontent.com/101482/81727011-28fff380-943d-11ea-89e5-6847bb3ff1c1.png">

[...]

PR #2972, opened 2020-04-16
Configurable annotation motivation selection fixes #2896

### LINKED ISSUE #2896: Make annotation motivations configurable
There are a few points where annotations are selected by motivation, following the pattern `...motivations: ['oa:commenting', 'sc:painting']...`.   It would be helpful if the motivations here were drawn from the configuration file, so that additional motivations, such as `supplementing`, could be included.

mejackreed:
The data structure in the settings is aimed to hopefully make https://github.com/ProjectMirador/mirador/issues/2946 easier?

### DIFF
--- a/src/lib/MiradorViewer.js
+++ b/src/lib/MiradorViewer.js
@@ -42,7 +42,9 @@ class MiradorViewer {
    * Process config into actions
    */
   processConfig() {
-    const mergedConfig = deepmerge(settings, this.config);
+    /** merge type for arrays */
+    const overwriteMerge = (destinationArray, sourceArray, options) => sourceArray;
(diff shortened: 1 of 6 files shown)


src/lib/MiradorViewer.js
```diff
@@ -42,7 +42,9 @@ class MiradorViewer {
    * Process config into actions
    */
   processConfig() {
-    const mergedConfig = deepmerge(settings, this.config);
+    /** merge type for arrays */
+    const overwriteMerge = (destinationArray, sourceArray, options) => sourceArray;
```
mejackreed, line 46: This may be the only "controversial" thing?

cbeer: From our current settings, it looks like this would only  apply to motivations (where it makes sense) and windows (an empty array, so it doesn't matter).

PR #2982, opened 2020-04-21
RTL fixes for dir="rtl" viewing of Mirador.

### LINKED ISSUE #2843: Overview ticket of technical undertaking for right to left support (RTL)
This ticket aims to serve as an overview of for implementation of RTL support in Mirador 3. It is understood that there will need to be design as well.

Areas to support:
 - [x] WindowTopBar elements (Buttons, Title, etc)
 - [x] Side panel reversal (first on right -> left -> bottom)
 - [x] Metadata / text RTL
 - [x] Navigation switched

How should it be initialized? In a configuration setting? What about if [viewingDirection](https://iiif.io/api/presentation/2.0/#primary-resource-types) is something different?

mejackreed:
Fixes part of #2843 and fixes part of https://github.com/sul-dlss/dlme/issues/942
Kapture 2020-04-21 at 17 06 43

The main piece of work this does not capture is the relabeling of section headers for "book view" multi canvas metadata. I think that should be done separately or we should talk about how best to handle that.

### DIFF
--- a/src/components/CompanionArea.js
+++ b/src/components/CompanionArea.js
@@ -20,17 +20,19 @@ export class CompanionArea extends Component {
 
   /** @private */
   slideDirection() {
-    const { position } = this.props;
+    const { direction, position } = this.props;
+    const defaultPosition = direction === 'rtl' ? 'left' : 'right';
+    const oppositePosition = direction === 'rtl' ? 'right' : 'left';
 
     switch (position) {
       case 'right':
       case 'far-right':
-        return 'left';
+        return oppositePosition;
       case 'bottom':
       case 'far-bottom':
         return 'up';
       default:
-        return 'right';
+        return defaultPosition;
     }
   }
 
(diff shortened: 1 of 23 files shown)


src/components/CompanionArea.js
```diff
@@ -18,19 +18,34 @@ export class CompanionArea extends Component {
...
+    const { companionAreaOpen, direction } = this.props;
+    if (companionAreaOpen) {
+      if (direction === 'ltr') {
+        return <ArrowLeftIcon />;
+      }
+      return <ArrowRightIcon />;
+    }
+    if (direction === 'rtl') return <ArrowLeftIcon />;
+    return <ArrowRightIcon />;
+  }
```
cbeer, line 32: Is this style inflicted on us by the style checker, or could we make these parallel (at the very least not using both `ltr` and `rtl`.. but maybe checking  `direction` before openness would help?)?
(Assuming the style checker hates unnecessary `else`, maybe like:`) 
```suggestion
  collapseIcon() {
    const { companionAreaOpen, direction } = this.props;
    if (direction  === 'rtl') {
      if (companionAreaOpen) return <ArrowRightIcon />;
      return <ArrowLeftIcon />;
    }
    
    if (companionAreaOpen) return < ArrowLeftIcon />;
    return < ArrowRightIcon />;
  }
```

 Or would it be better to extract some kind of `CollapseIcon` that could encapsulate it? (Seems overkill but..)

PR #2987, opened 2020-04-22
Support Annotations in bookview for right-to-left viewing fixes #2956

### LINKED ISSUE #2956: Support multi-canvas world offset for SvgSelector annotations
A good example is canvas 6 here: https://iiif.bodleian.ox.ac.uk/iiif/manifest/748a9d50-5a3a-440e-ab9d-567dd68b6abb.json

mejackreed:
https://iiif.bodleian.ox.ac.uk/iiif/manifest/748a9d50-5a3a-440e-ab9d-567dd68b6abb.json

Screen Shot 2020-04-22 at 2 16 21 PM

### DIFF
--- a/src/lib/CanvasWorld.js
+++ b/src/lib/CanvasWorld.js
@@ -21,14 +21,32 @@ export default class CanvasWorld {
   }
 
   /**
-   * canvasToWorldCoordinates - calculates the canvas coordinates respective to
-   * the world.
+   * contentResourceToWorldCoordinates - calculates the contentResource coordinates
+   * respective to the world.
    */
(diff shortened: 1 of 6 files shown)


src/lib/CanvasWorld.js
```diff
@@ -43,6 +43,24 @@ export default class CanvasWorld {
     ];
   }
 
+  /** */
+  canvasToWorldCoordinates(canvasId) {
```
mejackreed, line 47: This method is very similar to the one above it. I think this is ok, as we may want to support fragment placement, rotation, and things like that which may mean we will likely expand the scope of either of these.

PR #3026, opened 2020-05-13
Implement support for collections

cbeer:
TODO:
- [x] Tests
- [x] Designs for the collection companion window
- [x] Does the collection companion window replace the window modal dialog?
- [x] Update i18n strings to user-friendly versions (currently using manifesto terminology for developer sanity)  

Multi volumed work test: https://wellcomelibrary.org/iiif/collection/b18031511

Kapture 2020-07-24 at 13 40 00


src/containers/CollectionDialog.js
```diff
@@ -0,0 +1,84 @@
...
+    padding: '16px',
+  },
+  dark: {
+    color: '#000000',
+  },
+  dialogContent: {
+    padding: 0,
+  },
+  light: {
+    color: '#BDBDBD',
```
jkeck, line 64: I'm totally out of the loop on MUI theming, should we be using any palette colors or theme sizes for the padding above here?

src/components/WindowSideBarCollectionPanel.js
```diff
@@ -0,0 +1,184 @@
...
+
+    return (
+      <CompanionWindow
+        title={t(this.isMultipart() ? 'multipartCollection' : 'collection')}
+        windowId={windowId}
+        id={id}
+        titleControls={(
+          <>
+            { parentCollection && (
+              <List>
```
jkeck, line 86: Should this be using a `MenuList` for keyboard arrow navigability or is the desired behavior to tab through the collection items?

mejackreed: I think we are ok here for the `List` because this is the top level thing to go up to the parentCollection.

mejackreed: @cbeer is the consideration here for a `List` also how the text should be displayed?

src/containers/CollectionDialog.js
mejackreed: I've pushed up this change

src/components/WindowSideBarCollectionPanel.js
cbeer: I think that's right -- it's visual styling only. And, for now, it's a list of a single item (presumably items that belong to multiple collections could happen in the future), so internal navigation may not matter much here... down on line 189 is a different story -- I suspect this code predated the `MenuList`-ification and probably could/should be updated?

mejackreed: Great, I've added a commit that uses the `Menu` approach for consistency.

PR #3029, opened 2020-05-14
Support facing-pages and non-paged viewing hints

cbeer:
TODO:
- [ ] collect some examples of this
- [ ] probably find some places where we've hard-coded expecting books to have 2 page every group
- [ ] figure out if CanvasGroupings#getCanvases is even used

### DIFF
--- a/src/lib/CanvasGroupings.js
+++ b/src/lib/CanvasGroupings.js
@@ -7,15 +7,19 @@ export default class CanvasGroupings {
   constructor(canvases, viewType = 'single') {
     this.canvases = canvases;
     this.viewType = viewType;
-    this._groupings = null; // eslint-disable-line no-underscore-dangle
+    this._groupings = null;
+    this._canvasGroupingMap = null;
   }
 
   /**
+   * @deprecated
    */
   getCanvases(index) {
     switch (this.viewType) {
+      case 'scroll':
+        return this.groupings()[0];
       case 'book':
-        return this.groupings()[Math.ceil(index / 2)];
+        return this.groupings()[this._canvasGroupingMap[index]];
       default:
         return this.groupings()[index];
     }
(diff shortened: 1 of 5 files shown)


cmahnke:
I would like to bring this forward, since I have a simple use case. I'ts partly digitized stuff. Actually I'm uncertain if I try it the right way since it's my assumption that this should be a quite obvious scenario:
In one of my projects I want to present old magazines, but I'm a bit lazy: so my idea is to just digitize the interesting pages, some should be displayed double sided (two leaves next to each other) and some shouldn't. 

My fist problem is how to encode it. The [specification](https://iiif.io/api/presentation/2.0/#canvas) seems to be quite clear:

> In a sequence with the viewingHint value of “paged” and presented in a book viewing user interface, the first canvas should be presented by itself – it is typically either the cover or first recto page. Thereafter, the canvases represent the sides of the leaves, and hence may be presented with two canvases displayed as an opening of the book. If there are canvases which are in the sequence but would break this ordering, then they **must** have the viewingHint property with a value of “non-paged”. Similarly if the first canvas is not a single up, it **must** be marked as “non-paged” or an empty canvas added before it.

But this don't seem to work with this patch or Universalviewer. 

BTW, maybe my understanding is wrong, but apply statements like '**must**' only to the content encoders and not to viewer implementors? Seems a bit strange to me, that there are viewers are labeled to be IIIF compliant that don't implement this part of the specification. So I guess it's a misunderstanding on my side. 

Can anyone provide a working example?

mejackreed:
@cmhanke I think one of our challenges in pushing this forward is the lake of content examples. There are several edges of the specification that we end up not supporting because: we can't find any examples of their usage, or no one has ever requested the feature. Providing content or fixture data for these areas of the spec definitely help push this type of feature development forward.

With regards to spec viewer compliance, maybe that is something to bring up within the IIIF community. We have done our best to push forward the most widely used features and functionalities of IIIF in Mirador. With more fixture content providers and use cases driven by users, hopefully we can support even more. Having fixture content available to developers is often a limiting factor here in us moving these features forward.

cmahnke:
Well, even though my employer is part of the IIIF Consortium, I'm not doing this as a part of my job. I already tried to look up examples fo myself, since I actually thought that similar use cases (like unbounded book parts, materials like maps or figures) might happen from time to time, but had no luck yet. There is a [PR](https://github.com/IIIF/cookbook-recipes/pull/198) for the [IIIF cookbook](https://github.com/IIIF/cookbook-recipes) which addresses an comparable case in the IIIF-P 3.0 specification, but that's not really useable to me yet, since I try to  stick to version 2.X to have a broader choice of viewers.

Anyhow I'll report back here, when i've found something usable.

src/lib/CanvasGroupings.js
```diff
@@ -26,27 +29,55 @@ export default class CanvasGroupings {
...
-      // Odd page
-      if (i % 2 !== 0) {
-        groupings.push([canvas]);
-      } else {
-        // Even page
-        groupings[Math.ceil(i / 2)].push(canvas);
+
+      const hint = canvas && (
+        (canvas.getBehavior && canvas.getBehavior())
+        || (canvas.getViewingHint && canvas.getViewingHint())
```
cmahnke, line 57: If I understand the results of my debugger correctly, `canvas` is a  [Canvas Object from the `manifesto.js` package](https://github.com/IIIF-Commons/manifesto/blob/master/src/Canvas.ts). This doesn't have a `getViewingHint()` method (at least not in a ES Module build of version 4.2.0). But this finding answers the last question from your Checklist, `CanvasGroupings.js` get's called.

cmahnke:
Good News Everyone!
After [patching `manifesto.js` locally](https://github.com/cmahnke/blaufusstoelpel/blob/main/patches/manifesto.js%2B4.2.0.patch) this patch works. Here is an example:
* https://xn--blaufusstlpel-qmb.de/post/pramo-7-74/ (second image)
* https://xn--blaufusstlpel-qmb.de/post/pramo-11-73/ (second and third image)
Manifest:
* https://xn--blaufusstlpel-qmb.de/post/pramo-7-74/index.json
* https://xn--blaufusstlpel-qmb.de/post/pramo-11-73/index.json

The next issue on my agenda is to test how alternating between singe page views and double facing pages behaves and if there are some changes needed to make this work.

Here a illustration of what I'm trying to archive:
Source magazine (or book in other cases)
```
+---+ +---+---+ +---+---+ +---+---+ +---+---+ +---+
|   | |   |   | |   |   | |   |   | |   |   | |   |
| F | | 1 | 2 | | 3 | 4 | | 5 | 6 | | 7 | 8 | | B |
|   | |   |   | |   |   | |   |   | |   |   | |   |
+---+ +---+---+ +---+---+ +---+---+ +---+---+ +---+

F = Front
B = Back
```
Planed view:
```
+---+ +---+ +---+---+ +---+ +---+---+ 
|   | |   | |   |   | |   | |   |   | 
| F | | 1 | | 3 | 4 | | 6 | | 7 | 8 | 
|   | |   | |   |   | |   | |   |   | 
+---+ +---+ +---+---+ +---+ +---+---+
```
Description: Some pages aren't digitized and shouldn't  be part of the presentation view, the digitized pages should be displayed next to each other (like page 3,4 and 7,8 in the example) if they would be in the opened magazine and display as single pages if the opposite page isn't digitized (like page 1 and 6 in the example).

cmahnke:
The needed change to manifesto has been [merged and release](https://github.com/IIIF-Commons/manifesto/releases/tag/v4.2.1). Maybe we can now make some progress on this one.

cmahnke:
@cbeer, @mejackreed : Can you give an estimation when this be merged? Currently there seem to be changes to `src/lib/CanvasGroupings.js` that should be include here as well.

mjgiarlo (dismissed):

glenrobson:
As discussed we have recipe for non-paged going through the TRC approval process. Issue is here:

https://github.com/IIIF/trc/issues/88

and preview link here:

https://preview.iiif.io/cookbook/3333-choice/recipe/0035-foldouts/

Note this wouldn't meet the Princeton need in (https://github.com/pulibrary/figgy/issues/4541) as the `non-paged` images aren't hidden but just appear as a single image so as not to break the preceding or following page sequences.

tpendragon:
> `non-paged` images aren't hidden but just appear as a single image so as not to break the preceding or following page sequences.

What do you mean they appear as a single image? In the thumbnail pane?

cmahnke:
By the way, Universal Viewer supports this feature nowadays...

marlo-longley:
Todo:

- Rebase against current main
- Test with reference to https://iiif.io/api/cookbook/recipe/0035-foldouts/
- Promote out of draft

Chris's original concerns should be checked out:
-  probably find some places where we've hard-coded expecting books to have 2 page every group
-  figure out if CanvasGroupings#getCanvases is even used

PR #3030, opened 2020-05-14
Support manifest-level behaviors when selecting a default window view type

### LINKED ISSUE #3028: Add support for IIIF v3 `behavior` element

cbeer:
Fixes #3028


mejackreed:
Are there any associated manifests that you could find that have this?

cbeer:
😆 😢 

Just http://preview.iiif.io/cookbook/0010-book-2-viewing-direction/recipe/0010-book-2-viewing-direction/manifest-ttb.json (with the TTB branch).. although the images don't load and are huge, so it's not a great example.

Issue #2904, opened 2019-12-03
Manifests do now show thumbnails


GroovinChip:
Example: https://rosetest.library.jhu.edu/rosademo/iiif/rose/SeldenSupra57/manifest

In the list of manifests, it will be blank where the thumbnail should be

mejackreed:
I think Mirador expects that the `thumbnail` `@id` is a resolvable image. In this example, `https://image.library.jhu.edu/iiif/rose%2fSeldenSupra57%2fcropped%2fSeldenSupra57.001r` does not return an image.

I'm not 100% sure if this is spec compliant but all of the examples from https://iiif.io/api/presentation/2.1/ seem to show resolvable image urls as the `@id`.

PR #3043, opened 2020-05-18
Use IIIF thumbnails when provided

### LINKED ISSUE #3038: M3 does not display static thumbnails for some Manifests
M3 seems to have a problem displaying canvas' thumbnails without an image service. I've already faced this issue while working on a recipe for the IIIF cookbook, but I've fixed the Manifest in the meantime by adding a thumbnail's service... 
There are other fixtures Manifests from the cookbook that have the same issue (see [this one](https://github.com/IIIF/cookbook-recipes/blob/0010-book-2-viewing-direction/recipe/0010-book-2-viewing-direction/manifest-rtl.json) by @triplingual), but unfortunately the IIIF service on iiif.io is currently broken...

Otherwise I have this sample Manifest to test with (https://regisrob.fr/iiif/manifest/Dioscoride_Codex_vindobonensis). The thumbnails are not displayed at all in M3 (in Gallery view nor in the Thumbnail list)

```
"thumbnail":
  {
     "@id": "https://species-id.net/o/media/thumb/8/81/62-Dioscoride_A.jpg/100px-62-Dioscoride_A.jpg",
    "@type": "dctypes:Image",
    "format": "image/jpeg",
    "height": 119,
    "width": 100
}
```

But this one works fine (https://gallica.bnf.fr/iiif/ark:/12148/btv1b10500687r/manifest.json), thumbnails are displayed as expected:
```
"thumbnail": {
  "@id": "https://gallica.bnf.fr/ark:/12148/btv1b10500687r/f1.thumbnail"
}
```

cbeer:
Extract a `ThumbnailFactory` that looks up thumbnails  for all types of IIIF manifest resources, and looks at the `thumbnail` property when available.

TODO:
- [x] fix up tests
- [x] kill off captioned thumbnails (and instead offer a variant style)
- [x] rename CanvasThumbnail to IIIFThumbnail?
- [x] aspect ratio fun for the NGA fixture
- [x] cut down on some near duplicate rendering (and excessive calls to ThumbnailFactory.)

Future improvements:
- [ ] do something with level 0 images that offer sizes?
- [ ] clean up some of the duplicative conditionals
- [ ] make the minimum thumbnail size configurable
- [ ] check if there's some additional memoization that needs to happen, or otherwise minimize calls into the thumbnail logic
- [ ] use CanvasThumbnail for logos
- [ ] fix manifesto's logic for image api 0 + 1 (or at least file a bug report)


mejackreed:
Also fixes https://github.com/ProjectMirador/mirador/issues/2904 ?

src/components/ThumbnailCanvasGrouping.js
```diff
@@ -72,10 +72,12 @@ export class ThumbnailCanvasGrouping extends PureComponent {
...
         >
           {currentGroupings.map((canvas, i) => (
-            <CaptionedCanvasThumbnail
+            <IIIFThumbnail
               key={canvas.id}
-              canvas={canvas}
-              height={(position === 'far-right') ? style.height - (1.5 * SPACING) : height - (1.5 * SPACING)}
+              resource={canvas}
+              labelled
+              maxHeight={(position === 'far-right') ? style.height - (1.5 * SPACING) : height - (1.5 * SPACING)}
```
mejackreed, line 79: I'm wondering if the change from height -> maxHeight has anything to do with this?

Kapture 2020-05-19 at 15 25 53

cbeer: Fixed; labels weren't being clamped in quite the right place for the right nav.

PR #3047, opened 2020-05-20
Pass through the annotation as a prop so that a plugin can use this in a much easier way

mejackreed:
The intent here is to pass through the annotation object so that a plugin customizing the display has an easier time.

### DIFF
--- a/src/components/CanvasAnnotations.js
+++ b/src/components/CanvasAnnotations.js
@@ -75,6 +75,7 @@ export class CanvasAnnotations extends Component {
                 component={listContainerComponent}
                 className={classes.annotationListItem}
                 key={annotation.id}
+                annotation={annotation}
                 annotationid={annotation.id}
                 selected={selectedAnnotationIds.includes(annotation.id)}
                 onClick={e => this.handleClick(e, annotation)}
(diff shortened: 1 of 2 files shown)


src/components/CanvasAnnotations.js
```diff
@@ -75,6 +75,7 @@ export class CanvasAnnotations extends Component {
                 component={listContainerComponent}
                 className={classes.annotationListItem}
                 key={annotation.id}
+                annotation={annotation}
```
cbeer, line 78: Could you provide an example of how a plugin will pick up this data?

mejackreed: Sure. For a component that wraps `CanvasAnnotations` they can enhance those props by specifying a `listContainerComponent`. https://github.com/ProjectMirador/mirador-annotations/blob/master/src/plugins/canvasAnnotationsPlugin.js#L18

That `listContainerComponent` would then be able to take advantage of this. Maybe this is all just too much indirection?

cbeer: Got it; i think it's fine, just some magic in `ListItem` that's new to me.

PR #3055, opened 2020-05-22
annotations: Make search annotations prettier

jbaiter:
This switches the visualization of search annotations from the standard outline-only display to a translucent filled rectangle that is more in line with how highlighted text is usually rendered on images in digital library search engines.

For this, we introduce a new `isSearch` option to the `CanvasAnnotationDisplay` class that enables this way of rendering.

Comparison of old (i.e. general annotation rendering) visualization and new:

searchhl_old
searchhl_new

Currently the alpha-value of the annotation rectangles is hardcoded to `0.3` with a constant, but if you can give me some pointers on how to use the configuration API to make this user-configurable, I can switch that out.

I also did not include unit tests, since I couldn't find any existing tests with assertions on the actual canvas pixels. If tests are desired, for this I can try to find a way,  but it'll probably be significantly more code than the actual changeset.

### DIFF
--- a/src/lib/CanvasAnnotationDisplay.js
+++ b/src/lib/CanvasAnnotationDisplay.js
@@ -1,3 +1,6 @@
+/** Opacity of the search annotations */
+const SEARCH_ANNOTATION_ALPHA = 0.3;
+
 /**
  * CanvasAnnotationDisplay - class used to display a SVG and fragment based
  * annotations.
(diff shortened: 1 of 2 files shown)


src/lib/CanvasAnnotationDisplay.js
```diff
@@ -75,6 +79,10 @@ export default class CanvasAnnotationDisplay {
           this.context.globalAlpha = element.attributes['fill-opacity'].nodeValue;
         }
         this.context.fill(p);
+      } else if (this.isSearch) {
```
mejackreed, line 82: In this case would we have a search annotation that is being represented by an SVG? I guess this is possible?

PR #3060, opened 2020-05-25
Show the search panel if there's a default search query

cbeer:
This is a feature in sul-embed, but  I'm not sure if  we actively chose not to port it upstream for some reason?


mejackreed:
I can't remember the reasoning here either

cbeer:
👍 I think that it was just hard or annoying to make sure the search gets initialized under all scenarios. I think #3063 will provide that.

jbaiter:
FWIW, we've implemented this with this saga for our new website, maybe this can be useful for people with this use case until it gets implemented in M3 itself:

```
/** Trigger a search when we have a query in the request parameters. */
function* triggerSearchFromParams(searchParams, { manifestJson }) {
  if (!searchParams.q) {
    return
  }
  const maximizedWindowIds = yield select(getMaximizedWindowsIds)
  const windowId = maximizedWindowIds?.[0]
  if (!windowId) {
    return
  }
  const companionWindows = yield select(getCompanionWindowsForPosition, {
    windowId,
    position: 'left',
  })
  if (Object.keys(companionWindows).length === 0) {
    return
  }
  const searchService = manifestJson.service
  if (
    !searchService ||
    searchService.profile !== 'http://iiif.io/api/search/1/search'
  ) {
    return
  }
  // Trigger a search for the query in the request parameters
  yield put(
    requestSearch(
      windowId,
      companionWindows[0].id,
      `${searchService['@id']}?q=${searchParams.q}`,
      searchParams.q
    )
  )
  const { searchJson } = yield take(ActionTypes.RECEIVE_SEARCH)
  // Focus the search panel in the sidebar
  yield put(
    updateCompanionWindow(windowId, companionWindows[0].id, {
      content: 'search',
      position: 'left',
    })
  )
  if (searchParams.page) {
    // Select the first hit on the page passed in via the request parameters
    yield take(ActionTypes.SELECT_ANNOTATION)
    const targetAnno = searchJson.resources.filter(
      anno => anno.on.indexOf(`/canvas/${searchParams.page}#xywh`) >= 0
    )[0]
    if (targetAnno === undefined) {
      return
    }
    yield put(
      setContentSearchCurrentAnnotation(windowId, companionWindows[0].id, [
        targetAnno['@id'],
      ])
    )
    yield put(selectAnnotation(windowId, targetAnno['@id']))
  }
}
```

camillevilla (approved):

PR #3061, opened 2020-05-25
Refactor reducer plugins out of the PluginContext and into MiradorViewer

cbeer:
I was surprised to see we create the root reducer twice (once on initialization, once in the plugins).   

This also opens up the possibility of reducer-only plugins

One thing this drops is the ability to inject an existing plugin-less store.. we can easily re-add that, although it seems like the caller should just as easily add them in the first place.


cbeer:
Merged with #3063

PR #3064, opened 2020-05-28
Slightly darken/lighten the workspace background

cbeer:
I noticed the design mocks used a slightly darker background here. I'm not sure if using the current color was an intentional thing, or just happened to happen.

Screen Shot 2020-05-27 at 18 53 14

### DIFF
--- a/src/containers/WorkspaceArea.js
+++ b/src/containers/WorkspaceArea.js
@@ -1,7 +1,7 @@
 import { compose } from 'redux';
 import { connect } from 'react-redux';
 import { withTranslation } from 'react-i18next';
-import { withStyles } from '@material-ui/core/styles';
+import { withStyles, lighten, darken } from '@material-ui/core/styles';
 import { withPlugins } from '../extend/withPlugins';
 import { WorkspaceArea } from '../components/WorkspaceArea';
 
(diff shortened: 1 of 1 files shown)


ggeisler:
👍 I think we were trying to stick to a strict color palette so the workspace background color ended up as one of the defined grays we are using elsewhere in the app (e.g., `palette.light`), but your change does work much better and given it uses a lighten function rather than defining a new gray in the palette, this seems like a nice update to me.

camillevilla:
This looks fine to me. If a future themer wants more fine-grained control over the workspace background color they're welcome to file an issue.

I think some of the palette angst @ggeisler remembers is when we were trying to use gray for MUI's primary color. That ended up being a headache so we went with blue for `primary`. Now `theme.palette.shades.light` actually isn't used in that many places. I think it's just workspace backgrounds and the companion window header area. 

Here's an example. Left side uses the `lighten` function.  The right side just changes `theme.palette.shades.light`. 
Screen Shot 2020-05-28 at 12 43 32 PM

ggeisler:
Your memory of history sounds correct @camillevilla.

I think you merged the `lighten` function version, but just for future reference, the problem with changing `theme.palette.shades.light` to get a darker workspace background is that it affects at least one other spot: the sidebar header background. I don't think we want to change that, so the `lighten` function approach works better. It's more clear how it improves things when you change the workspace type to Elastic.

PR #3066, opened 2020-05-29
Use webpack-dev-server and HMR for development

cbeer:


src/lib/MiradorViewer.js
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import { Provider } from 'react-redux';
 import { v4 as uuid } from 'uuid';
-import { App } from '../components/App';
+import HotApp from '../components/App';
```
mejackreed, line 5: I guess exporting a HotApp for a production build has a minimal footprint so it shouldn't be an issue?

https://github.com/gaearon/react-hot-loader#what-about-production

mejackreed:
Also can we remove `concurrently` from the devdependencies?

src/lib/MiradorViewer.js
cbeer: That's the claim 🤷‍♂️

cbeer:
> One question, should we update the `build` command to provide `NODE_ENV=production`? It seems to be suggested here: https://github.com/gaearon/react-hot-loader#what-about-production

I guess it would have been good if I read that part (at least after I figured out what all these pieces were..). I ended up passing in the webpack mode to babel, but if there's some reason we want separate control??

mejackreed:
Yeah I'm not sure exactly, but I am running into issues running both `npm run build` and `NODE_ENV=production npm run build` I would have assumed that would have worked. I think this will cause our package publication to fail.

cbeer:
What about with `2b3156c`? Or, can you say more  about the issues?

mejackreed:
Seems like all is working well now. Thanks for making those changes!

PR #3071, opened 2020-06-02
WIP: Only replace root reducers if necessary

lutzhelm:
Currently, the [`App`](https://github.com/ProjectMirador/mirador/blob/35d8459b068f8c6e1b119356fbb244662e062ac5/src/components/App.js) component can't be used by react-redux applications that use Mirador as a React component because the plugin preprocessing will always replace the redux store's root reducers even if there are no plugins with their own reducers. By replacing the root reducers, the containing application's root reducers will not be available any more.

A solution could be to only replace the root reducers if there is at least one plugin with a reducer:
* If Mirador is used as standalone application (or at least not inside a redux application), everything will work as before, with or without any plugins.
* If Mirador is used inside a react-redux application, the containing application has to set up the store with its own reducers, the Mirador reducers and any M3 plugin reducers, if necessary.

### DIFF
--- a/src/extend/pluginPreprocessing.js
+++ b/src/extend/pluginPreprocessing.js
@@ -1,5 +1,6 @@
 import update from 'lodash/update';
 import { connect } from 'react-redux';
+import isEmpty from 'lodash/isEmpty';
 import { validatePlugin } from './pluginValidation';
 import CompanionWindowRegistry from '../lib/CompanionWindowRegistry';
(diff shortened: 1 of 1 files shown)


cbeer:
https://github.com/ProjectMirador/mirador/pull/3061 is perhaps an alternative?

cbeer:
Fixed by #3061.

PR #3015, opened 2020-05-11
Support non-default sequences (or sequence-like ranges?)

cbeer:
Blocked waiting for some more examples of this (especially in IIIF v3)

PR #3074, opened 2020-06-02
Update selectors to support multiple sequences or sequence-like ranges

cbeer:
This extracts just the backend state changes from #3015, with UI work to come later (possibly/probably post-3.0).


cbeer:
Maybe we can pair up on that? The ToC stuff confuses me.

PR #3077, opened 2020-06-03
Add a Serbian translation

DigitLib:
This add translation in Serbian Cyrilic all strings translated.

Please if it possible also to add one manifest from Serbia
{ manifestId: "https://zavicajna.digitalna.rs/iiif/iiif/api/presentation/2/4aa44ad1-0b74-4590-ab09-534a38cb7c53%252F00000001%252Fostalo01%252F00000012/manifest", provider: "Библиотека 'Милутин Бојић'"},

Thank You and greetings from Serbia!


src/i18n.js
```diff
@@ -21,6 +22,7 @@ export default () => {
     it,
     ja,
     nl,
+    sr,
```
mejackreed, line 25: Could you please move this below `pt-BR` so that the linting in the CI build will complete? Thanks!

src/config/settings.js
```diff
@@ -213,6 +213,7 @@ export default {
     'zh-CN': '中文(简体)',
     'zh-TW': '中文(繁體)',
     it: "Italiano",
+    sr: 'Serbian',
```
mejackreed, line 216: Would this be `српски` ? We are aiming to provide these strings in the native language?

src/i18n.js
DigitLib: Put it on the end is it OK?

src/config/settings.js
DigitLib: Done.

src/i18n.js
mejackreed: I think it needs to be between `pt-BR` and `zh-CN` see https://travis-ci.org/github/ProjectMirador/mirador/builds/694337134#L362

DigitLib: Great!!!! Do it now! TNX!!!

PR #3090, opened 2020-06-08
Point the canvas navigation arrows based on the viewing direction

### LINKED ISSUE #1957: vertical display mode
I originally reported the issue [here](https://github.com/ProjectMirador/mirador-design/issues/40) before realizing it might no be the right place, one of the two should probably be closed. For a rather large corpus of Asian manuscripts and a few modern editions, it would be a very noticeable improvement (over Mirador2 and UV) to have a mode where all the images are displayed vertically. Diva and the Internet Archive viewer have this feature and it makes reading very smooth, while our ([BDRC](https://www.tbrc.org/)'s) users report a rather poor experience with Mirador and UV.

An example with the Internet Archive viewer:

https://archive.org/details/bdrc-W4CZ57131/page/n7

For an example with Diva go to

http://library.bdrc.io/show/bdr:V22084_I0886

and click on "View images in Diva". That kind of display would be a tremendous help to many Asian users!

cbeer:
Fixes #1957?

Screen Shot 2020-06-08 at 07 28 23

### DIFF
--- a/__tests__/src/components/ViewerNavigation.test.js
+++ b/__tests__/src/components/ViewerNavigation.test.js
@@ -99,4 +99,42 @@ describe('ViewerNavigation', () => {
       expect(wrapper.find('div').props().dir).toBe('rtl');
     });
   });
+
+  describe('when viewingDirection is top-to-bottom', () => {
+    beforeEach(() => {
+      wrapper = createWrapper({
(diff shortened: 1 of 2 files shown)


triplingual:
👏🏻👏🏻👏🏻👏🏻👏🏻

PR #3094, opened 2020-06-08
SPIKE: Drag and drop

### LINKED ISSUE #2356: Support IIIF drag and drop icons
As a community, we've invested a lot in training users in this interaction pattern. We should use this ticket to define first implementation requirements for this behavior and define any major departures from Mirador 2's implementation of this.

### LINKED ISSUE #2629: react-dnd HTML5 backend from react-mosaic can conflict with other non-Mirador components
This is perhaps best addressed via a note in the wiki, unless an in-code fix is much easier than expected. The gist is that if Mirador 3 in Mosaic mode is embedded in another React app that uses `react-dnd` in a different component, the two can produce the dreaded `Cannot have two HTM5 backends at the same time` error. This can happen even when toggling from Mirador 3 to the other component, due to the asynchronous tear-down behavior in React 16.

This comment succinctly captures the problem, the challenges `react-dnd`'s architecture poses to an in-code fix, and the easiest workaround:
https://github.com/facebook/react/issues/11106#issuecomment-418591660
The workaround is to set a state flag in one of the components so that it doesn't render until mounted.

### LINKED ISSUE #2903: Drag and Drop captures window level events & prevents embedding in app with DnD
Thanks for all your work on Mirador 3 - a great thing!

I'm looking at embedding Mirador in a React based app. I can create it as a component following instructions here in github. Looks cool!

The issue is as soon as I instantiate a Mirador instance the drag and drop of my own components external to Mirador stops working. These are based on plain HTML5 DnD not a library.

I believe this is because Mirador (actually the react-dnd library) captures window level dragstart events (see screenshot).

Screenshot 2019-12-03 at 16 46 02

If I remove the listeners in Chrome I get my DnD back.

I tried moving my DnD functionality over to using the react-dnd library but then I ran into an issue that you can't create two overlapping HTML5Backend s in react-dnd - it only expects one per application. The DndProvider needs to wrap the whole app.

Is this a bug? I'm not sure. It is a failure of encapsulation but I'm not sure how you might fix it or I might find a work around. Any ideas welcome.

How dependent are you on the react-dnd library?

My current options are to redesign my app to not use DnD (bad) or find some way to remove the event listeners and filter whether events start in my components or Mirador (possibly crazy).

I'm pretty new to react & redux so I may be being naive!

cbeer:
Fixes #2356
Fixes  #2629
Fixes #2903

- [x] adding new manifests to the `WorkspaceAdd` by dropping a manifest 
- [x] adding new manifests to the `WorkspaceAdd` by dropping a drag+drop link
- [x] adding new windows by dropping a manifest onto the workspace
- [x] adding new windows by dropping a drop+drop link onto the workspace
- [x] adding new windows by  dropping the image(s) onto the  workspace
- [x] adding  new manifests to the catalog by  dropping the image(s) onto the  workspace

TODO:
- [ ] fix something up in react-mosaic-component so it reuses the DnDContext from mirador  instead of wanting  its own
- [x] add some tests for windows coming in with manifests
- [x] add tests for dropping on  workspaceadd and  workspace
- [x] add tests for  IIIFDropTarget
- [ ] figure out any design affordances


webpack.config.js
```diff
@@ -46,7 +46,13 @@ const baseConfig = mode => ({
       resourceRegExp: /@blueprintjs\/(core|icons)/, // ignore optional UI framework dependencies
     }),
   ],
-  resolve: { extensions: ['.js'] },
+  resolve: {
+    alias: {
+      // needs shared global state for context to work
+      'react-dnd': path.resolve(path.join(__dirname, 'node_modules', 'react-dnd')),
```
cbeer, line 52: I don't know how acceptable this is, but it seems like it works well enough for now?

mejackreed: Yeah it will be fine for this UMD build. However I think adopters who build their own package may need to do something similar

cbeer: And they'd probably want to anyway, for exactly the same reasons

Issue #2971, opened 2020-04-16
Loading Annotations


alansutherland:
I'm trying to load annotations, fetchAnnotations gets them successfully from my manifest and the sidebar displays the number of annotations. An element with the annotations list is created for each but no text is displayed and the annotations do not show over my image.

Checking your demo https://mirador-dev.netlify.com/__tests__/integration/mirador/ 

each of your annotations “on” value includes a “selector” which specifies a value for x/y coordinates and width/height:

```
        "on": {
                "@type": "oa:SpecificResource",
                "full": "https://iiif.harvardartmuseums.org/manifests/object/299843/canvas/canvas-47174892",
                "selector": {
                    "@type": "oa:FragmentSelector",
                    "value": "xywh=697,707,720,833"
                },
                "within": {
                    "@id": "https://iiif.harvardartmuseums.org/manifests/object/299843",
                    "@type": "sc:Manifest"
                }
            },
```

whereas my “on” value just includes a link:

```
"on" : [ "https://iiif.europeana.eu/image/GPNNPGBENPQFFOB7PHRRSCK4GJFXZCJ7YFLGCZCJ5QMNJOWVVRJA/presentation_images/c5927d61-022b-11e6-a696-fa163e2dd531/node-3/image/BNL/La_clef_du_cabinet_des_princes_de_l'Europe/1704/09/01/00159/full/full/0/default.jpg#xywh=301,199,644,80" 
]
```

**though this does include xywh data maybe it’s not the right data format Mirador is expecting?**

Do I also need to provide annotation text within my resource value?

alansutherland:
@mejackreed is there any chance you could help me with this?

mejackreed:
This seems like it could work. OpenAnnotations in Presentation v2 are being parsed in `AnnotationResource` here: https://github.com/ProjectMirador/mirador/blob/master/src/lib/AnnotationResource.js. I think a few different things are going on which are causing the annotations not to display properly in Mirador.

Using https://iiif.europeana.eu/presentation/9200301/BibliographicResource_3000126341271/manifest and https://iiif.europeana.eu/presentation/9200301/BibliographicResource_3000126341271/annopage/1 as an example.
1. Mirador checks the annotation's target to make sure that it is a canvas that is being displayed here: https://github.com/ProjectMirador/mirador/blob/master/src/components/OpenSeadragonViewer.js#L189 .

```json
{
[...]

PR #3096, opened 2020-06-08
Enable post processing of requests before they hit the store

mejackreed:
Related to https://github.com/ProjectMirador/mirador/issues/2971#issuecomment-631479528 by @rwd and discussions with @jbaiter. This PR enables Mirador adopters to post process actions (HTTP requests) made so that they can be manipulated before they hit the store.

An example would be throwing away annotation data you do not want to display for some reason (or enhancing it).

### DIFF
--- a/src/state/sagas/iiif.js
+++ b/src/state/sagas/iiif.js
@@ -31,14 +31,14 @@ function fetchWrapper(url, options, { success, degraded, failure }) {
 
 /** */
 function* fetchIiifResource(url, options, { success, degraded, failure }) {
-  const { preprocessors = [] } = yield select(getRequestsConfig);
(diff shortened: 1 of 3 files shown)


src/state/sagas/iiif.js
```diff
@@ -215,5 +215,6 @@ export default function* iiifSaga() {
     takeEvery(ActionTypes.REQUEST_ANNOTATION, fetchAnnotation),
     takeEvery(ActionTypes.RECEIVE_ACCESS_TOKEN, refetchInfoResponses),
     takeEvery(ActionTypes.ADD_RESOURCE, fetchResourceManifest),
+    // takeEvery(ActionTypes.RECEIVE_ANNOTATION, processReceiveAnno),
```
cbeer, line 218:
```suggestion
```

PR #3097, opened 2020-06-09
Ensure that viewer state is preserved on import during setCanvas

mejackreed:


src/state/sagas/windows.js
```diff
@@ -54,7 +54,9 @@ export function* setWindowStartingCanvas(action) {
         || miradorManifest.canvasAt(canvasIndex || 0)
         || miradorManifest.canvasAt(0);
       if (startCanvas) {
-        const thunk = yield call(setCanvas, windowId, startCanvas.id);
+        const thunk = yield call(
+          setCanvas, windowId, startCanvas.id, null, { preserveViewport: true },
```
cbeer, line 58: I'm not sure about this one; we end up here if the window properties didn't call out a specific canvas

src/state/sagas/windows.js
```diff
@@ -43,7 +43,7 @@ export function* setWindowStartingCanvas(action) {
   const windowId = action.id || action.window.id;
 
   if (canvasId) {
-    const thunk = yield call(setCanvas, windowId, canvasId);
+    const thunk = yield call(setCanvas, windowId, canvasId, null, { preserveViewport: true });
```
cbeer, line 46:
```suggestion
    const thunk = yield call(setCanvas, windowId, canvasId, null, { preserveViewport: !!action.payload });
```

I think we'll also want to reset the viewport if this was an update (because we get here if the manifest changes?)

PR #3102, opened 2020-06-11
Add support for hovering + clicking on annotations on canvases

### LINKED ISSUE #3058: Enable annotation highlighting via canvas
I think this has been generally discussed but I'm not sure has been ticketed so...

Currently, annotation highlighting can only be done via the "Highlight all" option and navigating (mouse hover and keyboard focus) annotations in the Annotation sidebar. I think we might also want to enable annotation highlighting via hovering over the annotation target on the canvas, when the "Highlight all" option turned on.

Currently, if the user has the "Highlight all" option turned on, all annotation targets are highlighted in blue (by default) on the canvas, but by design there is no corresponding highlighting of the annotations in the sidebar. This new requirement would recognize when the user has hovered over a canvas target and _highlight_ (not _select_) the corresponding annotation in the Annotation sidebar. When the user moved off the canvas target, the sidebar annotation highlight would be removed (similar to how the highlighting works as the user hovers over and then moves past an annotation in the sidebar).

The value of this is it would enable the user to quickly explore the visible canvas annotation targets and see each target's corresponding annotation text in the sidebar.

There might be some additional design needed (for example, do we need to support keyboard accessibility for this feature by making the canvas targets keyboard accessible) but for now I just want to add this ticket for potential discussion of details, objections, etc.

### LINKED ISSUE #3099: Highlight canvas annotation targets when hovered
Currently, you can use the "Highlight all" toggle in the Annotation window header to display all canvas targets in the highlighted (default blue) state. If you don't have the "Highlight all" toggle turned on, however, you can only discover where canvas targets are by highlighting annotation bodies in the sidebar.

- To make the annotation highlighting and selection interaction via sidebar and canvas more similar, we suggest that even if the "Highlight all" toggle is off, when the user moves the mouse over the canvas, any annotation targets that encompass the location of the user's mouse should be highlighted. 

The effect should be very similar to what is currently in @cbeer's draft PR at: https://deploy-preview-3065--mirador-dev.netlify.app/__tests__/integration/mirador/. What is new is that the hovered annotation targets should be highlighted even when the "Highlight all" toggle is off.

This would enable a user to discover and select an annotation from the canvas, without having to turn on the "Highlight all" toggle.

A complicating aspect of this is when does this behavior apply? It seems that if the user has never opened the Annotation sidebar for a given object, they might not have any interest in seeing annotations associated with that object, and so highlighting them when that user happens to hover over the canvas for other purposes (to move the canvas, for example) might be unwanted and intrusive. 

So this target-highlight-on-hover behavior should only be applied when the user has indicated interest in viewing annotations, by:

- Having an Annotation sidebar or companion window open

- Having the "Highlight all" toggle turned on (e.g., they might have closed all Annotation windows and navigated to a different sidebar, but previously turned the "Highlight all" toggle on)

- Having a currently selected annotation on any canvas. That is, even if the "Highlight all" toggle is off, there are no open Annotations companion windows, and the user has navigated to another sidebar category, the presence of a selected annotation is sufficient to trigger the behavior. 

The last rule should apply to the entire object (I think), so if the user selects an annotation on Canvas 1 and then navigates to Canvas 2, the canvas target highlighting on hover behavior should be in effect, even if they are not displaying the Annotation sidebar and the "Highlight all" toggle is off (i.e., we are using the fact that they have selected an annotation somewhere in the object as an indicator of their interest in seeing any available annotations for this object).

cbeer:
Fixes #3099
Fixes #3058 

This also refactors search annotations to participate in the same hover/selected states as regular annotations, and makes selected annotations single-valued instead of a toggle-able state.

TODO:

- [x] tests for new behaviors (how are we testing OSD?)

Future work:
- [ ] Select/hover all parts of a multi-annotation search hit (currently behavior just shows the first or the one you click on..)
- [ ] performance improvements or short-circuiting; ideas?


src/components/OpenSeadragonViewer.js
```diff
@@ -97,28 +111,44 @@ export class OpenSeadragonViewer extends Component {
...
     );
 
-    const selectedContentSearchAnnotationsUpdated = !OpenSeadragonViewer.annotationsMatch(
-      selectedContentSearchAnnotations, prevProps.selectedContentSearchAnnotations,
+    const hoveredAnnotationsUpdated = (
+      xor(hoveredAnnotationIds, prevProps.hoveredAnnotationIds).length > 0
     );
 
+    if (hoveredAnnotationsUpdated) {
+      if (hoveredAnnotationIds.length > 0) {
```
cbeer, line 133: Better ideas?

src/config/settings.js
```diff
@@ -57,6 +57,17 @@ export default {
         secondary: '#00BFFF',
       },
       section_divider: 'rgba(0, 0, 0, 0.25)',
+      annotations: {
```
mejackreed, line 60: This is a nice little update here which enables more customization of how annotations and searches are displayed 😎

src/components/OpenSeadragonViewer.js
mejackreed: 🤷

PR #3103, opened 2020-06-15
Extract AnnotationsOverlay component

cbeer:
This disentangles (slightly...) the OpenSeadragonViewer's canvas display from the annotations overlay by extracting the annotations rendering and interactions into an `AnnotationsOverlay` component.

While I was at it, I also added an explicit `PluginHook` for OSDViewer to inject similar behavior externally.

### DIFF
--- a/src/plugins/OSDReferences.js
+++ b/src/plugins/OSDReferences.js
@@ -1,6 +1,3 @@
-import React from 'react';
-import PropTypes from 'prop-types';
-
 export const OSDReferences = {
   /** */
   get(windowId) {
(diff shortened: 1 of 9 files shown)


src/plugins/OSDReferences.js
```diff
@@ -12,32 +9,3 @@ export const OSDReferences = {
     this.refs[windowId] = ref;
   },
 };
-
-/** */
-class OSDReferenceComponent extends React.Component {
```
mejackreed, line 17: Is it the case that other plugins will not need to mount this anymore? Or that they should just use the provided add plugin? Will that work with two plugins needing this interface?

cbeer: Correct; applications that want an OSD reference don't need to do anything special, we just always stash a ref away into  `OSDReferences` (where plugins are getting their data from).

PR #3122, opened 2020-06-16
Adds an import/export state integration test that verifies the viewer…

mejackreed:
… state persists

### DIFF
--- a/__tests__/integration/mirador/import_export.js
+++ b/__tests__/integration/mirador/import_export.js
@@ -1,0 +1,26 @@
+/* global miradorInstance */
+
+describe('Import/Export state', () => {
+  beforeAll(async () => {
+    await page.goto('http://127.0.0.1:4488/__tests__/integration/mirador/');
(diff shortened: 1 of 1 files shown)


__tests__/integration/mirador/import_export.js
```diff
@@ -0,0 +1,26 @@
...
+        x: 1000,
+        y: 1000,
+        zoom: 0.001,
+      };
+      const f = miradorInstance.actions.importMiradorState(state);
+      miradorInstance.store.dispatch(f);
+      return windows[0];
+    });
+    const mirador = await page.evaluate(() => miradorInstance.store.getState());
+    await expect(mirador.viewers[windowOne]).toEqual(
```
cbeer, line 22: Hm. I wonder if we should also dig in (somehow?) and find out if the viewer is actually at that pan/zoom?

mejackreed: Yeah.. I spun around on a few approaches to try that. Any ideas how to?

Issue #3107, opened 2020-06-15
Accessibility audit: canvas elements


cbeer:
> The individual pages of the digitised material are each integrated via a canvas element. These do not have an alternative description in text form that describes the content of the image in short form.
> An easy way to do this is to add the aria-label="digitized view" attribute to the canvas element, as well as an aria-describedby attribute that references the page title below the browse buttons supplement.
> In addition, the canvas element must be marked   as a graphicusing role="img".

jvine:
Tested, and it turns out just what they said:

- add an ID to the `<div>` with `class="mirador-osd-info"` (contains the canvas count and label)

- to the `<canvas>` element, add:
   - `role="img" `
   - `aria-label="digitized view" `
   - `aria-describedby="id-of-mirador-osd-info" `

(Sometimes the `aria-label` assigned to the containing `<section>` is read instead of the `aria-describedby` on the `<canvas>`, but I think that's partly my maladroitness with the SR. Also, possibly the `aria-label` to `aria-labelledby` change proposed in #3123 will solve this.)

cbeer:
Is there any way to get away without touching the `<canvas>`? I'm pretty sure that's coming out of openseadragon.

PR #3132, opened 2020-06-17
Rework aria attributes on the viewer

### LINKED ISSUE #3123: Accessibility audit: page turning status
> Once a digitised item is open, users can navigate through the pages/views of the digitised material using the arrow buttons "Previous item" and "Next item/Next item". While sighted users perceive the "turning of pages" directly, blind users are not informed that the view has changed.
> To change this, a live region could be created using the role="status" attribute. After paging, the text (e.g., the text below the digitized page number and name) should be placed inside the element with the role="status" attribute. This would mean that the text would be read by the screen reader after turning the page without shifting the focus.

cbeer:
Fixes #3123 and #3107

### DIFF
--- a/src/components/OpenSeadragonViewer.js
+++ b/src/components/OpenSeadragonViewer.js
@@ -33,7 +33,7 @@ export class OpenSeadragonViewer extends Component {
    * React lifecycle event
    */
   componentDidMount() {
-    const { osdConfig } = this.props;
+    const { osdConfig, t, windowId } = this.props;
     if (!this.ref.current) {
       return;
     }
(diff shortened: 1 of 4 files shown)


src/components/OpenSeadragonViewer.js
```diff
@@ -42,6 +42,14 @@ export class OpenSeadragonViewer extends Component {
       id: this.ref.current.id,
       ...osdConfig,
     });
+
+    const canvas = viewer.canvas.firstElementChild;
```
cbeer, line 46: 🤷‍♂️

src/components/OpenSeadragonViewer.js
```diff
@@ -42,6 +42,14 @@ export class OpenSeadragonViewer extends Component {
       id: this.ref.current.id,
       ...osdConfig,
     });
+
+    const canvas = viewer.canvas && viewer.canvas.firstElementChild;
```
cbeer, line 46: 🤷‍♂️

mejackreed: `sad_kitty.gif`

PR #3133, opened 2020-06-17
Add de translation for exportCopied

cbeer:
From #2565


camillevilla:
Added @lutzhelm and @a-nnika since they edited and reviewed the most recent set of German translations.

PR #3144, opened 2020-06-20
Handle sidebar buttons within the tab component

cbeer:


src/components/WindowSideBarButtons.js
```diff
@@ -125,7 +124,15 @@ export class WindowSideBarButtons extends Component {
             )}
           />
         )}
-        <PluginHook TabButton={TabButton} PluginComponents={PluginComponents} />
+        { PluginComponents
```
mejackreed, line 127: I like where this is going. Do we then have a new type of plugin here? One where a mode is not specified but can be extended into this area?

cbeer: I don't know if it's a new type as far as its consumers go. It's just like `add`, only... well, it works in  `Tabs`, which reaches into is children to pluck out the `value` prop.

mejackreed: Ok.. I've updated the wiki accordingly.

PR #3148, opened 2020-06-21
Don't expose reducers and sagas as part of the index object

cbeer:
🤷‍♂️ 

I'm not sure what the point of these exports are, exactly, but presumably it should contain everything state related or nothing at all?


cbeer:
Hm. I was thinking one reason to expose them was for apps that want to provide their own redux store (but.. how likely is it they'll do that and also want to use the UMD build?).

I wonder if all this state stuff could/should be a separate artifact? There if you want it, but a cleaner interface for everyone else?

mejackreed:
Thats a good point. Some of our integration tests use this API, but that could be changed. We previously had the "state" only library as well "core", but removed that in https://github.com/ProjectMirador/mirador/pull/2910 .

mejackreed:
From discussion on Slack. Before continuing to expose this part of the API, lets get feedback from adopters who really want this rather than maintain this level of compatibility.

PR #3167, opened 2020-06-27
Use selectors to access redux state

### LINKED ISSUE #2931: 3 issues with integrating mirador in redux app

cbeer:
Fixes #2931. With this PR, all containers use selectors to access redux state, and all selectors go through the `miradorState` function to get mirador's slice of state.

The slice is configurable in settings:

```js
import settings from 'mirador/dist/es/src/config/settings';
settings.state.slice = 'mirador';
```


src/state/selectors/utils.js
```diff
@@ -0,0 +1,4 @@
+/** */
```
cbeer, line 1: I don't know if this is a sufficient faux extension point to address #2931?  I'm thinking we can either leave it as-is as let consumers who want to move mirador do some webpacking magic, or we could make the default `state.mirador` and probably satisfy most use cases?

PR #3175, opened 2020-06-30
Only show the annotation sidebar button if there are annotations for the manifest

cbeer:


src/containers/WindowSideBarButtons.js
```diff
@@ -33,6 +33,11 @@ function hasLayers(canvases) {
   return canvases && canvases.some(c => new MiradorCanvas(c).imageResources.length > 1);
 }
 
+/** */
+function hasAnnotations(canvases) {
+  return canvases && canvases.some(c => new MiradorCanvas(c).annotationListUris.length > 0);
```
mejackreed, line 38: Will this work with embedded v3 annotations?

src/components/WindowSideBarButtons.js
```diff
@@ -94,7 +95,7 @@ export class WindowSideBarButtons extends Component {
             icon={(<CanvasIndexIcon />)}
           />
         )}
-        {panels.annotations && (
+        {panels.annotations && (hasAnnotations || hasAnyAnnotations) && (
```
mejackreed, line 98: Is this extendable to just have enabled? I ask for the mirador-annotations plugin where we rely on this companion window to start creating annotations.

cbeer: I was thinking about changing the  `window.panels` config  to true/false/maybe, but then figured plugins who cared could just wrap `WindowSideBarButtons` and provide their own behavior?

PR #3177, opened 2020-06-30
Add some indication about annotations in the gallery view

cbeer:
This shows a badge in the gallery view with the number of annotations on each canvas. Annotation lists are requested as needed. Annotations only show up if an annotation companion window is open (or is selected in the sidebar but is collapsed or hidden).

Also includes #3178

Screen Shot 2020-07-01 at 10 45 14

TODO:
- [x] design approval
- [x] tests
- [ ] rate limiting considerations?


ggeisler:
This looks good to me. I think the presentation definitely conveys that it is displaying the annotation count on each canvas.

A couple of things come to mind, although I'm not sure how important they are, so we might need @jvine to chime in to come to a final decision on the design aspect.

- Although we're showing two pieces of info here (annotation count and annotation icon) just as we do in the gallery view of the search results, we're presenting them in a different way. I don't think the two views have to be consistent, necessarily, but I guess I also wonder if there is a good reason for them to be different?

- The blue background of the annotation count makes them feel more prominent, which is arguably makes it them easier to scan than the search results counts, but in a gallery view where there are a lot of canvases with annotations, is that going to be too much?

- The chat icon used is flipped from the orientation we use in the sidebar. I don't think this is necessarily a problem, and it balances better with the annotation count in this orientation. At the same time, it made me wonder if having the same orientation as the sidebar icon would more clearly communicate what the icon is about.

All of these are very minor. @jvine when you have a few minutes maybe you could react and we can go with your thoughts on the points I made above.

jvine:
Now that I'm focusing on it, I'm not sure why we would introduce a different pattern from the search results. Grey chip, centered, same icon as sidebar:

Screen Shot 2020-07-01 at 10 26 03 AM

cbeer:
and if there's both search hits and annotation results?

jvine:
Urgh, I was assuming a match to the open panel - e.g. search panel shows search results in gallery view, anno panel shows anno results; but of course we helpfully made all panels openable at once, and annos are different from search results in that they're "permanent". Are we intending to show the presence of annos in all contexts? If a manifest has annos, can users ever see the gallery view without the anno overlay?

In principle, if we're presenting a count, I don't think we should display it in a different way from the search count:  there are X instances of this thing. And I don't think we should introduce colour if it's not meaningful. 

We could stack them when both are present. And maybe reconsider the placement so we're not obscuring the center of the image. (I like the top right.)

But I do think we should consider tying the overlay to the open panel(s).

ggeisler:
> But I do think we should consider tying the overlay to the open panel(s).

Yeah, I was assuming the overlay was dependent on the open panel. That feels right to me, even without considering the potential conflict with annotations. If you're in gallery view, and have search open, you see the hit indicators on the canvases. If you switch to another sidebar category, gallery view looks the same, just that the hit indicators disappear from the canvases (which enables you to see the entire canvas unobscured, which arguably could be what you want if you switch to another category). Then if you move back to the search category, the hit indicators appear again.

@jvine's mockup of the annotation indicator looks good to me. Does feel better to have the icon flipped like the sidebar.

cbeer:
The search chips are tied to having an existing search going (so if you do a search, they'll show up until you either click clear the search or (in a popped out companion window) close the companion window.

Annotations chips are tied to having an active annotation companion window. 

You could have both annotations and search chips if you popped one of them out to the right.

If we want to avoid covering the center of the image, I think we either need a different placement for the annotation chip, e.g. :

Screen Shot 2020-07-01 at 11 01 01

or need to push the search chip down a little more? Any preference?

ggeisler:
> or need to push the search chip down a little more? Any preference?

That looks good to me. 

I'd suggest the search chip be the same distance from the bottom of the canvas as the annotation chip is from the top of the canvas, but that would put it too close to the canvas title.  The current separation between elements seems good to me.

jvine:
Maybe pin them to top right instead of centering? with the second just below the first?

mejackreed:
Is the intent that this does not get requested until the Gallery item is selected?

Kapture 2020-07-08 at 11 22 39

cbeer:
Hm. What browser? (and that `0` is a mistake)

mejackreed:
Firefox 78.0.1 (64-bit)

PR #3178, opened 2020-06-30
Update gallery view display for search results

### LINKED ISSUE #3176: Update icon used in search result hit count chips in gallery view
When you view search results in gallery view, you currently see the MUI chip component used to indicate which canvases have result hits on them. The chip component uses the avatar icon:

Screen Shot 2020-06-30 at 1 50 34 PM

When @jvine were looking at this for another reason earlier, it occurred to us that the search icon would work better in conveying that the chip was presenting a hit count for search results.

So the request is:

- [x] Remove the background circle that is underneath the chip icon (`MuiAvatar-circle` class?)
- [x] Replace the current icon with this one:

https://material.io/resources/icons/?search=sear&icon=search&style=sharp

The result should look like this:

Screen Shot 2020-06-30 at 1 58 55 PM

cbeer:
Fixes #3176 

Screen Shot 2020-06-30 at 15 36 04

### DIFF
--- a/src/components/GalleryViewThumbnail.js
+++ b/src/components/GalleryViewThumbnail.js
@@ -2,6 +2,7 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import Avatar from '@material-ui/core/Avatar';
 import Chip from '@material-ui/core/Chip';
+import SearchIcon from '@material-ui/icons/SearchSharp';
 import classNames from 'classnames';
 import MiradorCanvas from '../lib/MiradorCanvas';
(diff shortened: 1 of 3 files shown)


src/components/GalleryViewThumbnail.js
```diff
@@ -119,7 +123,6 @@ export class GalleryViewThumbnail extends Component {
 
 GalleryViewThumbnail.propTypes = {
   annotationsCount: PropTypes.number,
-  annotationSelected: PropTypes.bool,
```
camillevilla, line 122: Context: We do not support selecting multiple annotations as of #3102. This prop is no longer needed.

PR #3196, opened 2020-07-04
Try a little harder to find a good level 0 thumbnail image

cbeer:


src/lib/ThumbnailFactory.js
```diff
@@ -80,16 +86,62 @@ class ThumbnailFactory {
...
-    if (isLevel0ImageProfile(service)) {
-      return ThumbnailFactory.staticImageUrl(resource);
-    }
-
     const aspectRatio = resource.getWidth()
       && resource.getHeight()
       && (resource.getWidth() / resource.getHeight());
 
-    if (maxHeight && maxWidth) {
+    // just bail to a static image, even though sizes might provide something better
```
mejackreed, line 93: Is this handling the case where an image information response provides a size but the image server does not respond with it?

cbeer: if the manifest is lying about the available sizes?

mejackreed: Yes

cbeer: Nope; I don't think we handle that case for any of our thumbnail calculations, before or after this PR. Do we have examples of that in the wild?

mejackreed: No I was just trying to better understand what was happening here. But I think I get it now

PR #3197, opened 2020-07-05
replace immutablejs by lodash

charbugs:
immutable is a realtive big library and we make only little use of it. any functions we use from immutable can be replaced by lodash and lodash/fp.

### DIFF
--- a/__tests__/src/selectors/ranges.test.js
+++ b/__tests__/src/selectors/ranges.test.js
@@ -1,4 +1,5 @@
-import { setIn } from 'immutable';
+import set from 'lodash/fp/set';
+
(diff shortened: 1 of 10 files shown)


charbugs:
Has someone an idea what the code climate check is trying to tell me?

mejackreed:
I would be fine with ignoring the CodeClimate check.. it seems not relevant.

charbugs:
ok, and what about the travis check? it shows a red cross above but the logs say: #8234 passed.

mejackreed:
Maybe just a strange issue with travis. Thanks for moving this improvement forward @charbugs !!

Issue #3198, opened 2020-07-06
Provide a template for using es modules in codesandbox


mejackreed:
The previous approach doesn't work any longer with the hot module loader added. 

Example: https://codesandbox.io/s/mirador-with-multiple-plugins-umyto

It would be nice to have a template for prototyping plugins or sharing examples.

mejackreed:
This is now resolved in rc4

PR #3213, opened 2020-07-13
Replace react-hot-loader with react-refresh

mejackreed:
Per https://github.com/facebook/react/issues/16604#issuecomment-528663101 and https://github.com/gaearon/react-hot-loader/issues/1453 it seems like react-hot-loader is deprecated.

This replaces react-hot-loader w/ react-refresh and https://github.com/pmmmwh/react-refresh-webpack-plugin. I'm thinking that this implementation will be preferred as it does not keep the lingering react-hot-loader around in distributed package.

Related to #3198 

This change has also been tested on IE11 for the production umd build.


webpack.config.js
```diff
@@ -9,6 +10,7 @@ const baseConfig = mode => ({
   module: {
     rules: [
       {
+        exclude: /node_modules/,
```
mejackreed, line 13: I'm not sure about the implications here. Everything seems to work ok for me.

mejackreed: Relevant: https://github.com/pmmmwh/react-refresh-webpack-plugin/issues/92#issuecomment-628960771

webpack.config.js
```diff
@@ -9,6 +10,7 @@ const baseConfig = mode => ({
   module: {
     rules: [
       {
+        exclude: /node_modules\/(react-refresh-webpack-plugin)/,
```
mejackreed, line 13: We should only exclude this plugin from `babel-loader`

mejackreed: Tested on IE11

PR #3228, opened 2020-07-22
Enable plugins to provide their own configs that can get merged into …

### LINKED ISSUE #3020: Mirador 3 plugin localization
It would probably be helpful to have a documented way to add translations for a Mirador 3 plugin. I guess currently plugins will only be able to use Mirador's react-i18next instance by updating the config via dispatching `UPDATE_CONFIG` with a `translations` object.

Whatever the solution will be in detail, a way to provide a plugin's i18n could be adding a `translations` key to the plugin object. Something like:
```javascript
{
  component: FooBar,
  mode: 'add',
  target: 'WindowTopBarPluginMenu',
  translations: {
    en: { foo: 'bar', ... }, ...
  }
}
```

mejackreed:
…the main redux config

Fixes #3020  and allows for plugins to inject their own translations see: https://github.com/ProjectMirador/mirador-image-tools/pull/21

### DIFF
--- a/src/lib/MiradorViewer.js
+++ b/src/lib/MiradorViewer.js
@@ -45,7 +45,7 @@ class MiradorViewer {
   processConfig() {
     this.store.dispatch(
       importConfig(
-        deepmerge(getConfigFromPlugins(this.plugins), this.config),
+        deepmerge(this.config, getConfigFromPlugins(this.plugins)),
       ),
     );
   }
--- a/src/lib/MiradorViewer.js
+++ b/src/lib/MiradorViewer.js
@@ -45,7 +45,7 @@ class MiradorViewer {
   processConfig() {
     this.store.dispatch(
       importConfig(
-        deepmerge(this.config, getConfigFromPlugins(this.plugins)),
+        deepmerge(getConfigFromPlugins(this.plugins), this.config),
       ),
     );
   }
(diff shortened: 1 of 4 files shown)


src/extend/withPlugins.js
```diff
@@ -48,7 +49,7 @@ function _withPlugins(targetName, TargetComponent) { // eslint-disable-line no-u
   const whatever = React.forwardRef(PluginHoc);
 
   whatever.displayName = `WithPlugins(${targetName})`;
-  return whatever;
+  return withTranslation()(whatever);
```
mejackreed, line 52: I'm not sure if this is something bad here.. getting some warnings about `t` in the props validation. Would like a second set of eyes.

cbeer: Couldn't plugins call withTranslation themselves? Or is this another weird JS wiring issue?

mejackreed: Just verified.. no they totally can. I'm going to pop off this commit.

src/lib/MiradorViewer.js
```diff
@@ -38,10 +40,14 @@ class MiradorViewer {
...
 
   /**
-   * Process config into actions
+   * Process config with plugin configs into actions
    */
   processConfig() {
-    this.store.dispatch(importConfig(this.config));
+    this.store.dispatch(
+      importConfig(
+        deepmerge(this.config, getConfigFromPlugins(this.plugins)),
```
cbeer, line 48:
```suggestion
        deepmerge(getConfigFromPlugins(this.plugins), this.config),
```

I feel like the explicit configuration should always "win" over the plugin-provided config so the application can override the plugins?

mejackreed: Totally 💯

PR #3232, opened 2020-07-24
Annotation enhancements needed to support multiple selectors and target sources

mejackreed:
Based off of discussion in https://github.com/ProjectMirador/mirador-annotations/issues/21 and https://github.com/ProjectMirador/mirador-annotations/issues/20

### DIFF
--- a/__tests__/src/lib/AnnotationItem.test.js
+++ b/__tests__/src/lib/AnnotationItem.test.js
@@ -49,6 +49,12 @@ describe('AnnotationItem', () => {
     it('returns null when there is no target', () => {
       expect(new AnnotationItem().targetId).toBeNull();
     });
+
+    it('supports a source id', () => {
+      expect(
+        new AnnotationItem({ target: { source: { id: 'foo' } } }).targetId,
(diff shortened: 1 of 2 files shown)


__tests__/src/lib/AnnotationItem.test.js
```diff
@@ -124,6 +130,10 @@ describe('AnnotationItem', () => {
       expect(new AnnotationItem({ target: 'www.example.com/#xywh=10,10,100,200' })
         .fragmentSelector).toEqual([10, 10, 100, 200]);
     });
+    it('multiple selectors', () => {
+      expect(new AnnotationItem({ target: { selector: [{ type: 'FragmentSelector', value: 'www.example.com/#xywh=10,10,100,200' }] } })
```
robcast, line 134: AFAIK the `value` of a `FragmentSelector` is just the fragment, i.e. `value: 'xywh=10,10,100,200'` ;-)

mejackreed: Thanks for that @robcast. I've updated the test to reflect that.

PR #3250, opened 2020-09-05
Collections initialization proposal #2.

mejackreed:
Pros: The window is a little bit more "pure" here not knowing it has a
manifest to start viewing reduces some complexities.
Cons: We add a new initialization/window key "iiifContent" (or something
else). And we let the sagas/state figure things out and navigate between
IIIF::Manifest and IIIF::Collection


mejackreed:
Closing in favor of #3249

PR #3254, opened 2020-09-14
Consolidate IIIF auth behaviors

cbeer:
Refactor the IIIF auth implementation to:

- consolidate IIIF auth workflows into the `IIIFAuthentication` component
- extract `WindowAuthenticationBar` to provide UI controls for said component
- add support for IIIF v0.9, and provide configuration for additional auth profiles
- do a better job eagerly requesting + cleaning up around IIIF access tokens
- get rid of more thunks, in favor of sagas.


mejackreed (approved):
Thanks for the hard work in improving the API and state work here!

PR #3257, opened 2020-09-18
Allow plugin components to be given as react elements

cbeer:

### DIFF
--- a/src/components/PluginHook.js
+++ b/src/components/PluginHook.js
@@ -6,11 +6,15 @@ export const PluginHook = React.forwardRef((props, ref) => {
   const { classes, ...otherProps } = props; // eslint-disable-line react/prop-types
   return PluginComponents ? (
     PluginComponents.map((PluginComponent, index) => ( // eslint-disable-line react/prop-types
-      <PluginComponent
-        ref={ref}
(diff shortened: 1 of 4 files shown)


src/components/PluginHook.js
```diff
@@ -6,11 +6,15 @@ export const PluginHook = React.forwardRef((props, ref) => {
   const { classes, ...otherProps } = props; // eslint-disable-line react/prop-types
   return PluginComponents ? (
     PluginComponents.map((PluginComponent, index) => ( // eslint-disable-line react/prop-types
-      <PluginComponent
-        ref={ref}
-        {...otherProps}
-        key={index} // eslint-disable-line react/no-array-index-key
-      />
+      React.isValidElement(PluginComponent)
+        ? React.cloneElement(PluginComponent, { ...otherProps, ref })
```
mejackreed, line 10: Do we need the distinguishing prop `key` here? Or will we expect that passed in? Or will it just be ok?

cbeer: I think not (and we probably don't need it below either) because it's totally artificial 🤷‍♂️

mejackreed (approved):

Issue #3258, opened 2020-09-18
Mirador 3 release process: verify mapStateToProps are being used in Mirador components


mejackreed:
Rather than continuing to pass through unused props

PR #3267, opened 2020-09-24
Fix canvas label display in ViewerInfo.

jbaiter:
The component would always get be passed an `undefined` prop for the label, since the selector in the container used the `canvasIdx` to select the label, but the selector expected the identifier.

This was missed in #2851 when the canvas identifier was passed to components that were missing it.

Additionally, this commit removes the `canvasLabel` prop from the `WindowCanvasNavigationControls` container, since the label is never used in the actual component. (#3258)

**Before:**
image

**After:**
image

### DIFF
--- a/src/containers/ViewerInfo.js
+++ b/src/containers/ViewerInfo.js
@@ -5,7 +5,7 @@ import { withStyles } from '@material-ui/core/styles';
 import { withPlugins } from '../extend/withPlugins';
 import { ViewerInfo } from '../components/ViewerInfo';
 import {
-  getCanvasLabel, getCanvases, getCanvasIndex, getVisibleCanvasIds
+  getCanvasLabel, getCanvases, getCanvasIndex, getCurrentCanvas,
 } from '../state/selectors';
 
 /**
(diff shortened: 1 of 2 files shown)


src/containers/ViewerInfo.js
```diff
@@ -15,12 +17,13 @@ const mapStateToProps = (state, props) => {
   const { windowId } = props;
   const canvases = getCanvases(state, { windowId });
   const canvasIndex = getCanvasIndex(state, { windowId });
+  const canvasId = getVisibleCanvasIds(state, { windowId })[0];
```
cbeer, line 20:
```suggestion
  const canvasId: (getCurrentCanvas(state, { windowId }) || {}).id,
```

src/containers/ViewerInfo.js
```diff
@@ -4,7 +4,9 @@ import { withTranslation } from 'react-i18next';
 import { withStyles } from '@material-ui/core/styles';
 import { withPlugins } from '../extend/withPlugins';
 import { ViewerInfo } from '../components/ViewerInfo';
-import { getCanvasLabel, getCanvases, getCanvasIndex } from '../state/selectors';
+import {
+  getCanvasLabel, getCanvases, getCanvasIndex, getVisibleCanvasIds
```
cbeer, line 8:
```suggestion
  getCanvasLabel, getCanvases, getCanvasIndex, getCurrentCanvas
```

cbeer (commented):
I think we want to use `getCurrentCanvas` here for now (even though visible canvases is effectively the same thing..). We just went back and forth about whether `getCurrentCanvas` was still useful, and it seems like there's some ways it might be significant :shrug:

cbeer (approved):

PR #3277, opened 2020-10-06
Added lithuanian translation

sakocius:
Added a Lithuanian language option.


mejackreed (approved):
Thank you so much for the translation @sakocius 👏

PR #3280, opened 2020-10-06
Add Lithuanian language to settings for language picker

mejackreed:
Is this correct @sakocius ?

image


sakocius:
> Is this correct @sakocius ?
> 
> image

Yes, that's correct, thanks for catching what I missed 👍

camillevilla (approved):
👏 Thank you for contributing, @sakocius!

PR #3286, opened 2020-10-13
Provide back an instance from initialization allowing adopters to safely unmount and remove all listeners

### LINKED ISSUE #3234: Initialize mirador after clear DOM element
Hi

The Html :
`<div id="viewer1" allowfullscreen="allowfullscreen"></div>`

The script  to init  mirador::
```
Mirador.viewer({
        id: "viewer1",
        windows: [
          {
              manifestId:
                "https://iiif.bodleian.ox.ac.uk/iiif/manifest/e32a277e-91e2-4a6d-8ba6-cc4bad230410.json"
          }
        ],
    });
```
Everything works well.
But in some case,  I need to clear  the element that the mirador exists in it, in order to display another content, instead of mirador viewer.
I do it by javascript: : 
`document.getElementById('viewer1').innerHTML = "";`
But then, when I want to redisplay the mirador by running the above code that inits the viewer, I have some javascript  errors:
TypeError: Cannot read property 'appendChild' of null
TypeError: Cannot read property 'innerTracker' of undefined
Digging into the mirador code in order to get more details about that error is difficult, as I use the minimized code
Am I doing something wrong? 

Thank you very much for your support!

### LINKED ISSUE #3285: Find some way to tear down + cleanup styles within a containing page
When a Mirador instance is initialized in a webpage, it will inject styles each time. This can cause some unintended consequences.

mejackreed:
Fixes #3234
Fixes #3285

This will also cleanup injected styles from `withStyles` API from Material-UI.

### DIFF
--- a/src/lib/MiradorViewer.js
+++ b/src/lib/MiradorViewer.js
@@ -25,18 +25,12 @@ class MiradorViewer {
       || createStore(getReducersFromPlugins(this.plugins), getSagasFromPlugins(this.plugins));
     this.processConfig();
 
-    const viewer = {
-      store: this.store,
-    };
-
     ReactDOM.render(
       <Provider store={this.store}>
         <HotApp plugins={this.plugins} />
       </Provider>,
       document.getElementById(config.id),
     );
-
-    return viewer;
   }
 
   /**
@@ -48,6 +42,13 @@ class MiradorViewer {
         deepmerge(getConfigFromPlugins(this.plugins), this.config),
       ),
     );
+  }
+
+  /**
+   * Cleanup method to unmount Mirador from the dom
+   */
+  unmount() {
+    ReactDOM.unmountComponentAtNode(document.getElementById(this.config.id));
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/lib/MiradorViewer.js
```diff
@@ -25,18 +25,12 @@ class MiradorViewer {
       || createStore(getReducersFromPlugins(this.plugins), getSagasFromPlugins(this.plugins));
     this.processConfig();
 
-    const viewer = {
-      store: this.store,
```
cbeer, line 29: I think we need to return the store too, in case it wasn't provided?

mejackreed: Is it enough that is available as a property on the instance?

via https://deploy-preview-3286--mirador-dev.netlify.app/__tests__/integration/mirador/
```
> miradorInstance
Object { config: {…}, plugins: [], store: {…} }

> miradorInstance.store
Object { dispatch: dispatch(), subscribe: g(e), getState: f(), replaceReducer: replaceReducer(n), liftedStore: {…}, Symbol(observable): ()
 }

​


```

cbeer: 👍 ; i wasn't thinking JS.

cbeer (approved):

src/lib/MiradorViewer.js
mejackreed: Yep.. I think I originally wrote this thinking in 💎

Issue #3248, opened 2020-09-05
Revamp collection modal


ggeisler:
This addresses the problem in #3235 but I'm creating a new ticket because it is also makes a more general proposal to significantly revamp the collection modal. (@robcast this includes your suggestion for list filtering)

This is an overview of the proposal. As discussed with @mejackreed I will, as necessary, add details to individual tickets for different parts of the proposal when I'm able to, but in case the team has the opportunity to get started on this in the meantime, I wanted to convey the general idea.

Related tickets:

- #3273 - Considerations for the collection modal when in the elastic workspace
- #3274 - Responsive adjustments to the modal when in the `xs` viewport/window width

## Support initializing Mirador with collection-only manifests
To address the problem noted in #3235, the proposal is to support collection-only manifests by creating a variation of our standard object window, which I'll call a _placeholder window_. If Mirador is initialized with a manifest that only specifies a collection, we load a placeholder window and display the collection modal above it, as shown below. The placeholder window omits all the usual window affordances except for the title (from the collection manifest), the fullscreen icon (in case the user needs more real estate to work with the modal), and the close window button (in case the user doesn't care about this collection).

<img width="1234" alt="Screen Shot 2020-09-29 at 11 10 06 AM" src="https://user-images.githubusercontent.com/101482/94599788-cc637080-0245-11eb-9f80-94137bdf1f0f.png">

---

Notes on aspects of the revamped modal arrangement and content:

- Modal header identifies the current level of the collection navigation
- Left sidebar should follow the pattern of the window sidebar (same about and index icons, width, etc.)
- `Collection contents` shows the count of whatever is in the manifest list; `n collections` or `n items` or both as in the mockup above. (Use "items" instead of "manifests" as we do in the current modal.)
- The filter input is for narrowing the list, similar to the feature in the Mirador 2 catalog. We might only display this if there are a minimum number of items in the list, such as 10.
- The Collections and Items checkboxes filter the list based on manifest type. They only need to be displayed if the list includes both types of manifests. Otherwise, omit both checkboxes.
- Thumbnails are 44x44 px placeholders. Whenever we can get thumbnails from the manifest (https://iiif.io/api/presentation/3.0/#51-collection) or from previously loaded manifests, they should be displayed proportionally, with 44px on the long side.

## About view
The mockup above displays the index view, which should be the default view. There is also the about view, which displays the collection metadata:

[...]

PR #3290, opened 2020-10-16
Remove workspace based collectiondialog

mejackreed:
Part of #3248

Kapture 2020-10-16 at 07 52 05

I missed this originally, but the design calls for removing the workspace based CollectionDialog. This simplifies our states understanding of how all of this should work which is nice.


src/components/CollectionDialog.js
```diff
@@ -163,6 +155,13 @@ export class CollectionDialog extends Component {
     const { filter } = this.state;
 
     if (error) return null;
+    // If this component is optimistically rendering ahead of the window its in
+    // force a re-render so that it is placed correctly. The right thing here is
+    // to maybe pass a ref.
+    if (!document.querySelector(`#${containerId} #${windowId}`)) {
+      this.forceUpdate();
```
mejackreed, line 162: An odd part of this for sure.. but felt like this was sufficient until we start passing around `Window` refs.

cbeer (approved):

PR #3298, opened 2020-10-20
Constrain video height to window height

### LINKED ISSUE #3295: Video controls are inaccessible for some window aspect ratios
It appears that video controls become inaccessible when a window's aspect ratio is "wider" than the video. For example, when viewed on a 1920x1080 monitor:

mirador-video-controls-inaccessible-chrome-85

Adding `max-height: 100%` as a style rule on `<video>` seems to fix this:

mirador-video-controls-accessible-chrome-85

I'm seeing this issue with the deployment preview for https://github.com/ProjectMirador/mirador/pull/3264 at as well as a local instance of v3.0.0-rc.7, in both Chrome 85 and Firefox 81.

camillevilla:
Fixes #3295
Thanks @markmatney! Your suggestion was right on point.

## Before
wide video player with inaccessible controls

## After
wide video player with controls visible

### DIFF
--- a/src/containers/VideoViewer.js
+++ b/src/containers/VideoViewer.js
@@ -22,6 +22,7 @@ const styles = () => ({
     width: '100%',
   },
   video: {
+    maxHeight: '100%',
     width: '100%',
   },
 });


cbeer (approved):

markmatney:
Great :smile: we're excited for Mirador 3! Thank you!

Issue #2949, opened 2020-03-06
Request for feature that displays option to select from multiple sequences in a manifest.


MImranAsghar:
Request for a feature in Mirador 3 that displays multiple sequences to select from in the sidebar when the manifest is loaded. This would ensure that a manifest with multiple sequences works with Mirador 3. For example, a manifest like this: https://www.e-codices.unifr.ch/metadata/iiif/gau-Fragment/manifest.json

Example that contains at least two defined sequences for the order of the object views/canvases. This will help test the sidebar affordance that enables the user to choose a sequence in which to view the canvases.

mejackreed:
From design mockups, this is the original intent of how to insert a sequence (using the select drop down).

![Screen Shot 2020-03-06 at 9 18 46 AM](https://user-images.githubusercontent.com/1656824/76101382-939f3980-5f8b-11ea-828f-88a111c30690.png)

via https://xd.adobe.com/view/2b634206-c940-4835-5f00-30be1edfd7a3-5d52/screen/e2af26c1-1db8-49d8-bae5-11ccc5c90e4d/index-Alt-Seq

@MImranAsghar were you thinking something similar?

MImranAsghar:
Basically, I would say viewing the alternate sequences instead of inserting. So, let's say if we already had a manifest with multiple sequences, then something like this where I can select an 'Alternate sequence' and after selection, it would display that sequence on the window.

giacomomarchioro:
Hello, we are writing a cookbook recipe for addressing this issue https://github.com/IIIF/cookbook-recipes/issues/27 we designed a mockup manifest with two sequences (https://api.npoint.io/c6611ef8b543b3beeee9) we can select the second sequence using the dropdown menu but the app freezes when we try to get the next page:

<img width="1083" height="443" alt="Image" src="https://github.com/user-attachments/assets/a9a762a5-b21b-44bb-86d1-a9c7c2cc1fd5" />

We got this error:

`canvases.js:21 Uncaught TypeError: e.getCanvases is not a function at canvases.js:21:37`

I was wondering if there are any updates or comments on this topic and about our proposed manifest.

marlo-longley:
@giacomomarchioro Hello. My read on your issue is that `getCanvases` is not implemented for IIIF v3 Ranges yet in the manifesto.js library, which we rely on here. 

[...]

PR #3299, opened 2020-10-21
Display multi seq on ui

MImranAsghar:
Adds multi seq UI display for https://github.com/ProjectMirador/mirador/issues/2949
  - Shows sequences as collapsible buttons in sidebar
Updates sequences selector in master by adding getSequenceCanvases selector

Before

Before

After UI update

After_1
After_2
After_3

### DIFF
--- a/__tests__/fixtures/version-2/multipleSequences.json
+++ b/__tests__/fixtures/version-2/multipleSequences.json
@@ -1,0 +1,1083 @@
+{
+  "@context": "http://iiif.io/api/presentation/2/context.json",
+  "@id": "https://www.e-codices.unifr.ch/metadata/iiif/gau-Fragment/manifest.json",
+  "@type": "sc:Manifest",
+  "label": "Urn\u00e4sch, Gemeindearchiv Urn\u00e4sch, Fragment",
+  "metadata": [
+    {
+      "label": "Location",
+      "value": "Urn\u00e4sch"
+    },
+    {
+      "label": "Collection Name",
+      "value": "Gemeindearchiv Urn\u00e4sch"
+    },
+    {
+      "label": "Shelfmark",
+      "value": "Fragment"
+    },
+    {
+      "label": "Document Type",
+      "value": "Fragment"
+    },
+    {
+      "label": "DOI",
+      "value": "10.5076/e-codices-gau-Fragment"
+    },
+    {
+      "label": "Title (English)",
+      "value": "Psalterium iuxta Hebraeos (Fragment)"
+    },
+    {
+      "label": "Material",
+      "value": "Parchment"
+    },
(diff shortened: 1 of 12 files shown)


mejackreed:
Thanks @MImranAsghar . Are there any public manfiests that you could link to here where we could look at this feature? It might help make evaluation of the design of this PR easier.

MImranAsghar:
Hi Jack, here is one I use often https://www.e-codices.unifr.ch/metadata/iiif/gau-Fragment/manifest.json

mejackreed (changes requested):
@MImranAsghar thanks so much for submitting this PR. We especially like how you added tests here for the work you added!

While reviewing with @ggeisler there is some concern about the collapsible panels and were wondering if you would be open to either:
 - adapting this pull request to use the designed select dropdown for sequences
 - open to pull requests that implement this feature

An example of this here:

Screen Shot 2020-10-21 at 5 27 17 PM

There is a desire to keep the toggle buttons to select index/thumbnails but would preference sequence selection from the select dropdown.

MImranAsghar:
Hi @mejackreed , I see, no worries :) For the select dropdowns on ui, I assume it was done before here https://github.com/ProjectMirador/mirador/compare/2949-multi-seq#diff-dd9bcb599eb3ffafce45ff153b58fa4bbcc455502e09c255fdf43c91f4b31c91R98 but seems like this would have to be done differently now because WindowSideBarCanvasPanel code in master has changed from when it was done? If so, I can adapt this pull request (https://github.com/ProjectMirador/mirador/pull/3299) I made yesterday for collapsible panels to display select dropdown instead.

mejackreed:
@MImranAsghar yes for sure some things have changed. However, I think the seam might be similar implemented as part of the `titleControls` here: https://github.com/ProjectMirador/mirador/blob/master/src/components/WindowSideBarCanvasPanel.js#L86-L100

We should maybe just add above (or maybe within a fragment) the formcontrol selection:

https://github.com/ProjectMirador/mirador/compare/2949-multi-seq#diff-dd9bcb599eb3ffafce45ff153b58fa4bbcc455502e09c255fdf43c91f4b31c91R98-R128

MImranAsghar:
@mejackreed Ah okay true, makes sense 👍 I guess i will move the formcontrol selection in that case. I will also add some tests for that.

MImranAsghar:
Hi @mejackreed , so I have updated the select dropdown UI from (https://github.com/ProjectMirador/mirador/tree/2949-multi-seq). Updated it to work with the latest master. I have created a fresh pull request https://github.com/ProjectMirador/mirador/pull/3304

mejackreed:
Superceded by #3304

PR #3304, opened 2020-10-26
Multi seqs display

MImranAsghar:
Update select dropdown UI from (https://github.com/ProjectMirador/mirador/tree/2949-multi-seq). Update it to work with the latest master; add some tests for multi-seq disp.

### DIFF
--- a/__tests__/integration/mirador/blank.html
+++ b/__tests__/integration/mirador/blank.html
@@ -13,7 +13,9 @@
     <script type="text/javascript">
      var miradorInstance = Mirador.viewer({
        id: 'mirador',
-       windows: [],
+       windows: [{
+         manifestId: 'https://www.e-codices.unifr.ch/metadata/iiif/gau-Fragment/manifest.json'
+       }],
        theme: {
          transitions: window.location.port === '4488' ?  { create: () => 'none' } : {},
        },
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -35,6 +35,7 @@
     "@material-ui/icons": "^4.9.1",
     "@material-ui/lab": "^4.0.0-alpha.53",
     "@researchgate/react-intersection-observer": "^1.0.0",
+    "canvas": "^2.6.1",
```
mejackreed, line 38: Can you say more about this dependency that is added?

MImranAsghar: Sorry was testing something, just removed it

__tests__/integration/mirador/blank.html
```diff
@@ -13,7 +13,9 @@
     <script type="text/javascript">
      var miradorInstance = Mirador.viewer({
        id: 'mirador',
-       windows: [],
+       windows: [{
+         manifestId: 'https://www.e-codices.unifr.ch/metadata/iiif/gau-Fragment/manifest.json'
```
mejackreed, line 17: Can we remove the added manifest here, if needed we can create a new page "multiplesequence.html" or something similar?

MImranAsghar: Hey, just pushed

mejackreed (approved):
Thanks for pushing this work forward @MImranAsghar 👏

PR #3336, opened 2020-11-03
Remove RootRef and fix scrolling entire page

### LINKED ISSUE #3302: Clicking on items in canvas pane pushes entire viewer window up
Clicking on a items in a canvas pane pushes the entire viewer window up.

This is cool:
Screen Shot 2020-10-23 at 1 52 22 PM

But this is not:
Screen Shot 2020-10-23 at 1 56 32 PM

Per @mejackreed on Slack this may be a bug in the way the scroll state of canvases is saved.

mejackreed:
Fixes some underlying inconsistencies with scrolling behavior including:

 - Making sure that the ScrollTo component is the child
 - Removing RootRef which is removed from Material-UI and deprecated in
React.StrictMode
 - While scrollIntoView is awesome, it scrolls the entire page even when
in an iframe which can cause unexpected jumping.
 - Introduces `withRef` a HOC which is only used in `CompanionWindow` but we could use elsewhere

Fixes #3302

### DIFF
--- a/src/components/ScrollTo.js
+++ b/src/components/ScrollTo.js
@@ -1,6 +1,5 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-import RootRef from '@material-ui/core/RootRef';
 
 /**
  * ScrollTo ~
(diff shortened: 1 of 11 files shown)


src/components/ScrollTo.js
```diff
@@ -58,6 +58,17 @@ export class ScrollTo extends Component {
...
 
+  /**
+   * The container provided in the containersRef dome structure in which scrolling
+   * should happen.
+  */
+  scrollabelContainer() {
+    const { containerRef } = this.props;
+
+    if (!containerRef || !containerRef.current || !containerRef.current.domEl) return null;
+    return containerRef.current.domEl.getElementsByClassName('mirador-scrollto-scrollable')[0];
```
mejackreed, line 69: Unfortunately we need to reach into the dom here. We need to understand both the:  component that is sized only in view `containerRef` and the scrollable portion of this on which we need to scroll against.

cbeer (approved):

PR #3350, opened 2020-11-23
Missing dl-tag

boeschung:
For displaying the "rights" there is missing a `dl`-Tag, which is wrapping the `dt`- und `dd`-tags.


mejackreed (approved):
Thanks for the contribution!

PR #3356, opened 2020-11-27
Draft: Fix fragment rendering on scaled canvases in bookview

jbaiter:
Previously, when one of the two canvases was scaled in book-view, the search annotations on this canvas would not match the page image.
This PR fixes this by passing a scaling factor from `CanvasWorld` to `CanvasAnnotationDisplay`, multiplication with which fixes the offset fragment so it matches the rendered canvas image.

<details>
<summary>Before</summary>

image
</details>

<details>
<summary>After</summary>

image
</details>

Unfortunately the test manifest with the content search service is not public (yet), so I can't add it to the integration test suite :-/

To be honest, I'm not completely confident that this change is the best way to fix this behavior, since I have not yet completely understood how the scaling in bookview relates to the annotation overlay, but this seems to fix it in the cases I tested. If there's a better way to do this, let me know and I'll implement it :-)

-- Edit: **Don't merge yet,** I just found a case where this fix doesn't work, investigating...

### DIFF
--- a/src/lib/CanvasAnnotationDisplay.js
+++ b/src/lib/CanvasAnnotationDisplay.js
@@ -5,7 +5,7 @@
 export default class CanvasAnnotationDisplay {
   /** */
   constructor({
-    resource, palette, zoomRatio, offset, selected, hovered,
+    resource, palette, zoomRatio, offset, selected, hovered, scaleFactor,
   }) {
     this.resource = resource;
     this.palette = palette;
@@ -13,6 +13,7 @@ export default class CanvasAnnotationDisplay {
     this.offset = offset;
     this.selected = selected;
     this.hovered = hovered;
+    this.scaleFactor = scaleFactor;
   }
 
   /** */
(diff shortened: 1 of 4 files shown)


src/lib/CanvasAnnotationDisplay.js
```diff
@@ -104,9 +105,12 @@ export default class CanvasAnnotationDisplay {
 
   /** */
   fragmentContext() {
-    const fragment = this.resource.fragmentSelector;
+    let fragment = this.resource.fragmentSelector;
     fragment[0] += this.offset.x;
```
mejackreed, line 109: I'm wondering now if perhaps we are calculating the offset incorrectly? Scaling the offsets before applying seem to be a useful transformation.

mejackreed: So I'm thinking we might need to do the following:

1. Better calculate an offset based off of the scaling
2. Resize the fragment/svg based off of scaling

jbaiter:
I'm afraid I have to give up on this issue for now :-/ 
I created a small fixture based on another one by @mejackreed that demonstrates the issue with a canvas that is downscaled from the image, and another one that is upscaled. For neither canvas t he annotations render correctly:
https://gist.github.com/jbaiter/8d4889654d87ef400bd6dcef353cdd8f

Hopefully someone smarter than me can pick this up :-)

marlo-longley:
Closing this as stale during our audit, but there is great work tracked here. 
Since the related ticket #3469 is still open and linked to this PR, the work is not lost. Thank you for your work!

PR #3359, opened 2020-12-02
Update Serbian translation

DigitLib:
Added also translation for Collection dialog for 'collection' and button 'close'.

PR #3360, opened 2020-12-02
Added translatable collection dialog

DigitLib:
Added COLLECTION and CLOSE button translatable.

PR #3361, opened 2020-12-02
Add missing en keys

mejackreed:
Related to #3359 

Thanks to @DigitLib for pointing out these were used in `CollectionDialog` but never added to our translation files. 👏


mejackreed:
Duplicate of #3360

PR #3370, opened 2020-12-08
Local developer experience improvement have the ability to add local …

mejackreed:
…manifests

Kapture 2020-12-08 at 11 43 23


camillevilla (approved):
Oooooh! This should be really handy for future IIIF workshops. @glenrobson @hadro

PR #3372, opened 2020-12-10
Fix bad plugin hook name in AttributionPanel

jbaiter:
Seems this was missed during copy-pasting from `ManifestInfo` :-)

I grepped the codebase for other instances of a mismatch between Component name and plugin hook identifier, but found none, so this seems to be the only instance.

### DIFF
--- a/src/containers/AttributionPanel.js
+++ b/src/containers/AttributionPanel.js
@@ -46,7 +46,7 @@ const enhance = compose(
   withStyles(styles),
   withTranslation(),
   connect(mapStateToProps),
-  withPlugins('ManifestInfo'),
+  withPlugins('AttributionPanel'),
 );
 
 export default enhance(AttributionPanel);


cbeer (approved):

src/containers/AttributionPanel.js
```diff
@@ -46,7 +46,7 @@ const enhance = compose(
   withStyles(styles),
   withTranslation(),
   connect(mapStateToProps),
-  withPlugins('ManifestInfo'),
+  withPlugins('AttributionPanel'),
```
mejackreed, line 49: Nice catch

jbaiter: It actually caused a bug in our instance, since we had a plugin targetting `ManifestInfo` that suddenly broke the attribution panel :sweat_smile:

PR #3379, opened 2021-01-08
Fixes an issue with A/V to allow configurable element attribution and crossOrigin options

mejackreed:
Enables actual use of VTT via `crossorigin="anonymous"` See: https://preview.iiif.io/cookbook/0219-using-caption-file/recipe/0219-using-caption-file/manifest.json


lutzhelm (dismissed):
This looks really good to me. I played around with some more audio / video options; whether they work or not is dependent on the browser.
Just one thought: should we allow configuring (i.e. intentionally leaving out) the `controls` attribute?

lutzhelm (approved):
:thumbsup: Changes to `controls` look fine and do work.

Issue #3382, opened 2021-01-27
Optionally pass the mount element to the viewer constructor directly


charbugs:
When instantiating Mirador you have to pass an element id to the `Mirador.viewer` function. The constructor then searches for the element by `document.getElementById` and renders the viewer to that element. The `mirador.unmount` function again searches for that element by id and unmounts the viewer from that element. That means the constructor/unmount functions assume that the element is already/still connected to the DOM. But this may not always be the case.

For example a part of your code that is not aware of Mirador may remove the mount element without warning. In this case you have no chance to tell Mirador to deconstruct properly because the `unmount` function will fail. This may result in dangling style-elements under the head element. Another use case is that you want instantiate Mirador to perform some actions on it but render it later. This is currently not possible because the mount element must be connected to the DOM.

I propose to create an option to pass the mount element directly to the `Mirador.viewer` function. That way the constructor and unmount function would work even if the mount element is not connected to the DOM yet/anymore.

cbeer:
I think #3429 may have addressed this/part of this.

PR #3383, opened 2021-01-27
Option to pass mount element directly to constructor #3382

charbugs:
See #3382

### DIFF
--- a/__tests__/integration/mirador/unconnected.html
+++ b/__tests__/integration/mirador/unconnected.html
@@ -1,0 +1,50 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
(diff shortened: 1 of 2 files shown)


mejackreed (commented):
I really like the ability to mount on an existing dom element.

One question I have is about our usage of `containerId` https://github.com/ProjectMirador/mirador/search?q=containerId

And if you have seen any issues with not having that? Some of the interaction feature rely on that being available and being unique.

Should we require an id in the existing element and then tract that? Or do something else here?

charbugs:
Oh snap! I didn't noticed that because in my demos I always used an ID atomatically. I also noticed some other warnings depending on what actions you are dispatch while the container element isn't connected.

I will close the PR for now.

Meanwile i think a more explicit solution to accomplish what  I described in the issue would be to have an explicit `mount` method on the instance object. Something like:

```
const mirador = Mirador.viewer(..)
// do something with the instance, e.q. dispatch some actions on it
mirador.mount()
```

PR #3384, opened 2021-02-08
Accessibility review

morpheus-87:
This PR fixes some things that were found in our accessability review:
- change element for search hit title from h6 to a div, our reviewers said wrapping them in a headline wasn’t neccessary, since it’s already part of a `<ul>`, which suffices for screenreaders to properly interpret the structure
- make some accessability labels more meaningful
- add some missing `aria` attributes


ggeisler (changes requested):
👍 on the H6 -> div update, and the "Submit search" to "Search in the document" aria label update.

On changing the sidebar toggle button label from "Toggle sidebar" to "Show / hide sidebar" I'm not as sure. I have a couple of thoughts about this:

- I see the comment from the accessibility review report (note that I'm reading the English translation, which I believe is just Google Translate, so it's possible the advice is different in the original German report) on page 17 about how the German-language button label is not meaningful ("Switch sidebar" is what the translated report says it reads). But "Toggle sidebar" **is** meaningful in English, so one solution is to leave the English label as is and just update the German translation to whatever makes more sense in German than the literal translation of "toggle".

- If we do want to change the label to "Show / hide sidebar" I think that would be great, but only if we can use the specific label for the current state of the sidebar. For instance, if we can display the label "Show sidebar" when the sidebar is closed, and "Hide sidebar" when the sidebar is open. I think that is better than "Toggle sidebar". But "Show / hide sidebar" for both states feels like we are taking a shortcut. And it's inconsistent with the nearby expand/collapse sidebar button, where we do show the appropriate label for the current state ("Collapse sidebar" when it's open and "Expand sidebar" when it's closed).

morpheus-87:
Thank you very much for your feedback, @ggeisler ! The text from our accessability review was the following:
> The "Toggle sidebar" and "Close sidebar" controls relate to different elements. You must clearly name the respective element. The name of the control element "Toggle sidebar" does not describe the function very clearly and should, for example, be named in "Show / hide sidebar".
But you are right, the label should depend on the current state of the sidebar :+1: I will fix that.
I have just another question: is there any reason to have two elements (see the following screenshot) for toggling the sidebar?
screenshot-www-dev digitale-sammlungen de-2021 02 10-12_04_55

morpheus-87:
I also just added one more `aria-expanded` attribute.

ggeisler:
@morpheus-87 Thanks for working on the updates!

> I have just another question: is there any reason to have two elements (see the following screenshot) for toggling the sidebar?

The difference between the two elements is subtle, but intentional.

### Toggle sidebar

Screen Shot 2021-02-10 at 10 55 54 AM

---

This enables an implementer to configure the sidebar to be completely closed by default, so the user can focus on the workspace area without any interference from any UI controls in the workspace area. We do this at Stanford, for example, when using the Mirador viewer as an item viewer in various catalog-like pages, where the item metadata is often already on the page outside of the viewer and we don't expect most users to need to use the sidebar at all. So we give the user the cleanest view of the item but also an easy way to open the sidebar if they want to.


### Expand/collapse sidebar

Screen Shot 2021-02-10 at 10 55 01 AM

---

This enables the user who does open the sidebar to collapse it to de-clutter the workspace (when zooming and panning on the image, for example) but the visible tab to re-expand the sidebar serves as a reminder of sorts that the sidebar can be re-expanded and that there are other sidebar categories available. (Notice that when the sidebar is collapsed, the sidebar menu is still visible, while when the sidebar is closed it is not.) Again, I know the difference is subtle, but the idea is that the visible tab might be more handy for the user who is opening and closing the sidebar frequently while they examine and work with the item, looking at the different sidebar category content, etc.

ggeisler (dismissed):
I don't really have a Mirador dev environment working so I can't easily see these updates, but based on the file changes and what @morpheus-87 says he is trying to do, I assume this is good to go. I can double-check after the PR is merged and the updates are visible in the Mirador demo instance.

EDIT: I forgot about the Netlify preview. I checked that out and it looks good to me.

morpheus-87:
Thank you very much for your explanation, @ggeisler ! That makes totally sense to me :+1:

I also tried to fix the integration tests, but I don't understand, why one of them is failing :thinking:

morpheus-87:
Could someone please help me with the integration tests? Perhaps @mejackreed ?

morpheus-87:
I just updated my branch to `main`, is there a chance that this long living PR could be merged?

marlo-longley (approved):
@morpheus-87 thank you for stewarding this work across so many years! Very much appreciated, and great it could be rebased.

PR #3388, opened 2021-02-25
Upd README +examples +synced with current app state

RipDevil:
I am tightly working with Mirador and it's documentation doesn't present perfectly how to use it. I am planning on contributing in it a little if I see feedback from the maintainers :)

Any fixes to this request are welcome

Added some real (but still simple) examples: a basic and a react ones. Also changed how the README looks in general.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,32 +1,194 @@
-*NOTE: This README reflects the latest version of Mirador, Mirador 3. For previous versions, please reference that release's README directly. Latest 2.x release: [v.2.7.0](https://github.com/ProjectMirador/mirador/tree/v2.7.0)*
-# Mirador
-![Node.js CI](https://github.com/ProjectMirador/mirador/workflows/Node.js%20CI/badge.svg) [![codecov](https://codecov.io/gh/ProjectMirador/mirador/branch/master/graph/badge.svg)](https://codecov.io/gh/ProjectMirador/mirador) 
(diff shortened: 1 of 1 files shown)


mejackreed (commented):
Thank you so much for your contribution @RipDevil .

Have you seen this document? https://docs.google.com/document/d/1_LPOcwMOM051qzhqMixqLLKXZ5SH9Vjd3_UA-tZniek/edit#heading=h.jrn1bus9glbi

It has some information that could also be addressed. Would you be ok if I merged this and then followed up with some changes? Or would inline suggestions be best?

RipDevil:
@mejackreed feel free to change after merging :)

charbugs:
@RipDevil Basically it's a good idea the consolidate the installation and usage instruction on the project startpage (readme.md). 

But I think we should advise people to use the precompiled Mirador bundle (embed via `<script>` at runtime) where ever possible and mark the build time integration as an advanced approach that may be harder to maintain.

If I'm not wrong the only need to import Mirador at build time is for plugin integration because for several reasons errors may occure if one applies plugins at runtime.

So I think the runtime integration approach should be in focus in the readme file.

RipDevil:
@charbugs

If you think that's a sound idea, the altered readme.md file can be extended after merging this branch. 

My motivation here was to brush up the readme so it'd be more suitable with current Mirador version, easier to read and it will reflect usage. 

Probably I am biased here because I described how I use it thus we should use changes from the current branch and expand it 😁

_Not related: I also want to make some kind of a list of Mirador options with description. Like... a separate file in subdurectory where default options are placed._

RipDevil:
@mejackreed  @camillevilla 
Should we close this pull request if there is no need in it?

PR #3390, opened 2021-02-25
search: Improved support for paginated search responses

jbaiter:
We frequently have search responses with 100+ hits, which leads to really bad performance with Mirador 3 if fetched all at once. We switched to paginated responses of 50 hits each, which improved the problem somewhat (at least until you request more hits a few times).
However, one thing missing from the current implementation is some kind of indicator or display about how many hits are available in total (available from the `.within.total` field on the Content Search response).

This PR adds this in two places:

[image](https://i.imgur.com/e6uXB4V.mp4)

- Indicate the availability of more hits than currently shown with a `+` suffix after the "total" length in `SearchPanelNavigation`
- Indicate how many hits remain to be fetched with an extra line on the `More Results` button in the `SearchResults` component.

I only included i18n strings for German and English, so this is something that would have to be updated for the other languages as well.


src/components/SearchResults.js
```diff
@@ -130,7 +130,7 @@ export class SearchResults extends Component {
           >
             {t('moreResults')}
             <br />
-            {`(${t('resultsRemaining', { numLeft: searchNumTotal - searchHits.length })})`}
+            {`(${t('searchResultsRemaining', { numLeft: searchNumTotal - searchHits.length })})`}
```
mejackreed, line 133: Could we leave this key the same to retain semver?

mejackreed (changes requested):
Thanks for the contribution. I'm of course good with the updated German translation, but I wonder if we need to retain the same keys for semver compatibility?

mejackreed (approved):

PR #3393, opened 2021-03-03
Iiif tileformat option

git-arbitrarysystems:
Dear Mirador team,

At the Museum für Naturkunde Berlin we are developing a search portal for digitized objects. For image visualizations we use a IIIF server that returns webp images. We are trying out Mirador as an external visualisation-tool, and needed to get Mirador to use webp images. 

This branch adds a `tileFormat` property to the window object:
```
windows:[{
         manifestId: 'https://portal.museumfuernaturkunde.berlin/api/v1/iiif/manifest/183611/manifest.json', 
         tileFormat:'webp' // defaults to jpg
}]
```
as well as a to catalog entries:
```
catalog:[
{ 
         manifestId: 'https://portal.museumfuernaturkunde.berlin/api/v1/iiif/manifest/183611/manifest.json',  
         provider:'Museum für Naturkunde Berlin', 
         tileFormat:'webp' 
}]
```

All thumbnails and canvases will now check for the `tileFormat` property and if found, request images in this format. If no such property is found, it defaults to jpg. 

I hope the changes i made are appropriate and i would be happy to see them in a future release. 

Yours truly,
Jasper Funk-Smit
( Frontend developer at the Museum für Naturkunde - Berlin )


git-arbitrarysystems:
I removed the dist/ from the repository. Yes i saw the `preferredFormats` option in the IIIF 3.0 options. Using that should be the preferred option. I don't know how far the usage of 3.0 options is implemented in mirador, but when the time is ready you can let me know and i can assist in implementing `preferredFormats`.

cbeer:
Thanks for the PR. I need to come up to speed on this. If I understand the problem correctly, it's that... mirador has no way to know (reliably) what formats are preferred/supported without getting the info.json document, but jpg is a universal requirement from the IIIF specification, so it's always safe even if it's not the best option.

Here's  a couple concerns I have:

- it looks like openseadragon already does format negotiation using the IIIF v3 preferredFormats: https://github.com/openseadragon/openseadragon/blob/master/src/iiiftilesource.js#L243 . I wonder if it's best to just let OSD figure out what works instead of trying to force its behavior (and.. assume content providers know best?)
- i'm concerned about having to hard-code particular formats into the mirador configuration. It looks like this configuration is at the window or catalog entries, but what if the provider changes their available formats, or the user adds a manifest from some other location, or the user's browser doesn't support a particular format.

I'll give it a little more thought and see if I can offer a suggestion or possible direction.

cbeer:
@azaroth42 suggested:

> e.g. in the example here: https://iiif.io/api/presentation/3.0/#57-content-resources they could simply add preferredFormats: [ "webp" ] after profile: level2 in the ImageService

That would, at least, give the manifest an opportunity to suggest appropriate formats without needing to request the info.json each time. I _think_ that the `IIIFThumbnail` class probably has enough information already to make that choice, so at least we wouldn't need to maintain the window or catalog state through the various components.

I don't think that resolves the question of what browsers might support, though. I don't know if we want to add some global configuration for supported formats (like OpenSeadragon) or allow `IIIFThumbnail` to return multiple urls and use e.g. `<picture>` with appropriate fallbacks to allow the browser to decide.

PR #3394, opened 2021-03-03
Support two additional modes of canvas thumbnail specification

markmatney:
These changes add support for two modes of specifying exactly which thumbnails to fetch for a particular canvas (`'with a canvas'`):
- `'uses the width and height of a thumbnail without a IIIF Image API service'`
- `'uses embedded sizes of a IIIF Image API service to find an appropriate size'`

The several `test-tightening` commits are changes that I saw as necessary to make the test cases either easier (IMO) to reason about, more aligned to ThumbnailFactory's actual usage, or more specific (eliminating false positives), so that I could better ensure that I wasn't breaking existing functionality. For instance:
- while `{ "url": "myId" }` and `{ "url": "myId", "width": 80, "height": 120 }` may both cause `expect({ "url": "myId", "width": 80, "height": 120 }).toMatchObject(...)` to return true, only the latter will cause `toEqual(...)` to do the same
- the application never calls the ThumbnailFactory directly on an Image, so I wrapped each test Image in a Canvas
- callers also always provide at least one size constraint on the UI (`maxWidth`, `maxHeight`), so I made sure it was always getting passed at least one of the two
- the construction of the test data using combinations of `createSubject`, `createImageSubject`, `iiifService`, and the `...` operator often made it difficult for me to tell what the test data actually looked like, so I deleted those "helper" methods (I believe the test data should be further moved to a test fixture JSON file)

The most important code changes are at `3b9526c` and `d845c5e`; the former (IMO) clarifies how an Image is selected as a candidate from which to derive a "thumbnail" by using two methods to do so rather than five, and the latter is necessary for the two tests I've added to pass.

I hoped to make it possible, by allowing one to walk through a series of focused commits, to observe that I'm not breaking existing functionality. Also, I see that the four `test-tightening` commits at the tip of this branch might be good to group with the rest of them at the base, and I am happy to attempt a rebase if desired.

(P.S. These changes also add one test (`'with a manifest'` > `'does nothing with a plain URL'`) that I saw that was missing, for functionality that I inadvertently broke at one point while making these changes.)

### DIFF
--- a/__tests__/src/lib/ThumbnailFactory.test.js
+++ b/__tests__/src/lib/ThumbnailFactory.test.js
@@ -48,39 +48,37 @@ describe('getThumbnail', () => {
 
   describe('with a IIIF resource', () => {
     for (const type of ['Collection', 'Manifest', 'Canvas', 'Image']) {
+      describe('with a thumbnail', () => {
+        it('return the thumbnail and metadata', () => {
+          expect(createSubject({ '@id': 'xyz', '@type': type, thumbnail: { '@id': url, height: 70, width: 50 } }, type)).toMatchObject({ height: 70, url, width: 50 });
+        });
 
-  describe('with a thumbnail', () => {
-    it('return the thumbnail and metadata', () => {
-      expect(createSubject({ '@id': 'xyz', '@type': type, thumbnail: { '@id': url, height: 70, width: 50 } }, type)).toMatchObject({ height: 70, url, width: 50 });
-    });
+        it('return the IIIF service of the thumbnail', () => {
+          expect(createSubject({ '@id': 'xyz', '@type': type, thumbnail: iiifLevel1Service }, type)).toMatchObject({ url: `${url}/full/,120/0/default.jpg` });
+        });
 
-    it('return the IIIF service of the thumbnail', () => {
-      expect(createSubject({ '@id': 'xyz', '@type': type, thumbnail: iiifLevel1Service }, type)).toMatchObject({ url: `${url}/full/,120/0/default.jpg` });
-    });
+        describe('with image size constraints', () => {
+          it('does nothing with a static resource', () => {
+            expect(createSubject({ '@id': 'xyz', '@type': type, thumbnail: { '@id': url } }, type, { maxWidth: 50 })).toMatchObject({ url });
+          });
 
-    describe('with image size constraints', () => {
-      it('does nothing with a static resource', () => {
-        expect(createSubject({ '@id': 'xyz', '@type': type, thumbnail: { '@id': url } }, type, { maxWidth: 50 })).toMatchObject({ url });
+          it('does nothing with a IIIF level 0 service', () => {
+            expect(createSubject({ '@id': 'xyz', '@type': type, thumbnail: iiifLevel0Service }, type, { maxWidth: 50 })).toMatchObject({ url: 'arbitrary-url' });
+          });
+
+          it('calculates constraints for a IIIF level 1 service', () => {
+            expect(createSubject({ '@id': 'xyz', '@type': type, thumbnail: iiifLevel1Service }, type, { maxWidth: 150 })).toMatchObject({ height: 300, url: `${url}/full/150,/0/default.jpg`, width: 150 });
+          });
+
+          it('calculates constraints for a IIIF level 2 service', () => {
+            expect(createSubject({ '@id': 'xyz', '@type': type, thumbnail: iiifLevel2Service }, type, { maxHeight: 200, maxWidth: 150 })).toMatchObject({ height: 200, url: `${url}/full/!150,200/0/default.jpg`, width: 100 });
+          });
+
+          it('applies a minumum size to image constraints to encourage asset reuse', () => {
+            expect(createSubject({ '@id': 'xyz', '@type': type, thumbnail: iiifLevel2Service }, type, { maxHeight: 100, maxWidth: 100 })).toMatchObject({ height: 120, url: `${url}/full/!120,120/0/default.jpg`, width: 60 });
+          });
+        });
       });
-
-      it('does nothing with a IIIF level 0 service', () => {
-        expect(createSubject({ '@id': 'xyz', '@type': type, thumbnail: iiifLevel0Service }, type, { maxWidth: 50 })).toMatchObject({ url: 'arbitrary-url' });
-      });
-
-      it('calculates constraints for a IIIF level 1 service', () => {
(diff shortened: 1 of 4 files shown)


src/lib/ThumbnailFactory.js
```diff
@@ -184,83 +208,77 @@ class ThumbnailFactory {
...
-  return new ThumbnailFactory(resource, iiifOpts).get();
+    // Special treatment for external resources
+    if (typeof sourceContentResource === 'string') return { url: sourceContentResource };
+
+    return this.iiifThumbnailUrl(sourceContentResource);
+  }
 }
 
-export default getBestThumbnail;
+export default ThumbnailFactory;
```
mejackreed, line 284: This now changes the export of this file which we would consider a breaking change only available in the next major version. Can we maintain the previous export API? I think this was an intentional consideration.

src/components/IIIFThumbnail.js
```diff
@@ -162,14 +161,12 @@ export class IIIFThumbnail extends Component {
             </Typography>
           </div>
         )}
-        {children}
```
mejackreed, line 165: Removing the `children` prop rendering here might also be considered a breaking change. Can we please add this back?

__tests__/src/lib/ThumbnailFactory.test.js
```diff
@@ -1,155 +1,356 @@
-import { ManifestResource, Resource, Utils } from 'manifesto.js/dist-esmodule';
-import getThumbnail from '../../../src/lib/ThumbnailFactory';
+import { Canvas, Service, Utils } from 'manifesto.js/dist-esmodule';
+import ThumbnailFactory from '../../../src/lib/ThumbnailFactory';
 import fixture from '../../fixtures/version-2/019.json';
 
-const manifest = Utils.parseManifest(fixture);
```
mejackreed, line 5: One thing as a reviewer would help me would be to see the added changes separately from the changes of the underlying test suite. Would it be possible to extract only the modified pieces needed for the new functionality from here?

__tests__/src/lib/ThumbnailFactory.test.js
```diff
@@ -194,7 +400,42 @@ describe('getThumbnail', () => {
         ],
         type: 'Collection',
       });
-      expect(getThumbnail(collection)).toMatchObject({ url: 'https://example.org/manifest1/thumbnail.jpg' });
+      expect(getThumbnail(collection, { maxHeight: 120 })).toEqual({ url: 'https://example.org/manifest1/thumbnail.jpg' });
+    });
+  });
+
+  describe('selectBestImageSize', () => {
```
mejackreed, line 407: I really like how you were able to add some tests  around this. 👏 Because of the API change that we cannot make, is it possible to add this as an additional export so it can be tested?

src/lib/ThumbnailFactory.js
mejackreed: Maybe @cbeer has some thoughts on the intention behind the API export here?

mejackreed (changes requested):
👏 @markmatney for the excellent work here. I've added a few comments about backwards compatibility that we probably need to maintain. Also I'm wondering how difficult would it be to remove the testing changes done so that could be reviewed separately? I would be happy to pair up to see if we could work on that together.

I'm curious too from others including @cbeer and @regisrob who have worked in these areas before.

markmatney:
I think moving the test suite rewrite to a separate PR is a good idea and would be fairly straightforward.

__tests__/src/lib/ThumbnailFactory.test.js
markmatney: I am rather new to ES6 so I may be wrong, but I don't believe it is possible to add an additional export to ThumbnailFactory without breaking its API. E.g., if we were to do instead (in ThumbnailFactory.js)

```export { ThumbnailFactory, getBestThumbnail };```

then imports would have to change from

```import getBestThumbnail from '../../lib/ThumbnailFactory';```

to

```import { getBestThumbnail } from '../../lib/ThumbnailFactory';```

That would be a breaking change, right?

(A workaround would be to move `selectBestImageSize` to a different module.)

__tests__/src/lib/ThumbnailFactory.test.js
markmatney: sure, I can work on creating a new PR for the rewrite of the test file

__tests__/src/lib/ThumbnailFactory.test.js
mejackreed: Could something like this work?

`export { getBestThumbnail as default, ThumbnailFactory };` 

That way, the default export does not change (the API) but you can provide additional exports

markmatney: Getting back to this now -- it looks like this works! Will do.

markmatney:
This PR now includes a minimal change-set for implementing this feature. I've cached locally the test suite rewrite which I can create another PR for later on.

I kept 34e4d49 separate from ea547a8 so that one could compute a word-level diff on each, which I think makes the changes easier to see than if they were combined.

I see that the codeclimate check is flagging issues with `ThumbnailFactory.getSourceContentResource`. That method is recursive, which I thought was the best way to implement it given that it traverses the tree structure of the JSON-LD. I think it's not really any more complex than the code that was there before, and that `ThumbnailFactory.iiifThumbnailUrl` is also quite complex (perhaps moreso) despite not being flagged by codeclimate. Just my 2 cents 🙂

markmatney:
(FYI, the code at e2220c1 depends on changes introduced in ea547a8, so the build will fail if you checkout the former commit. Rebase error 😉 )

mejackreed:
++ @markmatney thanks for making the changes. Things are looking good, I am seeing some new/unusual behavior with the manifest: https://zavicajna.digitalna.rs/iiif/iiif/api/presentation/2/4aa44ad1-0b74-4590-ab09-534a38cb7c53%252F00000001%252Fostalo01%252F00000012/manifest 

Should those be coming in as large as they are ? Or should we handle it better elsewhere?

Screen Shot 2021-04-01 at 7 07 56 AM

mejackreed (dismissed):
++ Thanks @markmatney for taking this on.

mejackreed:
Ah looks like we have 1 failing test still @markmatney

markmatney:
Can that test be removed? According to `git grep -A 10 "IIIFThumbnail" src/components`, the `IIIFThumbnail` component is never instantiated without at least one of `maxWidth`, `maxHeight`.

mejackreed:
Sure 🤷

regisrob:
> I'm curious too from others including @cbeer and @regisrob who have worked in these areas before.

I am completely behind on this, but I just gave it a try and I can observe that performance on Gallica thumbnails is way better now (indeed Gallica manifests only have a static thumbnail for each canvas, but the server is optimized for that one and the response is a lot faster than a IIIF request on `full/,120` or any other size). Thanks a lot for this @markmatney!

PR #3399, opened 2021-03-12
Add children rendering to PrimaryWindow

mejackreed:
This enhances the `PrimaryWindow` API for plugins to be able to customize behavior when wrapping.

For example, for a 3D viewer plugin, this allows the following:

```javascript
  render() {
    const { TargetComponent, targetProps, threeDResources } = this.props;
    return (
      <TargetComponent {...targetProps}>
        { threeDResources && threeDResources.length > 0 && this.render3DViewer() }
      </TargetComponent>
    )
  }
```

Other implementations could support a map/geo viewer or really anything here is a nice tap in point.

The change here also prevents `this.renderViewer()` from being called if `children` prop is provided. I implemented something similar with a different prop argument name (which totally works), but `children` seemed like a more "React" way of doing things. However, I wonder if this could be problematic? It feels clean, but would be curious about other's thoughts here.

### DIFF
--- a/__tests__/src/components/PrimaryWindow.test.js
+++ b/__tests__/src/components/PrimaryWindow.test.js
@@ -20,6 +20,13 @@ describe('PrimaryWindow', () => {
     const wrapper = createWrapper();
     expect(wrapper.find('.mirador-primary-window')).toHaveLength(1);
   });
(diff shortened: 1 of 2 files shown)


jkeck (approved):
LGTM 👍

PR #3402, opened 2021-03-26
Fix #3401 - expect annotations without target selector

### LINKED ISSUE #3401: Annotations without neither SvgSelector nor xywh fragments cause error
Text annotations that lack any fragment selector, SVG selector or xywh URI fragment cause the error `fragment is null` at this line:
https://github.com/ProjectMirador/mirador/blob/5ca33205bf9bac636ba5ef0faf820f58c0d9751d/src/lib/CanvasAnnotationDisplay.js#L108

The error can be easily reproduced by saving a new annotation without selecting a target in the [annotation plugin demo](https://mirador-annotations.netlify.app/).

This is caused by a missing check in `CanvasAnnotationDisplay.toContext()` - if an annotation does not contain a SvgSelector, it's assumed that a fragment selector is present.

https://github.com/ProjectMirador/mirador/blob/5ca33205bf9bac636ba5ef0faf820f58c0d9751d/src/lib/CanvasAnnotationDisplay.js#L19-L26

I see two possibilities to fix this:
* just add a condition in `toContext()` - annotations that address the whole Canvas (or only have unknown selectors) will not be displayed on the Canvas
* assume a Canvas sized fragment selector if no (known) selector is present - said annotations will appear as a box around the Canvas

Personally I think I'd prefer variant 1.

lutzhelm:


lutzhelm:
I don't know why linting suddenly fails - neither `/home/runner/work/mirador/mirador/__tests__/src/extend/withPlugins.test.js` nor the eslint config have been recently changed.

Should I provide another PR, either with `eslint-disable-line` or with an actual destructuring assignment?

    > eslint ./ && npm run lint:translations && npm run lint:containers
    
    /home/runner/work/mirador/mirador/__tests__/src/extend/withPlugins.test.js
    Error:   129:35  error  Must use destructuring plugin assignment  react/destructuring-assignment

mejackreed:
Feel free to add it onto here as a separate commit if that is easier 👍 .. Probably an eslint version bump that is being pulled in via CI.

lutzhelm:
CI passes now, but the problem will occur for new commits to other existing PRs as well.

mejackreed:
++ Thanks @lutzhelm

PR #3415, opened 2021-04-09
Add npm script for running the test suite in a debugger

markmatney:
I found this to be pretty indispensable when I was modifying the test suite, so I figure others might find it useful too. Lifted from https://artsy.github.io/blog/2018/08/24/How-to-debug-jest-tests/.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,20 @@ $ npm run lint
 ```
 
 ## Debugging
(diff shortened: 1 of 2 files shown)


camillevilla (approved):
Looks useful to me, @markmatney. Thanks for contributing.

PR #3418, opened 2021-04-09
Bump some more dev deps

cbeer:


src/components/ManifestForm.js
```diff
@@ -80,16 +67,19 @@ export class ManifestForm extends Component {
...
     } = this.props;
+    if (!addResourcesOpen) return null;
+
     return (
       <form onSubmit={this.formSubmit}>
         <Grid container spacing={2}>
           <Grid item xs={12} sm={8} md={9}>
             <TextField
-              inputRef={this.inputRef}
+              autoFocus
```
mjgiarlo, line 82: The changes in this file do not look like they're obviously related to a dev dep bump. Can you say more about these?

mjgiarlo (commented):
Have a question about changes to the ManifestForm component.

src/components/ManifestForm.js
cbeer: IIRC, it was motivated by one of the dependency updates breaking a test for this behavior, but MUI's  `autoFocus` provides the behavior we want so we can get out of the whole business by using the upstream code.

mjgiarlo (approved):

camillevilla (approved):

PR #3419, opened 2021-04-09
Afford Windows development

ghost:
Use dev dependencies for cross-platform npm scripts.


charbugs:
Hi @wwelling 

thank you for your contribution :+1:  I'm actually surprised that we did not recognize this lack so far. 

Would you mind to give [shelljs](https://www.npmjs.com/package/shelljs) a try? It's a library and command line tool that mimics basic shell commands with Node. A script in the package.json would then look like:

```
"clean": "shx rm -rf ./dist",
```

This way we would add only one dependency instead of 4 which may be a little easier to maintain.

wwelling:
@charbugs sorry for the very delayed response. I have looked into using shelljs (shx) but it does not have the support to set environment variables as needed. I will update to replace what dependencies I can with shelljs but cross-env is still required.

https://github.com/shelljs/shx/issues/166
https://github.com/shelljs/shx/issues/127

PR #3432, opened 2021-05-03
Prevent item labels in Collection modal from being cut off

### LINKED ISSUE #3424: Item labels are cut off in the Collection modal
Labels that are larger than the width of the Collection modal do not wrap to the next line (`nowrap` behavior) and are then cut off.

Example Manifest: https://portail.biblissima.fr/iiif/collection/ark:/43093/coldata5151005ea5833e5a05e2639cbb210946cb7e0609

Capture d’écran 2021-04-15 à 16 58 36

What is the best approach to solve that? 
- an inline `style={{whiteSpace: 'normal'}}` on `MenuItem` in [CollectionDialog.js#L246](https://github.com/ProjectMirador/mirador/blob/master/src/components/CollectionDialog.js#L246)?
- or a new `className`?

regisrob:
Fixes #3424 

Example Manifest: https://portail.biblissima.fr/iiif/collection/ark:/43093/coldata5151005ea5833e5a05e2639cbb210946cb7e0609

Screen Shot 2021-05-12 at 07 47 48

### DIFF
--- a/src/components/CollectionDialog.js
+++ b/src/components/CollectionDialog.js
@@ -243,7 +243,11 @@ export class CollectionDialog extends Component {
             <MenuList>
               {
                 collections.map(c => (
-                  <MenuItem key={c.id} onClick={() => { this.selectCollection(c);}} className={classes.collectionItem}>
+                  <MenuItem
+                    key={c.id}
+                    onClick={() => { this.selectCollection(c); }}
+                    className={classes.collectionItem}
+                  >
                     {CollectionDialog.getUseableLabel(c)}
                   </MenuItem>
                 ))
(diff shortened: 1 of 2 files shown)


cbeer:
I've added a screenshot above. @ggeisler, does this look ok?

ggeisler:
@regisrob @cbeer Yes, this looks great to me.

That screenshot illustrates another minor styling issue: the main content area (from the Resource heading down) needs to be indented an additional `10px` or so so that it is left-aligned with the header content. But that is a different issue so I can create a separate ticket for that if we want to keep this one simple and get it merged.

cbeer (approved):

Issue #3420, opened 2021-04-12
Add support for image annotation


MImranAsghar:
Hi, I am currently looking into updating mirador to allow for image annotations. This is what i have been working on and believe would be needed for allowing image annotations.

1). Updated mirador-annotations plugin to add insert image dialog: This image insert/edit input dialog is needed to extend the annotation support to adding image annotation.
- These image annotation insert/edit changes attempt to implement image insert/edit workflow similar to Mirador 2 for the image annotation. For example, once a url is entered then clicking away from the url field will update the width and height inputs with the image width and height that the url corresponds to.

Made a pull request for this in mirador-annotations repo: https://github.com/ProjectMirador/mirador-annotations/pull/50

2). Update mirador to add image annotation viewing on a card when the annotation is hovered, The other part of image annotations is the image annotation card (list of image annotations displayed on a card on hovered annotations) which I have yet to make a pull request for that can be added to mirador 3.

  a). On the canvas both text and image annotations will be shown on the card for all the annotations that are hovered.
  b). User can scroll horizontally and vertically on the card to figure see the entire image.
  c). If the user hovers over the "annotation" item on the side bar menu, then the card will be shown on the canvas over that specific annotation in the center of the drawn annotation.

Would like to double check that i am on the right path. Thanks.

Here is how the input dialog looks (see pull request https://github.com/ProjectMirador/mirador-annotations/pull/50):

<img width="1440" alt="Screen Shot 2021-04-12 at 6 19 29 PM" src="https://user-images.githubusercontent.com/12305084/114470084-447afd80-9bbc-11eb-97d1-b350b807238f.png">

Here is how the card looks:

<img width="1440" alt="Screen Shot 2021-04-12 at 5 52 53 PM" src="https://user-images.githubusercontent.com/12305084/114468512-9ec68f00-9bb9-11eb-945a-1ade4405ae09.png">

Manifest URLs: Using this sample manifest for testing => https://iiif.harvardartmuseums.org/manifests/object/299843

ggeisler:
@MImranAsghar Thanks for making the updates to the modal window; it looks great now. One thought: I wonder if the title of the modal should be "Add image" rather than "Insert image"? The word "insert" feels a little too specific or technical, and the action the user is doing, I think, is simply adding the image as an annotation.

[...]

PR #3435, opened 2021-05-13
Add annot image display through a dialog and annot list in sidebar

MImranAsghar:
- Display an image annotation in sidebar Canvas Annotations
- Add annotation image dialog to display full annotation image (full size with scrolling)

Screen Shot 2021-05-13 at 10 48 30 AM
Screen Shot 2021-05-13 at 10 48 38 AM


ggeisler:
Hi @MImranAsghar. This looks like nice progress on this feature, thanks for continuing to work on it. Your screenshots are helpful but is there a way you can also enable others to actually try out the feature in this PR? 

The Netlify deploy-preview link doesn't appear to resolve to a preview that includes the annotation creation plugin so there isn't a way to try creating our own image annotation.

Or, if this PR is just focused on displaying an existing image annotation, there still isn't a way to see an image annotation in the deploy preview. Maybe you need to provide an example manifest that can be added to show something like your screenshots?

MImranAsghar:
Hi! This PR is focused on displaying the created annotations. Hmm, not sure how to make it easy to view other than deploying on cloud... but I was thinking maybe you can try it out locally? I was doing that when developing.

To try out locally:
1. Mirador folder needs to be moved under mirador-annotations
2. In `nwb.config.js`, the following line can be added under 'mirador' > 'aliases':
    - 'mirador':  path.resolve('./', 'mirador'),

marlo-longley:
This pr addressed https://github.com/ProjectMirador/mirador/issues/3420 -- good work contained here, needs an updated review and rebase (we should take it over)

marlo-longley:
@dnoneill this is related to your current work in #4446

dnoneill:
Tangentially but not really. There is an image field in the annotation spec and this is that.

PR #3443, opened 2021-06-11
Wrap the export json in an accordion; fixes #3440

### LINKED ISSUE #3440: Update Export workspace modal to hide JSON in an expand/collapse section
On a IIIF call I heard the Export feature referred to as "clunky" and I believe the display of the configuration JSON was at least part of what led to that impression (because using the Export feature is simply a matter of selecting the Copy button, which doesn't seem very clunky to me). While I don't think we want to get rid of the JSON in this modal (it can be handy to copy and paste sections from, for advanced users at least), perhaps we can make the feature feel simpler but putting the JSON in an expand/collapse section, and hide it in the initial state.

For example, this could be how the modal looks when initially displayed:

Screen Shot 2021-05-26 at 2 44 31 PM

---

This would involve:

- Put the existing JSON code section into an expand/collapse (similar to what we do in the Window sidebar panel, About section) with the text label `View workspace configuration` (as an `<h3>` but with the `MuiTypography-h4` class to style it as an `<h4>`)
- Update the label of the primary button to `Copy configuration` , which is long and sort of awkward but does make it more clear what you are copying, since you won't initially see the JSON

When the collapsed section is expanded:

Screen Shot 2021-05-26 at 2 47 52 PM

---

- The JSON section is displayed as it currently does, with the toggle icon updated to reflect the current expanded state

cbeer:
fixes #3440
Kapture 2021-06-11 at 14 21 47

### DIFF
--- a/src/components/WorkspaceExport.js
+++ b/src/components/WorkspaceExport.js
@@ -3,13 +3,17 @@ import Button from '@material-ui/core/Button';
 import Dialog from '@material-ui/core/Dialog';
 import DialogActions from '@material-ui/core/DialogActions';
 import DialogTitle from '@material-ui/core/DialogTitle';
+import DialogContent from '@material-ui/core/DialogContent';
 import Typography from '@material-ui/core/Typography';
 import Snackbar from '@material-ui/core/Snackbar';
 import IconButton from '@material-ui/core/IconButton';
+import ExpandMoreIcon from '@material-ui/icons/ExpandMore';
 import CloseIcon from '@material-ui/icons/Close';
+import Accordion from '@material-ui/core/Accordion';
+import AccordionSummary from '@material-ui/core/AccordionSummary';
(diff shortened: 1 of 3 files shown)


src/components/WorkspaceExport.js
```diff
@@ -87,12 +91,24 @@ export class WorkspaceExport extends Component {
...
-          </pre>
-        </ScrollIndicatedDialogContent>
+
+        <DialogContent>
+          <Accordion elevation={0}>
+            <AccordionSummary
+              classes={{ root: classes.accordionTitle }}
+              expandIcon={<ExpandMoreIcon />}
+            >
+              <Typography variant="h4">View workspace configuration</Typography>
```
camillevilla, line 101: Can you move this over to the locales file?

camillevilla (changes requested):

camillevilla (approved):

PR #3446, opened 2021-06-11
Pluralize item counts for English, at least; fixes #3268

### LINKED ISSUE #3268: Singularize item counts with only 1 item
In the catalog view we show item counts for each catalog resource. It appears that we always use "items" (plural) in the item count string. Is there an equivalent to the Rails `pluralize` method that we could apply here?

Screen Shot 2020-09-25 at 4 55 24 PM

cbeer:
fixes #3268
Screen Shot 2021-06-11 at 15 04 58


camillevilla (approved):

camillevilla:
🤔 I wonder why we weren't using react-i18next plurals before...
https://www.i18next.com/translation-function/plurals#singular-plural

PR #3474, opened 2021-07-08
Norwegian bokmål translation

tarjelavik:
Translation to norwegian "bokmål".


cbeer:
Thanks for this contribution. Could you also add Norwegian to the `availableLanguages ` configuration (https://github.com/ProjectMirador/mirador/blob/master/src/config/settings.js#L219-L234) so it is available by default?

tarjelavik:
Yes, i will do that :-)

cbeer:
Thanks!

PR #3476, opened 2021-07-08
Fix warnings related to Material UI updates

dicksonlaw583:
This PR resolves several errors that come up regarding properties renamed in Material UI 5:

```
Material-UI: the createMuiTheme function was renamed to createTheme.

You should use `import { createTheme } from '@material-ui/core/styles'`
```

```
Warning: Failed prop type: The prop `onEntering` of `ForwardRef(Menu)` is deprecated. Use the `TransitionProps` prop instead.

Warning: Failed prop type: The prop `onExit` of `ForwardRef(Menu)` is deprecated. Use the `TransitionProps` prop instead.
```

```
Warning: Failed prop type: Material-UI: variant="round" was renamed variant="circular" for consistency.
```

```
Material-UI: The `fade` color utility was renamed to `alpha` to better describe its functionality.
```

```
Warning: Failed prop type: The prop `justify` of `ForwardRef(Grid)` is deprecated. Use `justifyContent` instead, the prop was renamed.
```

### DIFF
--- a/__tests__/src/components/WindowTopMenu.test.js
+++ b/__tests__/src/components/WindowTopMenu.test.js
@@ -42,8 +42,10 @@ describe('WindowTopMenu', () => {
     expect(wrapper.find(Menu).first().props().anchorEl).toBe(null);
     expect(wrapper.find(Menu).first().props().open).toBe(false);
     expect(wrapper.find(Menu).first().props().onClose).toBe(handleClose);
-    expect(wrapper.find(Menu).first().props().onEntering).toBe(toggleDraggingEnabled);
-    expect(wrapper.find(Menu).first().props().onExit).toBe(toggleDraggingEnabled);
+    expect(wrapper.find(Menu).first().props().TransitionProps.onEntering)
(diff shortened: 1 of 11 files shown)


cbeer:
Do you know whether these are backwards compatible with MUI 4.11, or do we need to require MUI 4.12+?

dicksonlaw583:
In `package.json`, the Material UI dependency is stated as follows:

```
"@material-ui/core": "^4.11.0"
```

Any new installation of Mirador 3 today will grab 4.12.1, which will cause the listed errors to come up unless the changes I described are made. Because some of these are recent renames, they are not backwards-compatible with 4.11.x versions.

I tried changing the dependency statement to this:

```
"@material-ui/core": "~4.11.0"
```

But `npm install` was unable to resolve the dependencies that way. So at this point, I think we do need to require Material UI 4.12.

ewlarson:
I'm now seeing these warnings, too. Anything holding back this PR?

thatbudakguy (approved):
@cbeer OK to merge this?

PR #3477, opened 2021-07-13
do not ignore the package-lock.json by .gitignore

charbugs:
https://dev.to/adamklein/package-lock-json-in-git-or-not-50l5


cbeer:
I understand why you'd want to commit a `package.lock` in an application, but I guess I haven't traditionally viewed mirador as an app itself. If we commit a `package.lock`, are we no longer testing against the dependency versions that a downstream consumer might be using?

charbugs:
Oh, ok! I never thought about the app/lib difference when it comes to package-lock. But I guess it makes sense. Thanks for the hint.

PR #3530, opened 2022-01-11
Scroll the selected image to view in Gallery

priyamanalil:
When gallery view opens in mirador, it always displays images from the beginning of the sequence. This is not user friendly when a user switches from single or book view to gallery view, especially when the current image is not at the beginning of the sequence. Requires a lot of scrolling from the user.  So the aim is to keep the selected image in focus and in view while switching between the views.

For this, we modify `GalleryViewThumbnail` to scroll itself into view if it is marked as selected when it's mounted.


camillevilla (approved):
Thanks for contributing, @priyamanalil. This is a nice improvement to the user experience.

PR #3536, opened 2022-02-16
Add "related" to the getManifestHomepage selector, fix #3522

### LINKED ISSUE #3522: ManifestRelatedLinks lacks support for Presentation API 2 property "related"
Mirador 3 does currently not display any information within the [`related`](https://iiif.io/api/presentation/2.1/#related) property of a manifest, like given in the following example: https://iiif.ub.uni-leipzig.de/0000033918/manifest.json

I think they should be display in the same way as `homepage` URLs.

`related` in v2 might either be one or more [URLs or objects](https://iiif.io/api/presentation/2.1/#uri-representation) with an `@id` property and additional `label` and/or `format` properties.

lutzhelm:
Fixes #3522. This is an alternative to #3535. It reuses the `getManifestHomepage` selector for the v2 `related` property. While it is a much simpler approach, it looks like the less clean alternative to me.

### DIFF
--- a/__tests__/fixtures/version-2/related-objects.json
+++ b/__tests__/fixtures/version-2/related-objects.json
@@ -1,0 +1,14 @@
+{
+  "@context": "http://iiif.io/api/presentation/2/context.json",
+  "@id": "http://example.com/iiif/manifest/related-objects.json",
+  "@type": "sc:Manifest",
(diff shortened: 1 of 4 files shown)


lutzhelm:
Closed in favor of #3535.

PR #3611, opened 2023-01-23
Upgrade normalize-url

jcoyne:


cbeer:
Does this work with Safari? The upstream readme still says:
> If you need Safari support, use version 4: npm i normalize-url@4

jcoyne:
This needs to stay on 4.0 for safari: https://github.com/sindresorhus/normalize-url/pull/170#issuecomment-1234134039

jcoyne:
@cbeer yep, I was just looking at that.

PR #3614, opened 2023-01-24
Use the URL api rather than normalize-url

jcoyne:
The URL API is now in all browsers that we support: https://caniuse.com/url

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,7 +24,6 @@
         "jss-rtl": "^0.3.0",
         "lodash": "^4.17.11",
         "manifesto.js": "^4.2.0",
-        "normalize-url": "^4.5.0",
         "openseadragon": "^2.4.2",
         "prop-types": "^15.6.2",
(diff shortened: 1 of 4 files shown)


cbeer:
I think `normalize-url` provides additional functionality to deal with slightly malformed IIIF manifest data, doesn't it? I don't think it's completely equivalent to URL.

jcoyne:
@cbeer they are not equivalent, but I think that URI handles the cases we actually care about.  We do have some bad tests though that rely on this difference (https://github.com/ProjectMirador/mirador/pull/3620 for example)

marlo-longley:
Closing this as stale — it's now conflicting with main, normalize-url has moved from v4 to v9 since this was opened, and @cbeer's review concern about URL vs. normalize-url handling malformed IIIF ids differently was never resolved. If we want to investigate dropping the dependency we can investigate in a new ticket.

Issue #3592, opened 2023-01-13
Upgrade drag+drop libraries for compatibility with React 17+


cbeer:
Mirador uses react-dnd directly, and also through other packages:
- [ ] react-mosaic-component 5.x
- [ ] react-dnd
- [ ] react-dnd-multi-backend

TODO:
- [ ] update react-mosaic-component to support/allow react-dnd-multi-backend 8.x

cbeer:
Fixed by https://github.com/ProjectMirador/mirador/pull/3606.

PR #3606, opened 2023-01-13
Update react-mosaic, react-dnd and related libraries

cbeer:

PR #3625, opened 2023-01-25
Upgrade to react-mosaic-component 5.0

jcoyne:
Ref #3592


cbeer:
Does this bring in 2 separate copies of react-dnd etc?

jcoyne:
@cbeer I made it slightly better, but react-dnd-multi-backend and react-mosaic-component don't agree on which versions of dependencies (react-dnd, react-dnd-html5-backend, or react-dnd-touch-backend) should be used.

cbeer:
Replaced by #3606

PR #3626, opened 2023-01-25
Revert "Add the lockfile to the repo"

cbeer:
This reverts commit 24bf8c5b14d8182cc40c10566558c087341dce1d.

I think such a major change to the developer experience probably deserves discussion at a committer call to make sure we have consensus about tooling. 

Looking around at some of our dependencies, I'm not seeing that this is a widely adopted practice, and sure makes a lot of merge conflicts.


jcoyne:
@cbeer without this I'm completely stuck as doing a general update of dependencies breaks the build.  See https://github.com/ProjectMirador/mirador/pull/3617 for example.

.github/workflows/node.js.yml
```diff
@@ -22,7 +22,12 @@ jobs:
       uses: actions/setup-node@v2
       with:
         node-version: ${{ matrix.node-version }}
-    - run: npm ci
+    - run: npm install
+    - name: Archive package.lock
+      uses: actions/upload-artifact@v3
+      with:
+        name: package-lock.json
```
jcoyne, line 29: While one could use this approach to debug, it still means you have to debug after it breaks, rather than have it not break unless you push an update.

jcoyne:
> Looking around at some of our dependencies, I'm not seeing that this is a widely adopted practice, and sure makes a lot of merge conflicts.

Bootstrap checks theirs in https://github.com/twbs/bootstrap/blob/main/package-lock.json

package.json
```diff
@@ -130,5 +130,8 @@
   "peerDependencies": {
     "react": "^16.8.3",
     "react-dom": "^16.8.3"
+  },
+  "overrides": {
+    "jest-environment-node": "29.4.0"
```
jcoyne, line 135: How did you find this?

jcoyne (changes requested):

package.json
cbeer: https://github.com/smooth-code/jest-puppeteer/issues/512, where there's an explanation about why they can't do it in their dependencies... but jest fixed things up stream so we don't need it any more either.

jcoyne (approved):
I don't want to be a blocker, but it seems like a good idea to have a success lock state checked in like the other projects.

thatbudakguy:
I'm gonna go ahead and merge this since I think @cbeer is right that the decision to commit the lockfile ultimately deserves more discussion. I merged the initial PR to commit it because it unblocked @jcoyne finding out why CI was failing, but it sounds like having it committed is now functionally blocking other work because of merge conflicts.

My knowledge matches @jcoyne's re: it seems like a somewhat widely adopted practice to commit the lockfile (and yarn's docs explicitly recommend this) but for this project it seems like more discussion is needed to implement that with a minimum of friction.

PR #3634, opened 2023-02-08
Update puppeteer

cbeer:


jcoyne (dismissed):

jcoyne (approved):

cbeer:
jest, node, and react 16 don't play nice together (hence `--forceExit`). We could use https://www.npmjs.com/package/react-16-node-hanging-test-fix, but hopefully we'll be off React 16 soon enough that it won't matter 🤷

Issue #3325, opened 2020-10-29
Change cursor icon when user's cursor is over canvas image


ggeisler:
From usability study.

Use `grab` icon when cursor is over image.

Change to `grabbing` if they start to move the image.

https://developer.mozilla.org/en-US/docs/Web/CSS/cursor

willmartindev:
Hi! Could I make a pull request for this issue?

PR #3644, opened 2023-02-22
Change cursor icon when user's cursor is over canvas image

phil-plencner-hl:
This PR solves this issue: https://github.com/ProjectMirador/mirador/issues/3325

This PR adds viewer handlers for the [canvas-press](https://openseadragon.github.io/docs/OpenSeadragon.Viewer.html#.event:canvas-press) and [canvas-release](https://openseadragon.github.io/docs/OpenSeadragon.Viewer.html#.event:canvas-release) OpenSeadragon events then adds that value to the state. 

The state determines which [CSS cursor icon](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor) is used (`grab` or `grabbing`) in the style attribute of the `<section>` tag.

### DIFF
--- a/src/components/OpenSeadragonViewer.js
+++ b/src/components/OpenSeadragonViewer.js
@@ -23,7 +23,7 @@ export class OpenSeadragonViewer extends Component {
   constructor(props) {
     super(props);
 
-    this.state = { viewer: undefined };
+    this.state = { grabbing: false, viewer: undefined };
     this.ref = createRef();
     this.apiRef = createRef();
     OSDReferences.set(props.windowId, this.apiRef);
@@ -57,6 +57,15 @@ export class OpenSeadragonViewer extends Component {
     this.apiRef.current = viewer;
 
     this.setState({ viewer });
+
+    viewer.addHandler('canvas-press', () => {
+      this.setState({ grabbing: true });
+    });
+
+    viewer.addHandler('canvas-release', () => {
+      console.log('canvas-release!');
+      this.setState({ grabbing: false });
+    });
 
     // Set a flag when OSD starts animating (so that viewer updates are not used)
     viewer.addHandler('animation-start', () => {
@@ -345,7 +354,7 @@ export class OpenSeadragonViewer extends Component {
       children, classes, label, t, windowId,
       drawAnnotations,
     } = this.props;
-    const { viewer } = this.state;
+    const { viewer, grabbing } = this.state;
 
     const enhancedChildren = Children.map(children, child => (
       cloneElement(
@@ -359,6 +368,7 @@ export class OpenSeadragonViewer extends Component {
     return (
       <section
         className={classNames(ns('osd-container'), classes.osdContainer)}
+        style={{ cursor: grabbing ? 'grabbing' : 'grab' }}
         id={`${windowId}-osd`}
         ref={this.ref}
         aria-label={t('item', { label })}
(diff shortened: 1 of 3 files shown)


src/components/OpenSeadragonViewer.js
```diff
@@ -58,6 +58,14 @@ export class OpenSeadragonViewer extends Component {
 
     this.setState({ viewer });
 
+    viewer.addHandler('canvas-press', () => {
```
cbeer, line 61: would `canvas-drag` and `canvas-drag-end` be more appropriate events?

src/components/OpenSeadragonViewer.js
```diff
@@ -359,6 +367,7 @@ export class OpenSeadragonViewer extends Component {
     return (
       <section
         className={classNames(ns('osd-container'), classes.osdContainer)}
+        style={{ cursor: grabbing ? 'grabbing' : 'grab' }}
```
cbeer, line 370: Could we push the `grab` default up to the osdContainer styles?

src/components/OpenSeadragonViewer.js
phil-plencner-hl: @cbeer: Seems to work the same for me in my browsers. Is that better for touch screen devices? If so, I can commit the change.

cbeer:
I'm seeing something funny: it looks like the grab cursor shows up in the canvas navigation pane too:

Screen Shot 2023-02-22 at 12 49 32

I think we need to turn the cursor back to its default there, because the canvas isn't draggable at that point.

src/components/OpenSeadragonViewer.js
phil-plencner-hl: Done with this commit: https://github.com/ProjectMirador/mirador/pull/3644/commits/b02ff305ba68c466c91f73e294345754b0e26e53

src/containers/OpenSeadragonViewer.js
```diff
@@ -58,6 +58,7 @@ const mapDispatchToProps = {
 const styles = {
   osdContainer: {
     flex: 1,
+    pointer: 'grab',
```
marlo-longley, line 61: do we want to use the 'cursor' selector?

phil-plencner-hl: Whoops!
https://github.com/ProjectMirador/mirador/pull/3644/commits/4d79018b9434b78c3828248527f143277b986947

src/components/OpenSeadragonViewer.js
phil-plencner-hl: And this commit:
https://github.com/ProjectMirador/mirador/pull/3644/commits/4d79018b9434b78c3828248527f143277b986947

src/containers/OpenSeadragonViewer.js
phil-plencner-hl: navigation pane fixed with this commit:
https://github.com/ProjectMirador/mirador/pull/3644/commits/0fc4c3b2198c9ee987a474ae97fb150c2ae89915

src/components/OpenSeadragonViewer.js
phil-plencner-hl: This commit changes it to use `canvas-drag` and `canvas-drag-end`:
https://github.com/ProjectMirador/mirador/pull/3644/commits/fd2681febf0fb3d6c01148cc2a271b740b4a6aa4

src/components/OpenSeadragonViewer.js
cbeer, line 370: Now that the `grab` style is upstream, we can just omit it here:

```suggestion
        style={{ cursor: grabbing ? 'grabbing' : undefined }}
```

phil-plencner-hl: Done!
https://github.com/ProjectMirador/mirador/pull/3644/commits/891cc827d949ec7a3ea5d9a7f5559424726a281f

Issue #2233, opened 2019-03-18
Couple of tooltip display issues


jvine:
1. Tooltips for the expand/collapse toggle appear in the upper left of the window rather than associated with the icon.
<img width="390" alt="Screen Shot 2019-03-18 at 1 40 02 PM" src="https://user-images.githubusercontent.com/6945691/54562080-7391f700-4983-11e9-91cd-0ac538c076c4.png">
<img width="230" alt="Screen Shot 2019-03-18 at 1 41 40 PM" src="https://user-images.githubusercontent.com/6945691/54562137-94f2e300-4983-11e9-9d5f-7053050d064d.png">

2. The "Exit full screen" tooltip seems to be associated with the "Full screen" icon. I can't capture this in a gif. Try this:
- hover over Full screen icon - "Full screen" tooltip appears
- click Full screen icon - tooltip quickly changes to "Exit full screen" then disappears
- hover over Exit full screen icon - no tooltip

jkeck:
FWIW, I believe there is an issue w/ all tooltips in Fullscreen mode.

jvine:
Ah totally right, they are all missing. I was distracted by the ghost of the text appearing briefly.

PR #3645, opened 2023-02-23
Use context to manage access to the workspace ref instead of going through redux

cbeer:
We're passing the container id around in order to provide the right context for tooltips, dialogs, etc (e.g. #2233). Storing this in redux has been... fine, but it's going to be a pain with react-testing-library because we'll have to mock out that state and deal with more connected components than we'd otherwise have to.

The new workspace context gets us out of that, means we're passing around refs instead of DOM IDs (probably a net good?), and lets us fall back on a reasonable default if there's no context available.

### DIFF
--- a/src/containers/WorkspaceOptionsMenu.js
+++ b/src/containers/WorkspaceOptionsMenu.js
@@ -1,18 +1,11 @@
 import { compose } from 'redux';
-import { connect } from 'react-redux';
 import { withTranslation } from 'react-i18next';
-import { getContainerId } from '../state/selectors';
+import { withWorkspaceContext } from '../contexts/WorkspaceContext';
 import { WorkspaceOptionsMenu } from '../components/WorkspaceOptionsMenu';
 
-/** Used for connect */
-const mapStateToProps = state => ({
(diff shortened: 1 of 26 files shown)


src/containers/WorkspaceOptionsMenu.js
```diff
@@ -1,18 +1,11 @@
...
-import { getContainerId } from '../state/selectors';
+import { withWorkspaceContext } from '../contexts/WorkspaceContext';
 import { WorkspaceOptionsMenu } from '../components/WorkspaceOptionsMenu';
 
-/** Used for connect */
-const mapStateToProps = state => ({
-  containerId: getContainerId(state),
-});
-
-// containerId: getContainerId(state),/
```
marlo-longley, line 12: Should this be deleted?

cbeer: Yes!

cbeer: Wait, yes, it is deleted.

marlo-longley (approved):

marlo-longley (approved):

Issue #3405, opened 2021-04-01
Support `logo` image within `provider` in IIIF Presentation v3 manifest


triplingual:
- I am the author or maintainer of this manifest: Yes
- ~URL: https://preview.iiif.io/cookbook/0234-provider/recipe/0234-provider/manifest.json~
- URL: https://iiif.io/api/cookbook/recipe/0234-provider/manifest.json
- Manifest hosting institution: IIIF

### Observed behavior

- Description: The Manifest itself is loading, but the image in the `logo` in the `provider` property is not. I even set the `logo` content to be the known-working JSON from issue #3269 and nothing.
- 
![Screen Shot 2021-04-01 at 12 49 28](https://user-images.githubusercontent.com/1789625/113327293-c62f7900-92e8-11eb-982b-4aecd0bb21f5.png)

- No pertinent errors or warnings from the UI, developer console, or the [IIIF Presentation Validator](https://iiif.io/api/presentation/validator/service/) that I can see, but I could be missing something or ignorant about this.

triplingual:
(Noting that my only edit to the comment just now was to change the manifest URI to the more nearly permanent one now that the recipe is formally published in the cookbook.)

marlo-longley:
Closed by #3657

PR #3657, opened 2023-02-24
Handle v3 Provider logos as thumbnails

marlo-longley:
I also changed the provider test fixture because it seemed improperly formed to me -- we had objects instead of arrays as the values of the keys, i.e. `logo: {}` rather than `logo: [{}]` which is described [in the spec](https://iiif.io/api/presentation/3.0/#provider).

### DIFF
--- a/src/state/selectors/manifests.js
+++ b/src/state/selectors/manifests.js
@@ -1,6 +1,6 @@
 import { createSelector } from 'reselect';
 import createCachedSelector from 're-reselect';
-import { PropertyValue, Utils } from 'manifesto.js';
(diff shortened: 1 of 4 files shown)


src/state/selectors/manifests.js
```diff
@@ -109,6 +104,47 @@ export const getManifestProvider = createSelector(
     && PropertyValue.parse(provider[0].label, locale).getValue(),
 );
 
+/** */
+function providerLogo(provider) {
```
marlo-longley, line 108: CodeCov wanted me to make these smaller methods

marlo-longley:
Closes https://github.com/ProjectMirador/mirador/issues/3405

PR #3661, opened 2023-02-27
Use context to pass the window id around

cbeer:
All the prop drilling makes it a little harder to test. Maybe context makes it easier?


marlo-longley:
This is out of date but we can track the idea as an architectural improvement -- dev can review the idea and re-PR if decided.

PR #3664, opened 2023-02-27
Get vtt captions from v3 annotations

marlo-longley:
@phil-plencner-hl, this would merge into your branch, and then we can proceed with your original branch and PR that ?

### DIFF
--- a/__tests__/src/selectors/canvases.test.js
+++ b/__tests__/src/selectors/canvases.test.js
@@ -4,6 +4,7 @@ import minimumRequired from '../../fixtures/version-2/minimumRequired.json';
 import minimumRequired3 from '../../fixtures/version-3/minimumRequired.json';
 import audioFixture from '../../fixtures/version-3/0002-mvm-audio.json';
 import videoFixture from '../../fixtures/version-3/0015-start.json';
+import videoWithAnnoCaptions from '../../fixtures/version-3/video_with_annotation_captions.json';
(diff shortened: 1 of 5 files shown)


__tests__/src/selectors/canvases.test.js
```diff
@@ -400,7 +401,7 @@ describe('getVisibleCanvasNonTiledResources', () => {
   });
 
   describe('getVisibleCanvasCaptions', () => {
-    it('returns canvases resources', () => {
+    it('returns v2 canvases resources', () => {
```
marlo-longley, line 404: I am actually confused about this, because the `videoFixture` (from `import videoFixture from '../../fixtures/version-3/0015-start.json'`) is supposedly a IIIF v3 resource, but it has its captions in the old v2 style.

phil-plencner-hl (changes requested):
This looks good to me and works for me as well.
Do we need to merge into the `issue-3538` branch or can we merge it right to `master`? There's nothing we need to keep from `issue-3538`.

Also we should revert back the video.html page to the state in the `master` branch. I only updated the page temporarily to see the two types of videos side-by-side for testing purposes.

phil-plencner-hl (approved):
This looks good. The original video.html page is back and the previously broken video with the captions plays properly with the captions.
I also ran the tests and they passed.

phil-plencner-hl (approved):
Works good.

PR #3685, opened 2023-03-03
Passing ViewerNavigation Tests.

phil-plencner-hl:
PR for ViewerNavigation.test.js

### DIFF
--- a/__tests__/src/components/ViewerNavigation.test.js
+++ b/__tests__/src/components/ViewerNavigation.test.js
@@ -29,7 +29,7 @@ describe('ViewerNavigation', () => {
       setPreviousCanvas,
     });
     const buttons = screen.queryAllByRole('button');
-    expect(buttons[0].closest('div')).toBeInTheDocument();
+    expect(buttons[0].closest('div')).toBeInTheDocument(); // eslint-disable-line testing-library/no-node-access
   });
   describe('when next canvases are present', () => {
     it('nextCanvas button is not disabled', () => {
@@ -121,7 +121,7 @@ describe('ViewerNavigation', () => {
         viewingDirection: 'right-to-left',
       });
       const buttons = screen.queryAllByRole('button');
-      expect(buttons[0].closest('div')).toHaveAttribute('dir', 'rtl');
+      expect(buttons[0].closest('div')).toHaveAttribute('dir', 'rtl'); // eslint-disable-line testing-library/no-node-access
     });
   });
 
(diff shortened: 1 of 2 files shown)


__tests__/src/components/ViewerNavigation.test.js
```diff
@@ -16,125 +15,138 @@ function createWrapper(props) {
...
         viewingDirection: 'bottom-to-top',
       });
-    });
-
-    it('changes the arrow styles', () => {
-      const previous = wrapper.find(MiradorMenuButton).first().children(NavigationIcon).props();
-      const next = wrapper.find(MiradorMenuButton).last().children(NavigationIcon).props();
-      expect(previous.style).toEqual({ transform: 'rotate(90deg)' });
-      expect(next.style).toEqual({ transform: 'rotate(270deg)' });
+      expect(screen.getByLabelText('previousCanvas', { selector: 'button' }).querySelector('svg')).toHaveStyle('transform: rotate(90deg);'); // eslint-disable-line testing-library/no-node-access
```
cbeer, line 148: I think checking the role is probably preferable to checking by text

```suggestion
      expect(screen.getByRole('button', { name: 'previousCanvas' }).querySelector('svg')).toHaveStyle('transform: rotate(90deg);'); // eslint-disable-line testing-library/no-node-access
```

src/components/ViewerNavigation.js
```diff
@@ -41,6 +41,7 @@ export class ViewerNavigation extends Component {
 
     return (
       <div
+        role="navigation"
```
cbeer, line 44: I'll have to go back and look, but I dimly recall we intentionally decided that a navigation role was inappropriate here.

phil-plencner-hl: Ok. I'm not sure how else to select that item. The first test might not be necessary (just checking for existence of the wrapper `<div>`). However, there's another test that is specifically checking the viewing direction....
https://github.com/ProjectMirador/mirador/pull/3685/files#diff-69ebf93942b44352912effb10c94e7ae4687632918a4a61526ea1e37a9e7dbdcR122

__tests__/src/components/ViewerNavigation.test.js
phil-plencner-hl: @cbeer: I don't see a button role on anything in the debug? Changing the line to what you suggested makes the test fail.

cbeer: Unless the button has an explicit role, it should have the implicit role `button`:

phil-plencner-hl: @cbeer I was able to update them all to use the button role.

src/components/ViewerNavigation.js
phil-plencner-hl: @cbeer I updated the tests to not rely on the navigation role. I removed the navigation role from the component.

https://github.com/ProjectMirador/mirador/pull/3685/commits/967dea90aeb2bfba0d684bb8b3c12e878dde7ebd

PR #3687, opened 2023-03-03
Gallery view test

phil-plencner-hl:
PR for GalleryView.test.js

### DIFF
--- a/__tests__/src/components/GalleryView.test.js
+++ b/__tests__/src/components/GalleryView.test.js
@@ -1,10 +1,8 @@
-import { render, screen } from '@testing-library/react';
-import { renderWithProviders } from '../../utils/store';
+import { screen } from '@testing-library/react';
 import { Utils } from 'manifesto.js';
-import Paper from '@material-ui/core/Paper';
+import { renderWithProviders } from '../../utils/store';
 import manifestJson from '../../fixtures/version-2/019.json';
 import { GalleryView } from '../../../src/components/GalleryView';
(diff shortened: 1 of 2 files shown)


__tests__/src/components/GalleryView.test.js
```diff
@@ -19,28 +18,27 @@ function createWrapper(props) {
...
   beforeEach(() => {
     setCanvas = jest.fn();
-    wrapper = createWrapper({ setCanvas });
   });
   it('renders the component', () => {
-    expect(wrapper.find(Paper).length).toBe(1);
-    expect(wrapper.find(Paper).prop('component')).toEqual('section');
+    createWrapper({ setCanvas });
+    const buttons = screen.queryAllByRole('button');
+    expect(buttons[0].closest('section')).toBeInTheDocument(); // eslint-disable-line testing-library/no-node-access
```
marlo-longley, line 27: This seems a little roundabout -- just means if any ancestor is a `section` ? If the buttons/GalleryViewThumbnails are rendering, then can we just assume the component is rendering? Like maybe the first two `it` blocks end up being the same. Not sure.

cbeer, line 27: Assuming we want to check for a `<section>`, maybe we can just do:

```suggestion
    const { container } = createWrapper({ setCanvas });
    expect(container.querySelector('section')).toBeInTheDocument(); // eslint-disable-line testing-library/no-node-access, testing-library/no-container
```

PR #3696, opened 2023-03-09
Test the thumbnail intersection with RTL

marlo-longley:
@phil-plencner-hl here's what I got to work for that.
I'm curious what everyone thinks about removing the brackets from that argument. It doesn't seem necessary from the library docs  for onChange: https://www.npmjs.com/package/react-intersection-observer

onChange | (inView, entry) => void | undefined | Call this function whenever the in view state changes. It will receive the inView boolean, alongside the current IntersectionObserverEntry.
-- | -- | -- | --

I tested the functionality and it seems to work still. Without that, I couldn't get the test to work -- isIntersecting was always undefined in the component's `handleChagne` method


phil-plencner-hl:
@marlo-longley The tests for not calling `requestCanvasAnnotations` fail for me:

`GalleryView › on-demand annotation fetching › does nothing if there is an intersection and existing annotations

    expect(jest.fn()).not.toHaveBeenCalled()

    Expected number of calls: 0
    Received number of calls: 1

    1: called with 0 arguments

      102 |       createWrapper({ annotationsCount: 1, canvas, requestCanvasAnnotations });
      103 |       mockAllIsIntersecting(true);
    > 104 |       expect(requestCanvasAnnotations).not.toHaveBeenCalled();
          |                                            ^`

phil-plencner-hl (approved):
Merging this in, knowing we have disabled the broken tests.

Issue #3653, opened 2023-02-24
Rewrite Enzyme tests with react-testing-library: Workspace-adjacent tests


cbeer:
- [x] `__tests__/src/components/ManifestForm.test.js`
- [x] `__tests__/src/components/ManifestListItem.test.js`
- [x] `__tests__/src/components/ManifestListItemError.test.js`
- [x] `__tests__/src/components/Workspace.test.js`
- [x] `__tests__/src/components/WorkspaceAdd.test.js`
- [x] `__tests__/src/components/WorkspaceAddButton.test.js`
- [x] `__tests__/src/components/WorkspaceArea.test.js`
- [x] `__tests__/src/components/WorkspaceControlPanel.test.js`
- [x] `__tests__/src/components/WorkspaceControlPanelButtons.test.js`
- [x] `__tests__/src/components/WorkspaceElastic.test.js`
- [x] `__tests__/src/components/WorkspaceElasticWindow.test.js`
- [x] `__tests__/src/components/WorkspaceExport.test.js`
- [x] `__tests__/src/components/WorkspaceImport.test.js`
- [x] `__tests__/src/components/WorkspaceMenu.test.js`
- [x] `__tests__/src/components/WorkspaceMenuButton.test.js`
- [x] `__tests__/src/components/WorkspaceMosaic.test.js`
- [x] `__tests__/src/components/WorkspaceOptionsButton.test.js` (ML)
- [x] `__tests__/src/components/WorkspaceOptionsMenu.test.js` (ML)
- [x] `__tests__/src/components/MosaicRenderPreview.test.js`
- [x] `__tests__/src/components/WindowList.test.js`
- [x] `__tests__/src/components/WindowListButton.test.js`

Extracted from #3597

marlo-longley:
@cbeer Since I am refactoring those components I will rewrite the tests for:
```
__tests__/src/components/WorkspaceOptionsButton.test.js
__tests__/src/components/WorkspaceOptionsMenu.test.js
```

PR #3702, opened 2023-03-09
Rewrite tests for WorkspaceElastic with RTL

cbeer:
Part of #3653


marlo-longley:
Wow thank you for figuring out how to use the pointer with math to drag! Nice.

marlo-longley (approved):

Issue #3654, opened 2023-02-24
Rewrite Enzyme tests with react-testing-library: Lower-level component tests


cbeer:
```
__tests__/src/components/CollapsibleSection.test.js
__tests__/src/components/CompanionArea.test.js
__tests__/src/components/CompanionWindow.test.js
__tests__/src/components/CompanionWindowFactory.test.js
__tests__/src/components/ErrorDialog.test.js
__tests__/src/components/MiradorMenuButton.test.js
__tests__/src/components/SanitizedHtml.test.js
__tests__/src/components/ScrollIndicatedDialogContent.test.js
__tests__/src/components/ScrollTo.test.js
__tests__/src/components/NestedMenu.test.js
__tests__/src/components/BackgroundPluginArea.test.js
__tests__/src/components/PluginHook.test.js
__tests__/src/extend/withPlugins.test.js

```

Extracted from #3597

PR #3709, opened 2023-03-13
Update SanitizedHtml tests for RTL

cbeer:
Part of #3654

### DIFF
--- a/src/components/SanitizedHtml.js
+++ b/src/components/SanitizedHtml.js
@@ -10,7 +10,9 @@ export class SanitizedHtml extends Component {
   /**
   */
   render() {
-    const { classes, htmlString, ruleSet } = this.props;
+    const {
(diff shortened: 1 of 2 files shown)


src/components/SanitizedHtml.js
```diff
@@ -28,6 +30,7 @@ export class SanitizedHtml extends Component {
         dangerouslySetInnerHTML={{ // eslint-disable-line react/no-danger
           __html: DOMPurify.sanitize(htmlString, htmlRules[ruleSet]),
         }}
+        {...props}
```
marlo-longley, line 33: Is this just to get the test id?

cbeer: Pretty much, but also seems like a nice thing for downstream consumes 🤷‍♂️

marlo-longley (approved):

Issue #3655, opened 2023-02-24
Rewrite Enzyme tests with react-testing-library: Other


cbeer:
- [x] `__tests__/src/components/App.test.js`
- [x] `__tests__/src/components/AppProviders.test.js`
- [x] `__tests__/src/components/CollectionDialog.test.js`
- [x] `__tests__/src/components/LocalePicker.test.js`
- [x] `__tests__/src/components/LanguageSettings.test.js`
- [x] `__tests__/src/components/FullScreenButton.test.js`
- [x] `__tests__/src/lib/MiradorViewer.test.js`
- [x] `__tests__/src/components/AccessTokenSender.test.js`
- [x] `__tests__/src/components/IIIFAuthentication.test.js`
- [x] `__tests__/src/components/WindowAuthenticationBar.test.js`
- [x] `__tests__/src/components/NewWindow.test.js`

Extracted from #3597

PR #3717, opened 2023-03-14
Update NewWindow tests to use RTL

cbeer:
Part of #3655


marlo-longley:
delete 'screen' import so that Github Actions cop is happy?

Issue #3650, opened 2023-02-24
Rewrite Enzyme tests with react-testing-library: Window-adjacent tests


cbeer:
- [x] __tests__/src/components/PrimaryWindow.test.js
- [x] __tests__/src/components/Window.test.js
- [x] __tests__/src/components/WindowCanvasNavigationControls.test.js
- [x] __tests__/src/components/WindowSideBar.test.js
- [x] __tests__/src/components/WindowTopBar.test.js
- [x] __tests__/src/components/WindowTopBarPluginArea.test.js
- [x] __tests__/src/components/WindowTopBarPluginMenu.test.js
- [x] __tests__/src/components/WindowTopBarTitle.test.js
- [x] __tests__/src/components/WindowTopMenu.test.js
- [x] __tests__/src/components/WindowTopMenuButton.test.js

Extracted from #3597

PR #3719, opened 2023-03-14
Convert tests for Window to RTL

marlo-longley:
When running the Window test (not the ErrorContent test) I get this warning:
` Warning: [JSS] Could not find the referenced rule "icon" in "Connect(WithPlugins(ErrorContent))".`
-- which comes from the style in the container set [here](https://github.com/ProjectMirador/mirador/blob/a708a7d9d30af99fd1501a26e08386c0868cbcc3/src/containers/ErrorContent.js#L33).

Is this OK to leave?


marlo-longley:
Ref #3650

PR #3721, opened 2023-03-15
Use react-intersection-observer's IntersectionObserver mock

cbeer:


phil-plencner-hl (approved):
This makes sense to me!

PR #3723, opened 2023-03-15
Convert PrimaryWindow test to RTL

marlo-longley:
- Edit text in ErrorContent test
- Convert PrimaryWindow test to RTL


marlo-longley:
Ref #3650

PR #3731, opened 2023-03-20
Convert tests for IIIFAuthentication to RTL

marlo-longley:


src/components/WindowAuthenticationBar.js
```diff
@@ -45,7 +45,7 @@ export class WindowAuthenticationBar extends Component {
     const { open } = this.state;
 
     const button = (
-      <Button onClick={this.onSubmit} className={classes.buttonInvert} color="secondary" {...ConfirmProps}>
+      <Button onClick={this.onSubmit} className={classes.buttonInvert} color="secondary" {...ConfirmProps} aria-label="confirm action">
```
cbeer, line 48:
```suggestion
      <Button onClick={this.onSubmit} className={classes.buttonInvert} color="secondary" {...ConfirmProps}>
```

I don't think we want to add this aria-label -- it seems like the button's content should be sufficient to identify it and should be a better label for our users anyway?

PR #3735, opened 2023-03-27
Convert tests for WindowAuthenticationBar to RTL

marlo-longley:


src/components/WindowAuthenticationBar.js
```diff
@@ -67,7 +67,7 @@ export class WindowAuthenticationBar extends Component {
 
     return (
       <Paper square elevation={4} color="secondary" classes={{ root: classes.paper }}>
-        <Button fullWidth className={classes.topBar} onClick={() => this.setOpen(true)} component="div" color="inherit">
+        <Button fullWidth className={classes.topBar} onClick={() => this.setOpen(!open)} component="div" color="inherit">
```
cbeer, line 70: I'm not so sure about this. This button/header/whatever isn't really a toggle, is it?

marlo-longley: The old test seemed to indicate it should recollapse. The functionality didn't actually line up so I'm not sure how the old test was passing?
 https://github.com/ProjectMirador/mirador/blob/b2170c5d4dd660d4f4cdcbbeb1753c4299c1ee0b/__tests__/src/components/WindowAuthenticationBar.test.js#L60

marlo-longley: Got it. I had the buttons mixed up with the [0] selector. I fixed it, removed `!open` and tried to make the collapse test a little clearer.

Issue #3494, opened 2021-09-02
Window top bar buttons do not work in elastic mode on mobile devices


lutzhelm:
The window top bar buttons (toggle sidebar, window view selection, fullscreen, close) don't seem to work in elastic mode on mobile devices. This seems to be platform and browser independent; it doesn't even work in the desktop browser developer tools.

So far I've checked
* Chrome 92.0 on Android 9
* Firefox 91.4 on Android 9
* Safari 14.1 on iOS
* Mobile views in the developer tools of desktop versions of Firefox and Chromium on Ubuntu

PR #3736, opened 2023-03-29
fix: elastic window DnD on mobile

bodo22:
This is a fix for #3494.
It's a workaround for a bug that has been around in react-draggable since [v4.3.0 of react-draggable](https://github.com/react-grid-layout/react-draggable/blob/master/CHANGELOG.md#430-apr-10-2020).

A good description of why this bug happens: https://github.com/react-grid-layout/react-draggable/issues/550#issuecomment-947166613.

By canceling the drag event when one tries to drag on the top menu buttons, they become clickable again. The same feature is already used here: https://github.com/ProjectMirador/mirador/blob/b65ef2c778d2d84aa82e9a1054a10bae6263d3ad/src/components/WorkspaceElastic.js#L58.

If this workaround is not sufficient, the only other way (besides fixing react-draggable) I see is to implement some custom "click" logic, like described here: https://github.com/react-grid-layout/react-draggable/issues/550#issuecomment-1221555269.

### DIFF
--- a/src/components/WindowTopBar.js
+++ b/src/components/WindowTopBar.js
@@ -48,6 +48,7 @@ export class WindowTopBar extends Component {
               <MiradorMenuButton
                 aria-label={t('toggleWindowSideBar')}
                 onClick={toggleWindowSideBar}
+                className={ns('window-menu-btn')}
               >
                 <MenuIcon />
               </MiradorMenuButton>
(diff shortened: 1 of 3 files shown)


bodo22:
@cbeer thanks!

I'm not sure this is answered somewhere already, apologies if it is.
Is there a release to npm procedure, or is anything blocking a new npm release (including this fix)? I see the last publish to npm is from 11/2021.

enriquediaz:
Hi @bodo22, there is [documentation](https://github.com/ProjectMirador/mirador/wiki/How-to-release-a-version-of-Mirador) about releasing to npm, but as you've guessed there is currently a blocker. Major changes in progress to bring Mirador up to date with its core dependencies, not to mention overhauling the testing suite which is the lion's share of the work in many ways. 

Work and planning is underway towards a pre-release that includes this and other updates, and we'll be reaching out to the community to solicit your help testing it out to get us to a full release further down the road. Stay tuned!

Issue #3752, opened 2023-03-30
Focused windows don't float to the front in elastic mode on mobile devices


cbeer:
Normally, when the user clicks on the window it becomes "focused" and appears above other windows. On mobile devices, this doesn't seem like it happens.

Found when testing #3736

PR #3757, opened 2023-04-03
Allow WorkspaceElasticWindow to focus

marlo-longley:
Fixes https://github.com/ProjectMirador/mirador/issues/3752


cbeer:
👍 ; Can we add a test for this to `WorkspaceElasticWindow.test.js`?

Issue #1969, opened 2019-02-26
additional http headers


eroux:
I have another use case for an http header to send to requests made by the viewer: UV will be embedded in a website where users will choose language preferences. From these language preferences we will generate a string that can go into a `Accept-Language` http header (or maybe a `X-BDRC-Accept-Language` if we can't encode some subtleties). We would like to pass this header to requests made to the iiif presentation api, so that our server can encode the labels of the various property values it returns. This could be as simple as a configuration like:
- an object containing the headers
- a function to get the additional headers (better if the user changes their preferences)

see https://github.com/UniversalViewer/universalviewer/issues/631#issuecomment-467395627

PR #3770, opened 2023-06-08
Add an Accept header for manifests

jcoyne:
Our IIIF server provides v2 APIs by default, but if the client provides an Accept header with the v3 profile, it will provide a v3 response.  This triggers that behavior.

### DIFF
--- a/src/state/sagas/iiif.js
+++ b/src/state/sagas/iiif.js
@@ -105,13 +105,24 @@ function* fetchIiifResourceWithAuth(url, iiifResource, options, { degraded, fail
   yield put((degraded || success)({ json, response, tokenServiceId }));
 }
 
+// Prefer v3 API response, but v2 is acceptable or even any JSON+LD or JSON.
+const MANIFEST_ACCEPT_HEADER = 'application/ld+json;q=0.9;profile="http://iiif.io/api/presentation/3/context.json", '
+                               + 'application/ld+json;q=0.7;profile="http://iiif.io/api/presentation/2/context.json", '
+                               + 'application/ls+json;q=0.5, '
+                               + 'application/json;q=0.2';
+
 /** */
 export function* fetchManifest({ manifestId }) {
   const callbacks = {
     failure: ({ error, json, response }) => receiveManifestFailure(manifestId, typeof error === 'object' ? String(error) : error),
     success: ({ json, response }) => receiveManifest(manifestId, json),
   };
-  const dispatch = yield call(fetchIiifResource, manifestId, {}, callbacks);
+  const dispatch = yield call(
+    fetchIiifResource,
+    manifestId,
+    { headers: { Accept: MANIFEST_ACCEPT_HEADER } },
+    callbacks,
+  );
   yield put(dispatch);
 }
 


jbaiter:
Warning: I just tried to implement this in my own application, and it seems to break compatibility with some IIIF endpoints, most notably that of the Vatican.

The reason for this is, that by setting the `Accept` header on the `fetch` Request, a pre-flight `OPTIONS` request is now sent to the Manifest endpoint. If that endpoint does not respond correctly, the browser is unable to make the request and users won't be able to view the Manifest.

Example: I just implemented this and discovered an unfortunate side-effect of setting the `Accept` Header: It surfaces broken `OPTIONS` request handlers in IIIF implementations...

Example: https://digi.vatlib.it/iiif/MSS_Arch.Cap.S.Pietro.C.132/manifest.json
Fetches fine without the `Accept` header, but breaks with a 403 (without a CORS header on the response) on the pre-flight request that is now sent due to the presence of the custom 'Accept' header.

jcoyne:
@jbaiter can you give me more details about this?  I'm not seeing this behavior in a simple test:

```html
<!DOCTYPE html>
<html>
<head>
</head>
<body>
<script>

const run = async function () {
  const url = 'https://digi.vatlib.it/iiif/MSS_Arch.Cap.S.Pietro.C.132/manifest.json'
  console.log('fetch with no headers')
  const result = await fetch(url)
  console.log("done")
  console.log(result)

  console.log('fetch with headers')
  const result2 = await fetch(url, { Accept: 'application/ld+json, application/json' })
  console.log("done")
  console.log(result2.json())

  console.log('fetch with headers')
  const result3 = await fetch(url, { Accept: 'application/ld+json;q=0.9;profile="http://iiif.io/api/presentation/3/context.json", application/ld+json;q=0.7;profile="http://iiif.io/api/presentation/2/context.json"'})
  console.log("done")
  console.log(result3.json())
}
run()
</script>
<p>Look in the console</p>
</body>
</html>
```

jbaiter:
Gotcha, I found out why my code was triggering the preflight request: I was using a `fetch` ponyfill (`cross-fetch`) that somehow didn't use native `fetch` in the browser but it's own XHR-based polyfill. Since the rules for a "simple request" are stricter for `XMLHttpRequests`, the `fetch` was triggering a preflight request in my case. This does not happen when using native fetch, so sorry for the false alert 🙏🏾

jbaiter:
Sorry, but turns out that wasn't it,  I'm seeing the same result with native `fetch` now (latest Firefox).

I think your sample code is not actually setting the `Accept` header, you have to set the `headers` value in the fetch options dict, and not pass in the headers as the second argument:

https://kk7jk2.csb.app/
https://codesandbox.io/s/divine-hooks-kk7jk2?file=/index.html

jbaiter:
I think we're running into this:

> If value’s length is greater than 128, then return false. 

https://fetch.spec.whatwg.org/#cors-safelisted-request-header

`Accept` headers are usually safe and don't trigger a preflight request, except if the above condition is met y which is the case, unfortunately:-/
Maybe we can strip down the value so it stays below 128 characters?

irv:
I don't have anything really useful to contribute but @jbaiter isn't is a semicolon `;` and not a colon `:` in the accept header?

marlo-longley:
Mirador should support this via the `config.requests.preprocessors` setting (added in #3082, following #1969/#2588) — if their server handles it people can add the Accept header themselves. 
I think this would avoid the issue on CORS brought up by @jbaiter 
I am closing this PR as stale, but noting the issue in our bucket of topics to document better.


If I've misunderstood the issue, please feel free to comment or reopen. Thanks everyone!

PR #3776, opened 2023-07-18
Fix Japanese Translation

dknijl:
It's easy to misunderstand the label "資料のURL" and enter a regular URL instead of the required manifest URI. Therefore, I would like to propose this PR to correct this translation issue.


gerdesque:
Hi @dknijl, thanks for your pull request. Due to the failed builds in this PR, I have duplicated your changes in https://github.com/ProjectMirador/mirador/pull/3910 and will therefore close this one.

PR #3788, opened 2023-10-02
add missing german translations

fstoe:
When running the linter I found out that a few translation keys are missing for German.


jcoyne (approved):
Thank you for adding these.

Issue #3789, opened 2023-10-02
IIIF Auth2 Support


barmintor:
Mirador should support the [IIIF Auth2](https://iiif.io/api/auth/2.0/) specification.

This will require changes to canvas processing, as non-image content resources may have services.

This will require changes to the authorization flow to support probe responses.

brittnylapierre:
Hey all! Was this being worked on already?) as referenced in the activity occurring above?

barmintor:
We are imminently launching a site built with a fork of M3 that uses Auth2, but building a PR from it will be some additional work.

marlo-longley:
Audit 2026: Top priority. 
Rebased branch here https://github.com/ProjectMirador/mirador/pull/4384 -- developer can elect to work off that, or start fresh.

PR #3790, opened 2023-10-02
Auth2 support

barmintor:
cf #3789


src/lib/canvasTypes.js
```diff
@@ -0,0 +1,37 @@
...
+const imageServiceProfiles = [
+  'level2',
+  'level1',
+  'level0',
+  'http://iiif.io/api/image/2/level2.json',
+  'http://iiif.io/api/image/2/level1.json',
+  'http://iiif.io/api/image/2/level0.json',
+];
+
+Object.freeze(imageServiceProfiles);
```
cbeer, line 27: I wonder if these could/should go into settings somewhere so downstream apps can customize them?

barmintor: something similar to `auth.serviceProfiles` maybe?

PR #3793, opened 2023-10-26
Fix MUI5 testing errors

marlo-longley:


barmintor (approved):
I think there are two changes to consider in the comments but I don't think they should hold up merging this PR to the mui5 branch.

marlo-longley:
Because of a migration, I had to add mui-x to our project to keep using Tree View:
https://github.com/ProjectMirador/mirador/pull/3793/commits/8932afd5dc5a015ea1c45fcd3ac6a293193d81a9

It is a paid product but is also free to use under MIT License, so I think we should. be good. But wanted to check on that.

barmintor:
@marlo-longley I came to the same conclusion for the React18 patches on our fork.

phil-plencner-hl (approved):
Ran `npm run test` and all tests pass. Approved.

PR #3795, opened 2023-10-27
Mui5 test fixups

barmintor:
Largely restoring marker classes used in tests, but there is a less ambiguous change:
- recent Material-UI uses a different role (combobox) for select elements (previously button)

This PR can otherwise be used to point to specs that need to have a select on the marker class removed - every commit should point to one failing test.


cbeer:
Merged in #3804.

PR #3796, opened 2023-10-27
Fix double rendering of SearchHit and its placement

marlo-longley:
before
Screenshot 2023-10-26 at 3 58 27 PM

after
Screenshot 2023-10-26 at 5 47 21 PM

There may be a better way to style this but it should probably be fixed!

### DIFF
--- a/src/components/SearchHit.js
+++ b/src/components/SearchHit.js
@@ -136,9 +136,6 @@ export class SearchHit extends Component {
                   verticalAlign: 'inherit',
                 }}
               />
-              <span id={canvasLabelHtmlId}>
-                {canvasLabel}
-              </span>
             </Typography>
             {annotationLabel && (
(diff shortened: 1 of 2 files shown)


src/components/SearchHit.js
```diff
@@ -136,9 +136,6 @@ export class SearchHit extends Component {
                   verticalAlign: 'inherit',
                 }}
               />
-              <span id={canvasLabelHtmlId}>
```
cbeer, line 139: I'm surprised you're removing the canvas label here. Was this intentional?

See https://github.com/ProjectMirador/mirador/issues/2683 for what we're expecting the layout to look like.

marlo-longley: Thanks Chris. I guess I was totally wrong about this, and it is an expected annotation label. I'll just close this PR.
Screenshot 2023-11-22 at 12 57 03 PM

PR #3834, opened 2023-12-04
Push companion window styles up to settings

cbeer:


src/config/settings.js
```diff
@@ -177,6 +177,68 @@ export default {
...
+          positionButton: {
+            marginLeft: -16,
+            order: -100,
+            width: 24,
+          },
+          resize: ({ ownerState }) => ({
+            display: 'flex',
+            flexDirection: 'column',
+            minHeight: 50,
+            minWidth: (ownerState?.position === 'left') ? 235 : 100,
```
marlo-longley, line 207: Can you remove the minWidth here if it has been relocated? 

https://github.com/ProjectMirador/mirador/blob/cefe7b7b520ae1cb637d935c7b3cfceaff96b14a/src/components/CompanionWindow.js#L110

cbeer: I don't think so? the react-rnd library wants to know that information too 🤷‍♂️

PR #3864, opened 2023-12-14
Update to React 18

cbeer:


daxid:
Hi, I'm stopping by after a discussion on Slack IIIF.

We are putting some effort to upgrade the plugin here : https://gitlab.tetras-libre.fr/iiif/mirador/mirador-annotations

The strategy is to follow what's happening on the Mirador side. We first target an upgrade to React17 and MUI5 (against the corresponding `mui5` branch of Mirador), then we will continue to React 18.

In the process we also target major UI improvements, video annotation support, etc.

Finally, our first investigations towards a React18 upgrade showed that the paper.js library that is used in `mirador-annotations` to draw and render the annotations may be a show stopper. Indeed, the library seems unmaintained since 2 years or so and the current version is recommended wit React17. We managed to make it work with React18.0 but not with the latest React18.2.

Anyway, I think that an in place replacement for paper.js would be very welcome, if not absolutely necessary. Any input on the matter is very welcome.

d-flood:
Thanks for all the work on this, @cbeer!

I've tested this branch in a project we're (https://github.com/artshumrc) working on right now. I did not find any issue or regressions with the core Mirador viewer. Progress seems to be going well for this upgrade. With the below exceptions, it was essentially a drop-in replacement that worked.

I'm sure it is no surprise that I needed to disable our image-tools and annotations plugins in order to build and run this version.

It would be easier for us to test these upgrades in a real world project if there was a prelease distribution that we could include in our CI/CD pipeline. Perhaps I am missing something obvious, but the only way I found to test this branch in our project was to build from source myself and include the distribution in our project.

cbeer:
Merging per Mirador Community Call 3/14

PR #3871, opened 2024-01-10
Refactor manifesto initialization to push property localization down

### LINKED ISSUE #3470: Multilingual manifest labels

cbeer:
From #3482:
> A recent change in the Manifesto library made it possible to pass a list of languages (in descending order of preference) to the PropertyValue.getValue and PropertyValue.getValues methods, instead of relying on a single default locale that would be set when parsing the manifest. This allows for a better user experience, since we can now dynamically respond to changes in the user's language preferences without having to re-parse the manifest. Additionally, we can make use of the browser's language preferences to better match the user's language abilities.

This does the same thing as #3482, without introducing new redux state or a new selector parameter, in order to maintain API compatibility for downstream plugins. Instead, this PR:
 
- introduces a `LocaleContext` and locale-context aware `IIIFResourceLabel` for rendering e.g. manifest, canvas, etc labels appropriately, replacing a lot of piece-meal label functions
- ensures components within a companion window include that context when calling selectors (to get the `getLocale`/`getManifestLocale` to work right... this is somewhat unfortunate, but probably less weird if we started to use hooks  or more extensive use of context like #3661 (and a companion-window equivalent))
- Clean up styling for the companion window locale picker to make it more like the workspace language selector

This can be tested at https://deploy-preview-3871--mirador-dev.netlify.app/__tests__/integration/mirador/i18n.html

Fixes #3470


cbeer:
Replaced by #4054.

PR #3872, opened 2024-01-11
Move window topbar content to menu on small screens

### LINKED ISSUE #2891: Window toolbar collapse in very small windows
This is a current issue in M3, and will be exacerbated as additional menus may be added to the window toolbar. In very small windows (e.g. multi-up comparison) we currently collapse the title until it disappears, then start clipping icons from the right. The functionality from those clipped icons is not available to the user unless the window is expanded, which they may not know/want to do.

Examples below show the addition of one menu, but the problem exists without that addition.

2-up window on laptop: All buttons visible, title slightly truncated.
moderate
Smaller: title nearly disappears
small
Smallest (in mosaic mode): title gone; close and fullscreen buttons are clipped. 
extra small

### Possible approach

- Maintain a balance between the title and buttons. 
- Always retain the close button and the dot button.
- As the window shrinks, collapse the remaining buttons into the dot menu, starting from the right. 

collapse1
collapse2
collapse3
collapse4

Needs more analysis, and there might be better solutions. Need to design how the dot menu would work as a 2-step menu. Recording for future work.

bodo22:
This is an attempt to address #2891


lutzhelm (commented):
* Keyboard navigation does not work in the small variant yet. I don't know how relevant keyboard navigation is for very small screens (i.e. probably phones), but I'd assume it's still an accessibility requirement.
* The way this is implemented results in a change of order of the buttons.
    *  up until now, the order was
        * Window view / thumbnail display
        * Plugin area
        * Maximize / restore window
        * Full screen
        * Close Window
    * The plugins / more button has now moved behind Maximize and Fullscreen. The question is if the latter two should better be kept together with the Close button, but this would make the change a little more complicated. I'd really like a bit more feedback on that.

Edit, for illustration...

Before:
topbar-before

After:
topbar-after

marlo-longley:
We should address these comments and rebase/merge if possible

PR #3901, opened 2024-04-17
Add maximized option to windows in config

bertyhell:
Since this file is used as the docs it is otherwise hard to know how to maximize a window from config

see source code:
https://github.com/ProjectMirador/mirador/blob/94d41eff7f7763c501ae0247b9ae892c5b3987fc/src/containers/WindowTopBar.js#L16C45-L16C54


lutzhelm (approved):
Good catch. Thank you!

PR #3908, opened 2024-06-05
Fix full screen mode issue where modals windows do not open in the foreground

gerdesque:
There is currently an issue with full screen mode where modal windows, when triggered, do not open in the foreground as expected. Instead, they open in the background and are only visible after leaving full screen mode. This results in a poor user experience, as users cannot interact with the modals while in full screen mode. This pull request aims to address and fix this issue. We will ensure that modals always open in the user's immediate view, regardless of mode.

To do this, I have replaced the current react-fullscreen library with a custom fscreen handle component. This was suggested in the following issue of react-fullscreen, see https://github.com/snakesilk/react-fullscreen/issues/63#issuecomment-1191911295

Current:
[fullscreen_modal_not_working.webm](https://github.com/ProjectMirador/mirador/assets/19520033/72b6a181-da93-4bee-a70c-77985dab285f)

New:
[fullscreen_modal_working.webm](https://github.com/ProjectMirador/mirador/assets/19520033/c80f9f43-9f79-4bba-92f4-96b33442b558)


marlo-longley:
Relates to https://github.com/ProjectMirador/mirador/issues/4356

marlo-longley:
Thank you for this report and this PR.
It is out of date, so during the audit I fixed it up and based a new PR: https://github.com/ProjectMirador/mirador/pull/4367
Closing this PR in favor of the updated one

Issue #3903, opened 2024-04-30
Window dropdown for plugins not keyboard navigable


mferrarini:
### Description
Initial testing indicates interface elements like Download PDF, Share, and Show Image Tools are not accessible through keyboard navigation, just by mouse. 

I was able to navigate through the "windows options" menu items with voiceover using ctrl+opt+left/right arrows, but this same combination does not work for me when voiceover is turned off. Did not work with NVDA either. This follows the same pattern of what we saw when investigating keyboard navigation through the info panels.

### Problem
The list items are given a tab-index of "-1" and therefore are being passed over when tabbing through the page.

### Possible solution
Removing the tabIndex="=1" from the `<ListSubheader>` material UI component should be all that's needed to make this work (and might also solve the problem in other areas as well). 

I noticed this component in the following files:
- [WindowThumbnailSettings.js](https://github.com/ProjectMirador/mirador/blob/master/src/components/WindowThumbnailSettings.js#L56)
- [WindowTopMenu.js](https://github.com/ProjectMirador/mirador/blob/master/src/components/WindowTopMenu.js#L14)
- [WindowViewSettings.js](https://github.com/ProjectMirador/mirador/blob/master/src/components/WindowViewSettings.js#L83)

lutzhelm:
I'm not sure if #3914 fixes this entirely; I will have to try this with an actual plugin installed. But the window view settings / thumbnail navigation options are now keyboard accessible.

marlo-longley:
Audit 2026: action -- check if this is still an issue in Mirador 4
Can use Mirador integration repo

PR #3914, opened 2024-06-11
fix keyboard navigation issues with windowTopMenu

fstoe:
Hello, 
this PR tries to fix the issue described here: https://github.com/ProjectMirador/mirador/issues/3903

The reason why view and thumbnail options could not be controlled correctly was due to the use of the MUI Menu component and not the wrong TabIndex as described in the issue. The component expects MenuItems as direct children. This was fixed by using a popover instead of the menu, (on which menu is based upon) and setting the MenuLists in WindowViewSettings and WindowThumbnailSettings.


fstoe:
Edit: It seems as the keyboard navigation worked as expected when M3 was still using Material Ui v4.

gerdesque (approved):

src/components/WindowViewSettings.js
```diff
@@ -63,10 +67,11 @@ export class WindowViewSettings extends Component {
         none of the click handlers work? */
     const menuItem = ({ value, Icon }) => (
       <ViewOption
-        selected={windowViewType === value}
-        key={value}
+        aria-selected={windowViewType === value}
```
cbeer, line 70: Is aria-selected a valid/useful attribute for menu items? I'm not seeing it on https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menuitem_role. Maybe this should be a `menuitemradio` + `aria-checked`?

src/components/WindowViewSettings.js
```diff
@@ -15,10 +16,13 @@ const ViewOption = styled(MenuItem, { name: 'WindowViewSettings', slot: 'option'
     ...(selected && {
       borderBottomColor: theme.palette.secondary.main,
     }),
+    backgroundColor: 'transparent !important',
```
cbeer, line 19: I'm fixing up the current styling in https://github.com/ProjectMirador/mirador/pull/3982. The original designers had a strong preference for the underscore-selected state and consistent (grey) focus state instead of the MUI default behavior.

src/components/WindowViewSettings.js
fstoe: Ah thanks for spotting :) 
i added `aria-checked` and `menuitemrole`

gerdesque (dismissed):

src/components/WindowThumbnailSettings.js
```diff
@@ -40,43 +44,66 @@ export function WindowThumbnailSettings({
...
+              <ThumbnailNavigationBottomIcon color={thumbnailNavigationPosition === 'far-bottom' ? 'secondary' : undefined} fill="currentcolor" />
+            }
+            label={t('bottom')}
+            labelPlacement="bottom"
+            value="far-bottom"
+          />
+        </ThumbnailOption>
+        <ThumbnailOption
+          aria-checked={thumbnailNavigationPosition === 'far-right'}
+          autoFocus={thumbnailNavigationPosition === 'far-right'}
```
lutzhelm, line 85: The `autoFocus` for the `WindowThumbnailSettings` (either for the `off`, `far-bottom` or `far-right` state) takes precedence over the `autoFocus` in the `WindowViewSettings`. This results in a weird tab order where the second control is in focus before the first.

There might be situations where there are no view options available, and thus `WindowViewSettings` is not rendered at all; only then should `autoFocus` be set for the `WindowThumbnailSettings`.

lutzhelm: For the moment I would be fine with just removing the `autoFocus` from `WindowThumbnailSettings`.

fstoe: I just removed it

lutzhelm: Thanks!

lutzhelm (approved):

PR #3922, opened 2024-06-26
Bug fix: preservation of viewport doesnt work

fstoe:
I recently found out that setting [preserveViewport in the osd settings](https://github.com/ProjectMirador/mirador/blob/master/src/config/settings.js#L545) does nothing. 
So I made some little changes to the OpenSeadragonViewer Component and to the setcanvas action.


src/components/OpenSeadragonViewer.js
```diff
@@ -118,7 +118,11 @@ export class OpenSeadragonViewer extends Component {
     ) {
       viewer.close();
       const canvasesChanged = !(isEqual(canvasWorld.canvasIds, prevProps.canvasWorld.canvasIds));
-      this.addAllImageSources((canvasesChanged || !viewerConfig));
+      if (canvasesChanged && viewer.preserveViewport) {
+        this.addAllImageSources(false);
+      } else {
+        this.addAllImageSources((canvasesChanged || !viewerConfig));
```
marlo-longley, line 124: I found this negative logic a bit hard to follow (not from what you wrote, but from what existed before!) -- what do you think about adding comments similar to:

```
 if (canvasesChanged && viewer.preserveViewport) {
        // Do not reset the zoom
        this.addAllImageSources(false);
      } else {
       // Reset the zoom if the canvas has changed or if there is no viewerConfig
        this.addAllImageSources((canvasesChanged || !viewerConfig));
      }
...
```
 ?

marlo-longley:
@fstoe I tested this locally and it works well, thank you for submitting.

marlo-longley (approved):

marlo-longley (approved):

PR #3926, opened 2024-07-29
Bug fix: viewer not scrolling to toc entry correctly

fstoe:
I have just fixed a bug. This is related to the table of contents and the ScrollTo component.
Previously, an empty containerRef was passed to it inital, which meant that it did not scroll to the correct position in the TOC.

I used this config prop: `windows: [{ manifestId: "https://iiif.ub.uni-leipzig.de/0000034390/manifest.json", canvasIndex: 330,}]` for testing.


gerdesque (approved):

lutzhelm:
It looks to me like this has already been fixed via #4077 in the meantime (also for other affected components), so this PR can be closed?

PR #3927, opened 2024-07-30
fix:mirador_crash

fstoe:
I found out that adding something through the add button in the catalog view which is json but not a manifest makes the viewer crash.
steps to reproduce:
- Open Mirador
- Add Resource with example url:  https://iiif.io/api/presentation/2/context.json
- Open legit resource


lutzhelm:
Hi @fstoe, could you add a test case for that patch? `createManifestoInstance` itself is not exported, but `getManifestoInstance` calls it directly, so this would be a good place to add a test:

https://github.com/ProjectMirador/mirador/blob/master/__tests__/src/selectors/manifests.test.js#L56-L68

Something like
``` javascript
it('does not crash if json is not a manifest', () => {
  const state = { manifests: { x: { json: {} } } };
  const received = getManifestoInstance(state, { manifestId: 'x' });
  expect(received).toBe(undefined);
});
```

fstoe:
Hey @lutzhelm, 
thanks for the comment. I just added a small test case

__tests__/src/selectors/manifests.test.js
```diff
@@ -59,6 +62,14 @@ describe('getManifestoInstance', () => {
...
   });
+  it('creates a manifesto collection instance and patches getSequences', () => {
+    const state = { manifests: { x: { json: collection } } };
+
+    const received = getManifestoInstance(state, { manifestId: 'x' });
+    expect(typeof received.getSequences).toBe('function');
+
+    expect(received.getSequences()).toEqual([]);
+  });
   it('is cached based off of input props', () => {
```
lutzhelm, line 73: This test has been removed on `main`. I don't know the reason, but it's the cause of the current merge conflict. Could you resolve this conflict e.g. by merging `main` into this branch? Just be sure that the test case actually stays removed.

__tests__/src/selectors/manifests.test.js
```diff
@@ -59,6 +62,14 @@ describe('getManifestoInstance', () => {
     const received = getManifestoInstance(state, { manifestId: 'x' });
     expect(received.id).toEqual('http://iiif.io/api/presentation/2.1/example/fixtures/19/manifest.json');
   });
+  it('creates a manifesto collection instance and patches getSequences', () => {
```
lutzhelm, line 65: Not important, but could you insert an empty line above? The eslint rules actually don't enforce that but it would be more consistent. Thanks!

lutzhelm (changes requested):

__tests__/src/selectors/manifests.test.js
fstoe: Done :)

lutzhelm (dismissed):

src/containers/ManifestListItem.js
```diff
@@ -23,7 +23,7 @@ const mapStateToProps = (state, { manifestId, provider }) => {
     : getCanvases(state, { manifestId }).length;
   return {
     active: getWindowManifests(state).includes(manifestId),
-    error: manifest.error,
+    error: manifest.error || !manifesto,
```
lutzhelm, line 26: I should have noticed this earlier: This change results in a list of errors in the catalog list for every manifest that is configured in the `catalog` settings but not in `windows`. While you can "Try again" successfully, this should be fixed. It's probably sufficient to change this line to
```
    error: manifest.error || (!manifesto && !!manifest.json),
```
Manifests that have not been used in windows before don't have a manifesto instance up until this point, and at the moment the component is initially rendered, `isFetching` is not yet true, so the same check that is used for `ready` should be used here additionally.

lutzhelm: I'm also thinking about how to test this automatically, no efficient idea yet, but I think it's not necessary for the scope of that PR.

fstoe: thanks for spotting.

lutzhelm: Thanks, looks better now.

lutzhelm (approved):

PR #3943, opened 2024-10-29
Revert "Update translation.json"

jcoyne:
This reverts commit 883051aa681bcf577a22273c2fcfb09a01659dde.

This commit changed the structure of the json so that it causes scripts/i18n-lint.js to fail


jcoyne:
Fixed by #3941

PR #3945, opened 2024-10-29
Remove proposal

jcoyne:


cbeer:
Do we also need to bump the babel versions to make sure we're using a version that bakes in those proposals? It looks like `plugin-proposal-object-rest-spread` was the last to be promoted, maybe around babel 7.22?

PR #3948, opened 2024-10-29
Avoid using deprecated method waitForTimeout

### LINKED ISSUE #3947: waitForTimeout was removed from puppeteer
https://github.com/puppeteer/puppeteer/pull/11780

jcoyne:
Fixes #3947

### DIFF
--- a/__tests__/integration/mirador/companion_windows.test.js
+++ b/__tests__/integration/mirador/companion_windows.test.js
@@ -6,7 +6,6 @@ describe('Companion Windows', () => {
     await expect(page).toFill('#manifestURL', 'http://127.0.0.1:4488/__tests__/fixtures/version-2/001.json');
     await expect(page).toClick('#fetchBtn');
     await expect(page).toClick('[data-manifestid="http://127.0.0.1:4488/__tests__/fixtures/version-2/001.json"] button');
-    await page.waitForTimeout(300);
     await expect(page).toMatchElement('.mirador-window');
   });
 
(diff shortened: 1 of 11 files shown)


marlo-longley:
Thanks @jcoyne -- locally I got 4 failing tests with the suggestion to add `testname-fn-timeout` instead. For example:
```
● Window Sidebars › renders canvas navigation and updates canvas after clicking a navigation item

    thrown: "Exceeded timeout of 10000 ms for a hook.
    Add a timeout value to this test to increase the timeout, if this is a long-running test. See https://jestjs.io/docs/api#testname-fn-timeout."

      2 |
      3 | describe('Window Sidebars', () => {
    > 4 |   beforeAll(async () => {
        |   ^
      5 |     await page.goto('http://127.0.0.1:4488/__tests__/integration/mirador/blank.html');
      6 |
      7 |     await expect(page).toClick('#addBtn');

      at beforeAll (__tests__/integration/mirador/window_sidebar.test.js:4:3)
      at Object.describe (__tests__/integration/mirador/window_sidebar.test.js:3:1)
      ```

Is there a reason to remove the wait time rather than replace it with the newer methods/syntax?

jcoyne:
@marlo-longley I don't understand what you mean by the newer syntax.  The method waitForTimeout is removed https://github.com/puppeteer/puppeteer/pull/11780

cbeer:
I'm not seeing any local failures from this change (although there's something funny about one of the plugin tests, unrelated to this.). I don't think any of the `waitForTimeout` calls are anything but trying to address flappy tests, but maybe the ecosystem has improved...

PR #3955, opened 2024-10-29
Stop using default props in BackgroundPluginArea

jcoyne:
defaultProps are deprecated for function components

### DIFF
--- a/src/components/BackgroundPluginArea.js
+++ b/src/components/BackgroundPluginArea.js
@@ -1,18 +1,13 @@
-import PropTypes from 'prop-types';
 import ns from '../config/css-ns';
 import { PluginHook } from './PluginHook';
+
+const BackgroundPluginAreaDefaultProps = {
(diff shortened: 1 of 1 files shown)


marlo-longley (commented):
Do we want to totally remove `propTypes` rather than leaving this?
```
BackgroundPluginArea.propTypes = {
  PluginComponents: PropTypes.array,
};
```

cbeer:
I'm not sure I understand why we're dropping propTypes entirely. I don't see anything upstream about deprecating those, and a trivial test of the upstream pattern seems to work fine?

 https://github.com/ProjectMirador/mirador/pull/3976

jcoyne:
@cbeer the prop-types library has been archived. It is no longer supported: https://github.com/facebook/prop-types

cbeer:
Fixed in https://github.com/ProjectMirador/mirador/pull/3976.

PR #3958, opened 2024-10-29
Disable react/prop-types eslint rule

jcoyne:
See https://github.com/jsx-eslint/eslint-plugin-react/issues/3753

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -43,6 +43,7 @@
       "ignoreRegExpLiterals": true
     }],
     "react/jsx-uses-react": "off",
+    "react/prop-types": "off",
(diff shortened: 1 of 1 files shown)


marlo-longley:
I am closing this as stale -- it targets .eslintrc, which no longer exists after the migration to flat-config ESLint.

The linked thread acknowledges that propType runtime enforcement is dropped in React 19, but they decide to keep the linting rule enforcement

Our codebase actively still enforces the linting rule. I think we should keep it until we figure out what we want to do about propTypes (Typescript?!) and follow ESlint's lead. This issue around propTypes enforcement is known and flagged for discussion on the dev team.

PR #3970, opened 2024-10-30
Remove unmaintined airbnb and flowtype eslint plugins

jcoyne:


cbeer:
👎 . Removing the airbnb defaults removes something like 2/3 of the style rules we want to enforce.

jcoyne:
@cbeer there's no indication that airbnb is going to continue to maintain it. It was last published 3 years ago. It's last commit was a year ago. Eslint version 8 is past it's EOL https://eslint.org/version-support/

marlo-longley:
I am running into this issue now. There is a comment from Nov 2024 saying the airbnb plugin is still being worked on?

https://github.com/airbnb/javascript/issues/2961

PR #3973, opened 2024-10-31
Convert some components to functional components

jcoyne:
Functional components are what all the documentation and tutorials on React show by default.  Keeping close to the default style reduces cognitive load on contributors.


marlo-longley (commented):
Thanks Justin. The conversion seems good, but is there a reason we want to also move from default to named exports? I am thinking that may cause complications for people to need to change their imports syntax.

jcoyne:
Oh, I hadn't considered that. I was just looking at the style used in the react docs (https://react.dev/learn) and copying from there.

marlo-longley:
My instinct is it might be best to keep existing export style (named vs default) while converting to functional.

PR #3975, opened 2024-11-06
Remove canvas.

cbeer:
This dependency was added to enable `<canvas>` support with the switch to jsdom, but it is seemingly unmaintained and just requires a couple tests change 🤷‍♂️


jcoyne:
👍

PR #3991, opened 2024-11-19
Add index.js files to src directories

cbeer:
Pushing most of the classes up to the top level should make downstream imports more robust (vs the current file-based `mirador/dist/es/*`imports).


lutzhelm:
One thing I noticed is that there is just a single default export in the index.js. As a result, this export always need to be imported entirely. Is that intentional?

E.g. instead of importing just the necessary items with a named import like this:
```
import { ScrollIndicatedDialogContent } from 'mirador';
```

... the items of interest now always have to be taken from the default import.
```
import Mirador from 'mirador';
const { ScrollIndicatedDialogContent } = Mirador;
```

It works fine, but I'm not sure if this can become cumbersome. (I might also overlook an easier way to handle this.)

cbeer:
Thanks, that wasn't my intent. I guess I should have been more suspicious of the existing code.

lutzhelm:
Redux state related things like `actions`, `selectors`, `reducers` still need to be imported en bloc, but I am not sure if it would be advisable expose every action, selector etc. individually via index.js.

I also stumbled upon some imports in the plugins that still need to be imported with the entire `dist/es/src` path, e.g. `mirador/dist/es/src/config/settings` or `mirador/dist/es/src/state/reducers/rootReducer`. They are only used to set up Context Providers for the tests. It might be possible to run those tests without those imports. And if they are actually needed, the full path imports are still available.

cbeer:
I've tweaked the state/index.js re-exports so maybe it's better now. There's some subtlety (maybe with default exports?) I'm not fully grasping. 

... and I've added those two as exports. 🤷‍♂️

cbeer:
Ok, try this? Our mess of default and non-default exports make this pretty confusing.

lutzhelm:
I have to update my findings for today:
* mirador-dl-plugin: works!
* mirador-textoverlay: state seems to work, but the controls don't show up
* mirador-image-tools: needs `withRef` and `withSize` references removed, but controls still don't show up

I was indeed not aware of the OSD changes, e.g. #4014; it could be that the plugins just have to be adapted to those changes.

PR #3993, opened 2024-11-20
Convert Workspace components to functions

cbeer:


marlo-longley (approved):
The conversion makes sense to me, and testing this branch with both elastic and mosaic layouts works as expected. Thanks for this work!

PR #3995, opened 2024-11-20
Convert CollectionDialog to a function.

cbeer:


marlo-longley (changes requested):
Hmm when I add this collection using the + UI and then click to view it, I get this

https://iiif.io/api/cookbook/recipe/0032-collection/collection.json

Screenshot 2024-11-27 at 7 32 44 PM

marlo-longley (dismissed):
My mistake please ignore.

PR #3997, opened 2024-11-20
Convert NewWindow to a function.

cbeer:


marlo-longley (commented):
Radical thought: should we rename this to be `NewBrowserWindow` ? 
It is only used once in the codebase as far as I can tell, in the authentication flow.

Given that "windows" are so central to Mirador's internal code, it could be less confusing (for example https://github.com/ProjectMirador/mirador/blob/3b04f56d00c17fb12e414e1a4ec905a3a3f1bb84/src/containers/Workspace.js#L19 -- this is totally unrelated)

marlo-longley (approved):

PR #3998, opened 2024-11-20
Convert IIIFAuthentication to a function.

cbeer:


marlo-longley (approved):
Looks good. I tested this with the internal `/auth.html` fixture page, and all 3 flows worked correctly.

PR #4006, opened 2024-11-22
Rewrite ScrollTo as a function.

cbeer:


__tests__/src/components/ScrollTo.test.js
```diff
@@ -1,85 +1,65 @@
...
+          <ScrollTo containerRef={containerRef} scrollTo><div data-testid="c" data-mockboundingrect={JSON.stringify({ bottom: 225, top: 150 })} style={{ height: 75 }} /></ScrollTo>
+        </div>
+      </div>,
+    );
 
-      it('does not scroll to the selected element when it is visible', () => {
-        jest.spyOn(ScrollTo.prototype, 'scrollToBoundingRect').mockImplementation(() => ({
-          ...visibleScrollToElBoundingRect,
-        }));
+    expect(Element.prototype.scrollTo).toHaveBeenCalled();
```
marlo-longley, line 47: Nice. Is there a way to test that it was actually called for `c`? Something like:
```
expect(Element.prototype.scrollTo).toHaveBeenCalledWith(0, -50);
```

src/components/ScrollTo.js
```diff
@@ -1,117 +1,57 @@
...
-    const { containerRef } = this.props;
+export function ScrollTo({
+  children, containerRef, offsetTop = 0, scrollTo, nodeId, ...otherProps
+}) {
+  const scrollToRef = useRef();
+  const prevScrollTo = usePrevious(scrollTo);
 
-    if (!containerRef || !containerRef.current) return {};
+  useEffect(() => {
+    if (!scrollTo || scrollTo === prevScrollTo) return;
```
marlo-longley, line 25: Could we add back some of the spirit of the deleted comment:
"If the scrollTo prop is true (and has been updated) scroll to the selected element"

It might just be me but I found it a little hard to first get that `prevScrollTo` and `scrollTo` were just the boolean prop and not any complex coordinates.

__tests__/src/components/ScrollTo.test.js
cbeer: Because jest/jsdom doesn't implement offsetHeight, the calculated value is totally wrong so I don't think we want to write a test for it 🤷‍♂️ (and because it's a property, I don't think there's a ~good~ bad way to mock it out like I did the other methods).

src/components/ScrollTo.js
```diff
@@ -1,117 +1,57 @@
...
+function usePrevious(value) {
+  const ref = useRef();
+  useEffect(() => {
+    ref.current = value;
+  }, [value]);
+  return ref.current;
+}
 
 /**
  * ScrollTo ~
```
cbeer, line 16:
```suggestion
 * ScrollTo ~
 * If the `scrollTo` prop is true (or when it is switched from `false` to `true`), scroll
 * the selected element into view
```

@marlo-longley ?

marlo-longley (approved):

Issue #3936, opened 2024-10-08
When searching it calls the search endpoint twice


fitnycdigitalinitiatives:
Hi,

I noticed this on our own Mirador instance and have confirmed it on the [mirador demo](https://mirador-dev.netlify.app/). If IIIF search is enabled and you run a search through the interface, you'll notice that it makes two identical calls to the search service at the same time, so I'm wondering if there's some minor bug in the code. It doesn't prevent it from working but does add unnecessary overhead to the search service. See in the screenshot, you can see the duplication in the dev panel for the Wellcome example in the demo (the first is for the autocomplete, but the next two are just identical).

![image](https://github.com/user-attachments/assets/adf371ae-5329-43a9-9d89-ccc862be1661)

Thanks,

Joe

fstoe:
One fetch is for the autocomplete, the other fetch is the actual fetching of the results. 
They look similar in the network console but have different request urls `https://iiif.wellcomecollection.org/search/v0/b18035723?q=the` vs `https://iiif.wellcomecollection.org/search/autocomplete/v1/b18035723?q=wei`

PR #4008, opened 2024-11-25
fix: fetchSearch

fstoe:
Hey,
i tried to fix this [issue](https://github.com/ProjectMirador/mirador/issues/3936).
It seems like the onSubmit sends two events. The first one being undefined and the second being the actual onSubmit one. This results in the dispatching of the `action fetchSearch` two times

### DIFF
--- a/__tests__/src/components/SearchPanelControls.test.js
+++ b/__tests__/src/components/SearchPanelControls.test.js
@@ -37,8 +37,22 @@ describe('SearchPanelControls', () => {
     await user.keyboard('somestring');
     await user.click(await screen.findByText('somestring 12345'));
     expect(fetchSearch).toHaveBeenCalledWith('window', 'cw', 'http://example.com/search?q=somestring+12345', 'somestring 12345');
-
     fetch.resetMocks();
+  });
+  it('should fetch result only once', async () => {
+    const fetchSearch = jest.fn();
+    const user = userEvent.setup();
+
+    createWrapper({
+      autocompleteService: { id: 'http://example.com/autocomplete' },
+      fetchSearch,
+      searchService: { id: 'http://example.com/search', options: { resource: { id: 'abc' } } },
+    });
+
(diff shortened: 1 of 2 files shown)


marlo-longley (changes requested):
Thank you for looking at this! I looked closer and noticed how the tests were failing. Here is what I came up with:

```
  submitSearch(event) {
    const {
      companionWindowId, fetchSearch, searchService, windowId,
    } = this.props;
    const { search } = this.state;
    if (!search) return;

    if (event && event.type === 'submit' && this.state.isSuggestionSelected) {
      event.preventDefault();
      return; // Do not trigger search again if a suggestion was selected
    }

    if (event) {
      event.preventDefault();
    }
    fetchSearch(windowId, companionWindowId, `${searchService.id}?${new URLSearchParams({ q: search })}`, search);
  }
  ```
  
  This approach requires adding something new to the component state (`isSuggestionSelected`), and also handling that elsewhere in the file (reset to false at the top of `handleChange`) and also in `selectItem`:
  ```
    selectItem(_event, selectedItem, _reason) {
    if (selectedItem && getMatch(selectedItem)) {
      this.setState({ search: getMatch(selectedItem), isSuggestionSelected: true }, this.submitSearch);
    }
  }
  ```

 That is just some ideas, it may not make sense overall but the tests are passing with that, and there are no double searches.

fstoe:
Ah thanks for the input for fixing the tests and updated the `SearchPanelControls` with you're suggestions

marlo-longley:
@fstoe nice! This seems to fix the double request bug for me and not break anything else, but I would be curious for someone else to test it out.

marlo-longley (dismissed):

fstoe:
> I'm curious if #3999 (rewriting these components to functions) fixes it or has the same problems.

I just tested but unfortunately the bug persists when using the functional component.

PR #4014, opened 2024-12-04
Extract components for handling OSD + tile sources.

cbeer:
This PR extracts components from `OpenSeadragonViewer` that manage the integration between OSD + react. By expressing the tile source as a react component, we're able to take advantage of react's internal rendering smarts instead of having to managing what may have changed.

### DIFF
--- a/src/components/OpenSeadragonComponent.js
+++ b/src/components/OpenSeadragonComponent.js
@@ -1,0 +1,189 @@
+import Openseadragon from 'openseadragon';
+import PropTypes from 'prop-types';
+import {
+  Children, useEffect, useId, useRef, useReducer,
+  useState, useCallback, cloneElement,
+} from 'react';
+import { useDebouncedCallback } from 'use-debounce';
+import { useTranslation } from 'react-i18next';
+import OpenSeadragonViewerContext from '../contexts/OpenSeadragonViewerContext';
+
+/** Handle setting up OSD for use in mirador + react */
+function OpenSeadragonComponent({
+  children = undefined, Container = 'div', osdConfig = {}, viewerConfig = {}, onUpdateViewport = () => {}, setViewer = () => {}, style = {}, ...passThruProps
+}) {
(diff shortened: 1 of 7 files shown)


src/components/OpenSeadragonComponent.js
```diff
@@ -0,0 +1,189 @@
...
+      const zoomRatio = (shift ? 1.0 / osdConfig.zoomPerDoubleClick : osdConfig.zoomPerDoubleClick);
+      viewer.viewport.zoomTo(currentZoom * zoomRatio, viewer.viewport.pointFromPixel(position), false);
+    });
+
+    viewer.addHandler('animation-finish', onViewportChange);
+    // @ts-expect-error
+    viewer.innerTracker.moveHandler = moveHandler;
+
+    viewerRef.current = viewer;
+    setViewer(viewer);
```
marlo-longley, line 101: I don't understand if setViewer is actually doing anything, since you are using a Context now.

marlo-longley:
This looks great. Thank you Chris! Small comments == 

1. needs a rebase

2. my main question which could be due to lack of understanding is -- if we can reduce the number of references to `viewer` and rely more on your new `OpenSeadragonViewerContext`. In my reading we have:
- `setViewer` in the parent component's state that gets passed down
- `apiRef` in the parent which is also holding a viewer and uses OSD methods like `fitBounds`
- the `viewerRef` in the child `OpenSeadragonComponent`
- also the `OpenSeadragonViewerContext`. 

Something to think about maybe for down the road and not this PR.

cbeer:
I think your analysis is good. I don't understand the difference between apiRef + viewer in the component state and why we may or may not need both (I think we're using it as a hack to force a re-render when we initialize the viewer?), and I think the `viewer` prop could be replaced with a context consumer eventually (although we send it through to plugins which may need to be updated too).

marlo-longley (approved):

Issue #3585, opened 2022-12-07
All elements of a Choice annotation are rendered simultaneously


irv:
When viewing a manifest that contains a canvas with a `Choice` annotation, Mirador will flatten the annotations and vacuum up all the image service annotations, passing them to open sea dragon and rendering them, even if there's no interaction with the layers view. If I have understood the code correctly, this seems to happen here: https://github.com/ProjectMirador/mirador/blob/5ca33205bf9bac636ba5ef0faf820f58c0d9751d/src/lib/MiradorCanvas.js#L76-L77 

This means if the user does open the layers view, it's very smooth when toggling the display of each layer because the tiles are loaded. But for instances where we have more then 2 layers, this is a lot of wasted image requests, especially if the user never interacts with them.

You can see this using just the cookbook recipe: https://projectmirador.org/embed/?iiif-content=https://iiif.io/api/cookbook/recipe/0033-choice/manifest.json

And looking at the network requests where it loads tiles for both the natural and x-ray version without any user interaction:

![image](https://user-images.githubusercontent.com/390869/206169978-b6226f53-2fc8-4315-a969-0f8333bcc1a0.png)

And this is obviously made worse with a larger viewport, as more tiles are requested at once...

I think a better approach would be to load the thumbnails for each `Choice` annotation, and then add the first element only (as this is supposed to be the default. See https://www.w3.org/TR/annotation-model/#choice-between-bodies). When interacting with the layers view, the additional tile sources can then be passed to open sea dragon.

lutzhelm:
Fixed via #4015.

PR #4015, opened 2024-12-05
Solve vertical display order and Choice behavior, fix #3693 and #3585

### LINKED ISSUE #3693: Choice in Mirador 3 does not display as expected and breaks from previous Mirador versions
Where a manifest uses Choice, M3 currently defaults to displaying all images in a choice stack at once (all are "on") and they are displayed in reverse order from what is expected. For a user, this means that in order to see a specific image in the stack, they must first turn off all of the layers to see the desired image. As well, the first image that appears in the manifest is at the top of the stack, meaning when this image is displayed and a user chooses to display another image, the original/first image must be turned off. The user interaction here should be additive, not subtractive.

The desired behavior would be to revert to how M2 handled choice: 
- Only the first painting annotation in a choice stack is enabled on load (this is inline with the spec)
- The presentation of images is upwards in a z-index from the first painting annotation encountered (this is necessary so that as new images are selected, they are visible without having to deselect the first image.

The current behavior breaks the functionality of choice manifests created following the IIIF Presentation spec (2 or 3) and used earlier with M2. Also, the layering of images from the top down mean that selected images are not shown.

You can see this using the cookbook recipe: https://projectmirador.org/embed/?iiif-content=https://iiif.io/api/cookbook/recipe/0033-choice/manifest.json

My particular use case is manuscripts with MSI. Each manuscript folio has 12-25 multi-spectral images in a given choice stack, so a user having to uncheck each layer and reordering them to be able to view a specific image is not a good user-experience.

This relates to another issue (https://github.com/ProjectMirador/mirador/issues/3585) that points out that the current M3 method results in many unnecessary image requests.

This is a major blocker for migrating from M2 to M3 since existing manifests do not work as expected.

Some examples:
- Sinai Palimpsests Project: https://sinai.library.ucla.edu | https://sinai-images.library.ucla.edu/iiif/ark:%2F21198%2Fz1kd1z25/manifest (this requires a login to view the images)
- https://prtd.app/#72f604db-6869-4c08-91ce-7c79502a7f35 | https://dvp.prtd.app/hamilton/manifest.json

This current M3 behavior also affects viewing when there are multiple resources layered on a canvas (without choice). Take for example the reconstruction of a manuscript folio: https://iiif.io/api/cookbook/recipe/0036-composition-from-multiple-images/. In this case, it does make sense to load all resources when the manifest is loaded into the viewer, but the order in which the resources are displayed is top to bottom, instead of bottom to top, meaning the subsequent resources are buried under the first resource and the user cannot view the "scene" as intended.

lutzhelm:
This PR demonstrates a way how #3693 and #3585 could be addressed. At the current state, it is more of a hack than an actual solution. The tests are not adjusted yet as I don't expect the behavior to stay like it is.

This addresses two problems:
- The less preferred options of a Choice are unnecessarily loaded and displayed from the start (#3585, #3693).
- The vertical image order is opposite to the order intended by the Image API authors (##3693).

At first I explored the possibility to solve the first problem via changes to the layers redux state, e.g. by setting all less preferred Choice option images invisible after adding a manifest to a window or after changing the current Canvas, but this would lead to extensive and error prone rework. The current "solution" does not require any changes to the state; instead it changes the default behavior. It is, however, not the cleanest way to do this. One problem is that currently, to my knowledge, manifesto does not indicate in any way if an image annotation is a Choice option, and if so, if it is the preferred one. There is a comment in the code that acknowledges that:

https://github.com/IIIF-Commons/manifesto/blob/63a230dfa0584ed71eeed956e1d0f134ec9b4bed/src/Annotation.ts#L25

We could try to make a suggestion for manifesto, but I'm not sure this will be ready before we want to release M4.

Another **issue for discussion** is the behavior of the Layer sidebar panel. Should images be ordered like they are in the annotations of the Canvas (current behavior, and also the behavior of Mirador 2), or would it be more comprehensible if they were ordered from top to bottom in the same way that they are ordered on a z-axis on the Canvas (behavior in the PR branch)? The current "Move layer to top" button would also make more sense for the latter behavior. It would have to be reshaped if the order in the panel stays as it is.

I'd appreciate any thoughts both on the layer panel order and on possible technical solutions with or without manifesto changes. I'd also be happy about more examples of v2 and v3 manifests with multiple images, Choices or both.


lutzhelm:
Thank you very much for your input!

I can see the rationale for sorting the images in the Layer controls by their significance within the manifest, but I am struggling to change the controls adequately. Both a "Move to top" or a "Move to bottom" button would be ambiguous as they could either refer to the stack on the Canvas or to the list in the layer controls; the effect on the respective other element would always be the opposite of what is actually happening in that element; moving an image to the top of the Canvas moves it to the bottom of the layers panel and vice versa.

One option to solve this would be something like "Move to background"; it would require replacing both the icon and all translations. However, it would deviate from the layering behavior e.g. of all image editors I can think of; they usually put the background layer at the bottom of a layers list. Would users understand what is happening?

The current example manifest include both variants; it has five image annotations, only one of which is a Choice. The full color default option and the four additional image annotations are loaded and displayed by default - instead of all 23 ( :open_mouth: ) image layers as before. The PR is already adding the image composition cookbook recipe example to the manifest list for the `layers.html`, but we should move it to the front as a second example - we would have least one v3 manifest displayed right away.

kirschbombe:
Maybe "Move to top of index"? And to your point about layering behaviors, I get what you mean re: editing apps, but the reverse is true in the context of the web, where the background image would appear first in the list (z-index behavior), so it's going to be relative I guess (and most users could likely not have experience with either context). I think I would try to avoid "top" and "bottom" as you suggested - unless you can be specific re: top of "what". I'll give it more thought.

lutzhelm:
I changed the behavior of the Layers panel to display the images in the panel in the order in the manifest. Currently, the "move to top" button has been replaced with "move to button". However, I wonder what the most useful option would actually be:
* move to background
* move to foreground - this would would make more sense because it makes an image visible without having to hide any other images
* both of those controls.

The changes are available [here](https://deploy-preview-4015--mirador-dev.netlify.app/__tests__/integration/mirador/layers.html).

cbeer:
> would it be more comprehensible if they were ordered from top to bottom in the same way that they are ordered on a z-axis on the Canvas

I've been playing with this the last couple days, and I think having the top layer in the list be the base layer in the viewer is confusing. The behavior here has some nice properties for some kinds of imagery (the base layer isn't buried, the layer order matches the manifest, and the order doesn't change from the current display), but I'm not sure it's worth it. Is it possible to switch the order in this PR, or is it better to ship this as an incremental improvement and try to reverse the layer list separately?

lutzhelm:
Thanks for having a look and playing around. My feelings about this are similar. It should be easy to reverse this within this PR; a previous working version already did this.

kirschbombe:
As someone who relies on layers and choice for many things, I disagree that the order should be reversed. For me at least, the concept of layers is that it is additive (and seems to match the use cases I've seen thus far), in that the first item (default) in a choice or a list of items is the preferred item and is the foundation that subsequent items on layered onto. Layers without choice would typically be used to construct or reconstruct a view from multiple items that are not necessarily aligned positionally on a canvas. If the layer is choice of items, then the only the default should display, which make the most sense at the top of the list of items in the viewer, with the subsequent item layering upward. That is how Mirador 2 handles this. I realize that this is complicated by the introduction of the "move to top/bottom" buttons, but feel the order in the viewer sidebar should maintain the order of the items. I don't think it's terribly difficult to understand the concept of building upwards (or outwards) toward the user. This is what users and manifest authors (at least the ones that I'm in conversation with) are already used.

lutzhelm:
Following the latest community call's suggestion, I've been looking at other implementations to see how they implement layer and choice controls. For the relevant cookbook recipes ([Composition from Multiple Images](https://iiif.io/api/cookbook/recipe/0036-composition-from-multiple-images/) and [Multiple Choice of Images in a Single View (Canvas)](https://iiif.io/api/cookbook/recipe/0033-choice/)), only Annona and Theseus are listed as conformant examples besides Mirador. I don't know if there is any more comprehensive or just different overview over supported features other than the [Cookbook viewer matrix](https://iiif.io/api/cookbook/recipe/matrix/).

Annona's controls order the image annotations for composition and the choice options in the same way that way appear in the manifest, both for [composition](https://ncsu-libraries.github.io/annona/tools/#/display?url=https://iiif.io/api/cookbook/recipe/0036-composition-from-multiple-images/manifest.json&viewtype=iiif-storyboard&settings=%7B%22fullpage%22%3Atrue%7D) and [choice](https://ncsu-libraries.github.io/annona/tools/#/display?url=https%3A%2F%2Fiiif.io%2Fapi%2Fcookbook%2Frecipe%2F0033-choice%2Fmanifest.json&viewtype=iiif-storyboard&manifesturl=&settings=%7B%22fullpage%22%3Atrue%7D).

Theseus does not offer controls for compositions (the API does not require those), and for [choices](https://theseusviewer.org/?iiif-content=https://iiif.io/api/cookbook/recipe/0033-choice/manifest.json), it offers radio boxes for each choice option, listed in the same order as in the choice.

Currently I'm thinking we should use the PR as it is right now because it is closest to the other solutions. If feedback primarily suggests to change it to the reverse order, we could still change this.

kirschbombe:
@lutzhelm, thank you for working on these updates! Do you know the timeline for the next release that will include these fixes?

lutzhelm:
@kirschbombe There is no exact timeline planned for the release. Yesterday, we agreed upon keeping this PR (among others) open for one more week to wait for feedback and to merge it afterwards. So if no additional issues surface while reviewing this, I would expect this to be part of another alpha release in, say, two weeks. We also talked about an actual stable release; there are not that many items left to be done. This might actually happen around March/April.

wlpotter:
Thanks for y'all's work on this! Just chiming in as a user and manifest author that works with layers and choice quite a lot. I agree with the decision to list the layer options in the side bar in the same order as they appear in the manifest. This is what I would expect as a manifest author, and aligns with how M2 handles choice.

I would also be concerned about the usability of having the default layer be put at the bottom of the sidebar list since this is the only layer that is visible on load (thanks for fixing that, by the way! 🎉 ). It is common for manifests I work with to have over a dozen layers, so the bottom layers do not appear in the sidebar list without scrolling. This could prove confusing and frustrating for users who would have to scroll all the way down to turn off or adjust the opacity of the default layer — which is oftentimes necessary to work with the non-default multispectral imaging layers.

Again, grateful for the work put in here! Just wanted to record a few comments in case this discussion comes up again in the future

PR #4018, opened 2024-12-05
Use the useContext hook instead of withWorkspaceContext HOC

cbeer:


marlo-longley (dismissed):
Nice. Much cleaner. Looks good!

marlo-longley (changes requested):
Looks like there is one more error in:

```
mirador/src/components/WorkspaceOptionsMenu.js
  19:32  error  'WorkspaceContext' is not defined  no-undef
```

marlo-longley (approved):

PR #4021, opened 2024-12-06
Convert withWidth from a HOC to a hook.

cbeer:


marlo-longley:
I was able to get the tests to pass by changing references to `startHere` in the tests to `Start Here`:

https://github.com/ProjectMirador/mirador/blob/acccddabcf5661e98c9c6834c2c15b9ff7fd13a5/__tests__/src/components/WorkspaceAddButton.test.js#L24
https://github.com/ProjectMirador/mirador/blob/acccddabcf5661e98c9c6834c2c15b9ff7fd13a5/__tests__/src/components/WorkspaceAddButton.test.js#L42

marlo-longley (approved):

PR #4027, opened 2024-12-10
have viewer update with xywh coordinates

dnoneill:


marlo-longley:
Hi Niqui -- 
The values in the new `data-xywh-coords` attribute seem significantly different than what is in the Mirador state. Is this what you would expect? For example, the x value in the screenshot is 1138 vs 14.

Screenshot 2024-12-10 at 6 56 28 PM

Also, what is the purpose of the `data-parent-window-id` attribute?

cbeer:
The mirador state is the center of the window, and Niqui's x/y coordinates are the top-left corner. That said, I'm not sure I understand the benefits of exposing this in the DOM vs using the redux state (or OSD viewer) directly. Is this a niche feature for https://github.com/sul-dlss/exhibits/issues/2604 (maybe?) or is there some general utility?

dnoneill:
I need the xywh coordinates to update the thumbnail view. So exhibits has the thumbnail that points the the IIIF image. In order to update that view, we need to be able to have the IIIF coordinates.

cbeer:
That seems like a use case pretty specific to Stanford's applications, and more appropriate as a local plugin? https://github.com/sul-dlss/sul-embed/pull/2279

PR #4029, opened 2024-12-11
Extract usePlugins for additive plugins.

cbeer:
Instead of relying on the withPlugins HOC to pass in the `add` plugins, we can use the `usePlugins` hook to do this instead. This is a little more explicit and hopefully easier to understand (and perhaps allows for multiple pluggable areas for a component?)


lutzhelm (commented):
Codecov complains about a missing line; I'm not sure how easy it would be to add a suitable test. Otherwise there's only the superfluous proptype.

lutzhelm (approved):

Issue #4023, opened 2024-12-09
Look into reducing the number of references to the OSD viewer


marlo-longley:
Look into whether we can reduce the number of references to `viewer` and rely more on your new `OpenSeadragonViewerContext`. In my reading we have:
- `setViewer` in the parent component's state that gets passed down
- `apiRef` in the parent which is also holding a viewer and uses OSD methods like `fitBounds`
- the `viewerRef` in the child `OpenSeadragonComponent`
- also the viewer held in `OpenSeadragonViewerContext` 

Something to think about (this also just helped me understand to write it out) -- and maybe for down the road not adding more work to this PR.

_Originally posted by @marlo-longley in https://github.com/ProjectMirador/mirador/issues/4014#issuecomment-2523524537_

marlo-longley:
Related to https://github.com/ProjectMirador/mirador/pull/4030 ?

marlo-longley:
I had some ideas around building in a hook for `useViewerInstance` and `useViewerOverlay`, that could help address cases such as:
https://github.com/ProjectMirador/mirador-annotations/blob/f2e80c7c8400f7802d4e89d88956ecdaa2546d64/src/AnnotationDrawing.js#L4

https://github.com/dbmdz/mirador-textoverlay/blob/31cfece4c0bf2cfb1d288d303c12452d9a4afbc2/src/components/MiradorTextOverlay.jsx#L224

Mirador would have to use those hooks internally as well.

PR #4030, opened 2024-12-11
Use a top-level context to track OSD viewers in the workspace instead…

cbeer:
… of the `OSDReferences` global.


marlo-longley:
Notes on this:

OSDReferences.get/.set — this still exists in parallel to the new Context -- external plugins are probably using it
This PR more of a speculative setup for future use, as no files within the repo are actually using the new OpenSeadragonWorkspaceContext

I think this was prompted by https://github.com/ProjectMirador/mirador/issues/4023 but I am going to close as stale and archive the idea.

PR #4033, opened 2024-12-12
Pass through all props to the plugins

cbeer:
the `PluginHook` depends on props that aren't declared or used in the actual component.


marlo-longley:
Would this PR replace https://github.com/ProjectMirador/mirador/pull/3991 ?

cbeer:
Nope, I just failed to rebase the relevant commit appropriately.

Issue #4016, opened 2024-12-05
Loading windows with Collections results error messages: "Too many re-renders", "cyclic object value"


lutzhelm:
(Just wanted to drop this here, might have a look at it myself later...)

Example Collection URLs:
- https://iiif.io/api/cookbook/recipe/0032-collection/collection.json
- https://wellcomelibrary.org/iiif/collection/b18031511
- https://www.e-codices.unifr.ch/metadata/iiif/collection.json

### Observed behavior
Configuring Mirador with a Collection as window resource or selecting a Collection from the resource list results in the following error messages. I suspect that this is caused by updated dependencies (that have not been pinned by a package-lock.json); I get similar errors if I try to load collections with fresh local installs of v3.2.0 through v 3.4.0.

UI error message:
```
Error
Too many re-renders. React limits the number of renders to prevent an infinite loop.
Call Stack
 renderWithHooks
  node_modules/react-dom/cjs/react-dom.development.js:2659:139
 mountIndeterminateComponent
  node_modules/react-dom/cjs/react-dom.development.js:3297:1445
 beginWork
  node_modules/react-dom/cjs/react-dom.development.js:3636:93
 callCallback
  node_modules/react-dom/cjs/react-dom.development.js:730:119
 invokeGuardedCallbackDev
  node_modules/react-dom/cjs/react-dom.development.js:750:45
 invokeGuardedCallback
  node_modules/react-dom/cjs/react-dom.development.js:771:126
 beginWork$1
  node_modules/react-dom/cjs/react-dom.development.js:4702:22
 performUnitOfWork
  node_modules/react-dom/cjs/react-dom.development.js:4518:150
 workLoopSync
  node_modules/react-dom/cjs/react-dom.development.js:4504:47
 renderRootSync
  node_modules/react-dom/cjs/react-dom.development.js:4500:159
```

Console messages:
```
Uncaught Error: Too many re-renders. React limits the number of renders to prevent an infinite loop.
    React 18
    Redux 21
    s http://localhost:4444/collections.html:3
    Redux 3
    addWindow webpack-internal:///./src/state/actions/window.js:33
    middleware Redux
    dispatch http://localhost:4444/collections.html:6
    key Redux
    handleOpenButtonClick webpack-internal:///./src/components/ManifestListItem.js:27
    React 23
    renderInto webpack-internal:///./src/lib/MiradorViewer.js:21
    MiradorViewer webpack-internal:///./src/lib/MiradorViewer.js:18
[...]

PR #4035, opened 2024-12-12
Remove forced re-render; it doesn't seem necessary any more.

cbeer:
Part of #4016.


marlo-longley (commented):
I tested this locally and it works to eliminate the `Converting circular structure to JSON` warning.
I used the `https://iiif.io/api/cookbook/recipe/0032-collection/collection.json` manifest to test this.

The only comment I have is to remove the unused `useReducer` import.

cbeer:
Huh. Funny eslint didn't complain about that.

marlo-longley (approved):

PR #4036, opened 2024-12-12
Don't let manifesto mutate the manifest json.

cbeer:
This is the other part of #4016. manifesto's been mutating the provided manifest to add its data structures. This isn't a big problem for normal manifests, but the collection representation has a cyclic reference.

### DIFF
--- a/src/state/selectors/manifests.js
+++ b/src/state/selectors/manifests.js
@@ -10,6 +10,7 @@ import { getConfig } from './config';
 /** */
 function createManifestoInstance(json, locale) {
   if (!json) return undefined;
+  // Use JSON stringify/parse to create a deep copy and prevent Manifesto from mutating the json
   const manifestoObject = Utils.parseManifest(JSON.parse(JSON.stringify(json)), locale ? { locale } : undefined);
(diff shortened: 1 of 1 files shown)


marlo-longley (dismissed):
This works locally for me.

My only suggestion is to add a comment otherwise this could be a tad cryptic.
Something like ... "Use JSON stringify/parse to create a deep copy and prevent Manifesto from mutating the json" ?

src/state/selectors/manifests.js
```diff
@@ -10,7 +10,7 @@ import { getConfig } from './config';
 /** */
 function createManifestoInstance(json, locale) {
   if (!json) return undefined;
-  const manifestoObject = Utils.parseManifest(json, locale ? { locale } : undefined);
+  const manifestoObject = Utils.parseManifest(JSON.parse(JSON.stringify(json)), locale ? { locale } : undefined);
```
cbeer, line 13:
```suggestion
  // Use JSON stringify/parse to create a deep copy and prevent Manifesto from mutating the json
  const manifestoObject = Utils.parseManifest(JSON.parse(JSON.stringify(json)), locale ? { locale } : undefined);
```

PR #3967, opened 2024-10-30
Remove eslint-config-react-app

jcoyne:
It is no longer supported by a maintainer.
Fixes #3966

PR #4038, opened 2024-12-13
Vite cleanup

marlo-longley:
This would go into the Vite branch, or main after it is merged.
This gets rid of a fair number of console warnings.
The eslint work is taken/inspired by #3967 but updated for this branch.


package.json
```diff
@@ -83,31 +84,23 @@
   },
   "devDependencies": {
     "@testing-library/dom": "^10.4.0",
-    "@testing-library/jest-dom": "^6.1.5",
```
cbeer, line 86: Despite the name, I think we can keep `@testing-library/jest-dom`:

https://testing-library.com/docs/svelte-testing-library/setup/

PR #4049, opened 2024-12-16
Update OpenSeadragonComponent to support 0 (e.g. false-y) rotation/x/…

cbeer:
…y/etc parameters


marlo-longley (approved):

marlo-longley:
Since we are supporting this feature in release-3.x is that needed too in that branch?

cbeer:
I think this was just a regression from extracting OpenSeadragonComponent?

Issue #3963, opened 2024-10-30
Address outdated dependencies


cbeer:
- [ ] MUI 6.x (likely waiting for Mirador 6.x?)
```
@mui/icons-material                        5.16.7           5.16.7          6.1.6  node_modules/@mui/icons-material               mirador
@mui/lab                          5.0.0-alpha.173  5.0.0-alpha.173  6.0.0-beta.14  node_modules/@mui/lab                          mirador
@mui/material                              5.16.7           5.16.7          6.1.6  node_modules/@mui/material                     mirador
@mui/utils                                 5.16.6           5.16.6          6.1.6  node_modules/@mui/utils                        mirador
@mui/x-tree-view                           6.17.0           6.17.0         7.22.0  node_modules/@mui/x-tree-view                  mirador
```

- [ ] eslint (waiting for a compatible eslint-config-airbnb)
```
@typescript-eslint/eslint-plugin           6.21.0           6.21.0         8.12.2  node_modules/@typescript-eslint/eslint-plugin  mirador
@typescript-eslint/parser                  6.21.0           6.21.0         8.12.2  node_modules/@typescript-eslint/parser         mirador
eslint                                     8.57.1           8.57.1         9.13.0  node_modules/eslint                            mirador
eslint-plugin-jest                         27.9.0           27.9.0         28.8.3  node_modules/eslint-plugin-jest                mirador
eslint-plugin-react-hooks                   4.6.2            4.6.2          5.0.0  node_modules/eslint-plugin-react-hooks         mirador
```

- [ ] chalk (5.x switched to ESM, needs some bundling work for no clear benefit.) 
```
chalk                                       4.1.2            4.1.2          5.3.0  node_modules/chalk                             mirador
```

- [ ] glob (11.x dropped node 18 support)
```
glob                                       10.4.5           10.4.5         11.0.0  node_modules/glob                              mirador
```

- [ ] normalize-url (5.x breaks support for browser usage) 
```
normalize-url                               4.5.1            4.5.1          8.0.1  node_modules/normalize-url                     mirador
```

- [ ] openseadragon https://github.com/ProjectMirador/mirador/pull/3929
```
openseadragon                               4.1.1            4.1.1          5.0.0  node_modules/openseadragon                     mirador
```

- [ ] uuid (?)
```
uuid                                        9.0.1            9.0.1         11.0.2  node_modules/uuid                              mirador
```

marlo-longley:
Audit 2026: 
[...]

PR #4053, opened 2024-12-17
Update normalize-url package

marlo-longley:
Part of #3963 

I see the issue noted with dropped browser support, but that comment was from 2020 and browsers may support the syntax by now -- I tested the files where this package is used and it seemed to correctly normalize -- am I missing something?

Screenshot 2024-12-17 at 9 33 16 AM

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "lodash": "^4.17.11",
     "manifesto.js": "^4.2.0",
     "merge-refs": "^1.3.0",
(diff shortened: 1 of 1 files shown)


cbeer:
It looks like they fixed safari support in 7.x, and [js-regexp-lookbehind](https://caniuse.com/js-regexp-lookbehind) is now supported (since Safari v16.4).. so maybe it's ok?

PR #4055, opened 2024-12-17
Use useId to generate unique ids for accessibility attributes.

cbeer:


src/components/ManifestRelatedLinks.js
```diff
@@ -28,18 +29,19 @@ export function ManifestRelatedLinks({
   ...rest
 }) {
   const { t } = useTranslation();
+  const titleId = useId();
+
   const pluginProps = {
     homepage, id, manifestUrl, related, renderings, seeAlso, t, ...rest,
```
marlo-longley, line 35: Is `id` used here anymore?

marlo-longley (approved):

PR #4074, opened 2025-01-07
Add forwardRef to CompanionWindow

cbeer:


marlo-longley:
This looks good. I just learned that React 19 will allow for `refs` to be passed to functional components, and `forwardRef` will be deprecated. So the purpose of this PR, silencing the warning about passing refs to functional components, won't be relevant on the next version of React.

> Starting in React 19, you can now access ref as a prop for function components:
https://react.dev/blog/2024/12/05/react-19#ref-as-a-prop

Anyway this PR seems important so that Mirador 4 doesn't have browser warnings. But won't be an issue down the line.

PR #4098, opened 2025-02-07
Re-export useTranslation for downstream consumers to share the react-…

cbeer:
…i18next provider that mirador initializes

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -9,6 +9,7 @@ export * from './extend';
 export * from './lib';
 export * from './plugins';
 export { default as settings } from './config/settings';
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -9,6 +9,7 @@ export * from './extend';
 export * from './lib';
 export * from './plugins';
 export { default as settings } from './config/settings';
+export { useTranslation } from 'react-i18next';
```
cbeer, line 12:
```suggestion
export { useTranslation, withTranslation } from 'react-i18next';
```

PR #4099, opened 2025-02-10
add translate hook to plugins

Marcel-Tronco:
Updating my mirador setup to the recent 4.0 alpha I ran into problems using the mirador-image-plugin, because it didn't get the translation function it expected as a prop `t`. This commit solved it for me locally.

Feel free to use this.

Best

### DIFF
--- a/src/components/PluginHook.js
+++ b/src/components/PluginHook.js
@@ -1,16 +1,19 @@
 import { forwardRef, isValidElement, cloneElement } from 'react';
+import { useTranslation } from 'react-i18next';
 
 /** Renders plugins */
 export const PluginHook = forwardRef((props, ref) => {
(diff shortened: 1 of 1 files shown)


cbeer:
Thanks for looking into this. @marlo-longley and I were talking about this problem over on #4095, and this seems like the option 2 I offered over there. We shipped #4098 (option 1) as an alternative to this that hopefully works better with hooks or more complicated plugins (otherwise you'd have to pass the plugin-provided `t` to any subcomponents, we think?).

This will take a little update in any plugins (e.g. https://github.com/ProjectMirador/mirador-dl-plugin/pull/82) but Mirador 4 already requires some updates so we'll just include them in the release notes soon.

Thanks again

PR #4113, opened 2025-02-25
Make MUI + react-i18next peer dependencies

cbeer:
React isn't sharing contexts across plugins + apps the way it used to (because webpack did something special? React 18 changed something?). In order to get shared contexts, Mirador (we believe) has to treat those as peer dependencies and require the installing app to provide them (like we're already doing for React etc).

... but that's slightly annoying for people who use the UMD build ("just render me a mirador"), so this PR also adds a separate Vite configuration that (should?) include all the peer dependencies so it can be used directly 🤷‍♂️ .

### DIFF
--- a/vite-umd.config.js
+++ b/vite-umd.config.js
@@ -1,0 +1,56 @@
+import { defineConfig } from 'vite';
+import react from '@vitejs/plugin-react';
+import fs from 'fs/promises';
+
+/**
+* Vite configuration
+*/
+export default defineConfig({
+  build: {
+    emptyOutDir: false,
+    lib: {
+      entry: './src/index.js',
+      fileName: (format) => (format === 'umd' ? 'mirador.min.js' : undefined),
+      formats: ['umd'],
+      name: 'Mirador',
+    },
+    rollupOptions: {
+      external: [
+        '__tests__/*',
+        '__mocks__/*',
+      ],
+      output: {
+        assetFileNames: 'mirador.[ext]',
+        exports: 'named',
+      },
+    },
+    sourcemap: true,
+  },
+  define: {
+    'process.env': {},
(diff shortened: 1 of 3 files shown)


vite-umd.config.js
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+* Vite configuration
+*/
+export default defineConfig({
+  build: {
+    emptyOutDir: false,
+    lib: {
+      entry: './src/index.js',
+      fileName: (format) => (format === 'umd' ? 'mirador.js' : undefined),
```
marlo-longley, line 13: Question/suggestion about clarifying the naming of these files and extensions. Are we mixing which version we consider "default"? `package.json` (with `type="module"`) and the filename `vite.config.js` imply that es is the default. But in `/dist`, `mirador.js` is actually the UMD build.

The Vite docs show this pattern:
```
$ vite build
building for production...
dist/my-lib.js      0.08 kB / gzip: 0.07 kB
dist/my-lib.umd.cjs 0.30 kB / gzip: 0.16 kB
Recommended package.json for your lib:

package.json (single entry)
{
  "name": "my-lib",
  "type": "module",
  "files": ["dist"],
  "main": "./dist/my-lib.umd.cjs",
  "module": "./dist/my-lib.js",
  "exports": {
    ".": {
      "import": "./dist/my-lib.js",
      "require": "./dist/my-lib.umd.cjs"
    }
  }
}
```

cbeer: I think we're just following the pattern from Mirador 3 (and it seems like an ok one... ES build users don't care what the artifact is called, but UMD users probably do?)

cbeer:
Merging this per today's committers meeting

PR #4117, opened 2025-02-27
Preserve viewport maybe

cbeer:


src/components/OpenSeadragonComponent.js
```diff
@@ -100,13 +107,38 @@ function OpenSeadragonComponent({
...
+
+    if (previousAspectRatio < 0.75 * newAspectRatio || previousAspectRatio > 1.25 * newAspectRatio) {
+      console.log("aspect ratio changed, updating viewport");
+      const rect = new Openseadragon.Rect(...worldBounds);
+      if (!rect.equals(viewport.getBounds())) {
+        viewport.fitBounds(rect, false);
+      }
+    }
+
+  }, [osdConfig, viewerConfig, worldBounds, viewerRef]);
```
barmintor, line 141: if `worldBounds` is an array, I think this effect is always going to run because the object reference will change. I think the approach is sound, but we'd need to unroll the array into integers, right?

marlo-longley:
Tracking in https://github.com/ProjectMirador/mirador/issues/4371 as possibly relevant

PR #4121, opened 2025-02-28
Rename components from .js to .jsx to appease vite (and fix HMR)

cbeer:


marlo-longley (approved):
This is probably more correct in any case; thank you for doing the conversion.

PR #4133, opened 2025-03-12
Handle multiple IIIF resource services; fixes #3553

### LINKED ISSUE #3553: Implementation of fetchInfoResponse is not complient with iiif
Like @stephenwf mentioned in https://github.com/ProjectMirador/mirador/issues/3552#issuecomment-1112335225

----

I think the code mentioned:

```
/**
 * fetchInfoResponse - action creator
 *
 * @param  {String} infoId
 * @memberof ActionCreators
 */
export function fetchInfoResponse({ imageId, imageResource, windowId }) {
  const imageService = imageResource && imageResource.getServices()[0];
  const infoId = (imageId || imageService.id);
  return requestInfoResponse(infoId, imageService, windowId);
}
```

is still prone to bugs. A completely valid Image resource for example, could break:

```
{
  "id": "https://example.org/path/image.jpg",
  "type": "Image",
  "service": [
     { "type": "Service", "profile": "my-dimensions", " .... " },
     { "type": "ImageService3", "profile": "level2", "...." }
  ]
}
```

As the code is not looking for an image service from the list, just the first service:

`imageResource.getServices()[0]`

I think Manifesto just surfaced the bug.

----

cbeer:


jcoyne:
link to #3553

lutzhelm (commented):
All changes are fine, but I wonder if the behavior of `getVisibleCanvasNonTiledResources` should be changed as well:
https://github.com/ProjectMirador/mirador/blob/a04410e1dc85e03c72b8ba7a8357ae765dd8288a/src/state/selectors/canvases.js#L219

Hypothetically, there might be resources with services non of which are IIIF image services.

src/state/selectors/canvases.js
```diff
@@ -216,7 +216,7 @@ export const getVisibleCanvasNonTiledResources = createSelector(
   ],
   (canvases, getMiradorCanvas) => flatten(canvases
     .map(canvas => getMiradorCanvas(canvas).imageResources))
-    .filter(resource => resource.getServices().length < 1),
+    .filter(resource => getIiifResourceImageService(resource).length < 1),
```
lutzhelm, line 219: should be `.filter(resource => !getIiifResourceImageService(resource)),`

marlo-longley (approved):

PR #4154, opened 2025-06-04
fix: prevent showCollection button from disappearing

fstoe:
Hey,
this is my idea to fix issue [4152 ](https://github.com/ProjectMirador/mirador/issues/4152).

PR #4155, opened 2025-06-18
Fix #4152, use separate state for collection info panel and collection selection dialog

### LINKED ISSUE #4152: Link to collection disappears when clicking "Show Collection"
I'm having trouble understanding the intended behavior of Mirador when it comes to browsing collections. The following steps seem like a bug to me.

Steps to reproduce:

1. Go to https://projectmirador.org/embed/?iiif-content=https://iiif.io/api/cookbook/recipe/0032-collection/collection.json
2. Click on "Show Collection"
3. Open the first item in the collection: "The Gulf Stream"
4. Click on the hamburger menu in the top-left to "Toggle Sidebar"
5. In the sidebar, there is a subsection for the collection: "Simple Collection Example" with a "Show Collection" button.
6. If you click on that "Show Collection" button, this subsection disappears from the sidebar.
7. If you click "Close" on the collection dialog in the center of the screen, you have no way of getting back to the collection. The "Simple Collection Example" that was previously present in the sidebar is now gone, and it's impossible to get it back.

I can't imagine why this would be intended behavior. Why does the "Simple Collection Example" disappear from the sidebar when you click on "Show Collection"? If you select an item from the collection in the center dialog, it will come back, but not if you click "Close", which puts you in a strange state where the collection is now inaccessible.

lutzhelm:
Replacement for PR #4154; addresses problems mentioned in [this comment](https://github.com/ProjectMirador/mirador/pull/4154#issuecomment-2949717883).

### DIFF
--- a/src/components/CollectionDialog.jsx
+++ b/src/components/CollectionDialog.jsx
@@ -76,7 +76,7 @@ export function CollectionDialog({
 
   /** */
   const hideDialog = () => {
-    hideCollectionDialog(windowId);
+    hideCollectionDialog([...collectionPath, manifestId], windowId);
   };
 
   /** */
--- a/src/components/CollectionDialog.jsx
+++ b/src/components/CollectionDialog.jsx
@@ -66,7 +66,7 @@ Placeholder.propTypes = {
  * a dialog providing the possibility to select the collection
  */
 export function CollectionDialog({
-  addWindow, collection = null, collectionPath = [], error = null, hideCollectionDialog,
+  addWindow, collection = null, dialogCollectionPath = [], error = null, hideCollectionDialog,
   isMultipart = false, manifest, manifestId, ready = false,
   setWorkspaceAddVisibility, showCollectionDialog, updateWindow, windowId = null,
 }) {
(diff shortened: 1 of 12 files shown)


src/components/CollectionDialog.jsx
```diff
@@ -97,10 +97,10 @@ export function CollectionDialog({
   const selectManifest = (m) => {
     if (windowId) {
       updateWindow(windowId, {
-        canvasId: null, collectionPath: [...collectionPath, manifestId], manifestId: m.id,
+        canvasId: null, collectionPath: [...dialogCollectionPath, manifestId], manifestId: m.id,
       });
     } else {
-      addWindow({ collectionPath: [...collectionPath, manifestId], manifestId: m.id });
+      addWindow({ collectionPath: [...dialogCollectionPath, manifestId], manifestId: m.id });
```
lutzhelm, line 103: I'm going to leave this line untested. It does not actually make a lot of sense that there would be no `windowId`. The only way this line is ever executed would be a plugin that reuses this component.

marlo-longley (dismissed):
This works for me, @lutzhelm. It addresses the issue found in PR #4154

marlo-longley (approved):

PR #4163, opened 2025-07-28
Simplify the directory structure for demo pages and integration tests

marlo-longley:
Not strictly needed but I find this a lot simpler than what we have currently. Great if others agree, but I won't hold up the release for it!

`__tests__/integration/mirador` becomes `__tests__/integration`
demo `.html` pages are moved out of the `__tests__/integration/mirador` directory because they are no longer needed during testing after recent work to decouple.


lutzhelm:
There is a [single line](https://github.com/ProjectMirador/mirador/blob/main/README.md?plain=1#L98) in the README about running single test files that should be changed.

PR #4170, opened 2025-08-25
Update i18next dependency

marlo-longley:
These tests will fail until #4168 is in.
But this is needed.


barmintor (approved):
This is the same conclusion we came to in our fork: https://github.com/cul/mirador/commit/74b187dfe1f63b1d06317a3b8966048186980fd7

PR #4125, opened 2025-03-04
Resume using the upstream react-mosaic-component package

cbeer:
This reverts commit 48320145dfe98fad99206bf2de3ed393414b5137.

PR #4175, opened 2025-09-10
Avoid React 19 deprecation warning by using library fork

marlo-longley:
See discussion: https://github.com/nomcopter/react-mosaic/issues/225

We will need to address this at some point. The fork is one approach. 
I decided it was important to do because over in the `mirador-integration` repo, this warning actually crashed the parcel dev server. This library was the root cause.

Screenshot 2025-09-10 at 1 59 58 PM


marlo-longley:
See if this actually fixes it?
https://github.com/ProjectMirador/mirador/pull/4125

PR #4196, opened 2025-10-15
Fix focus and navigation for ThumbnailNavigation

### LINKED ISSUE #4195: Keyboard pagination breaks after clicking a thumbnail
If you click on a thumbnail image (regardless of whether the thumb strip is on the bottom or right), and then try to use keyboard navigation for prev/next page, the page will advance and then immediately revert back to the page that you had clicked on.

https://github.com/user-attachments/assets/3f1231a1-3a19-4ee2-b514-4ee763dd1f5a

### Steps to Recreate
1. Use mouse or keyboard to click on / focus on the thumb nav pane (_not an individual thumb_)
2. Press `right` (if thumb pane is on bottom) or `down` (if thumb pane is at right)
3. Mirador successfully advances to next page
4. Use mouse and click on a thumbnail
5. Press `right` (if thumb pane is on bottom) or `down` (if thumb pane is at right)
6. Mirador only advances for a split second before reverting back to the page corresponding to the clicked thumbnail

This is related to #4145 but is not quite the same issue. It appears to currently be present in both Mirador 3 & 4.

marlo-longley:
Closes #4195

@seanaery does this help? Locally for me this fixed it and looks like it's fixed on the Netlify preview below. I think it was interfering with the `handleKeyDown` method in the parent `ThumbnailNavigation`.

I'll have to look at the a11y issues another time. Good catch!

### DIFF
--- a/src/components/ThumbnailNavigation.jsx
+++ b/src/components/ThumbnailNavigation.jsx
@@ -34,6 +34,15 @@ export function ThumbnailNavigation({
     gridRef.current?.scrollToItem(index, 'center');
   }, [canvasIndex, view]);
 
+  // Prevent loss of focus when navigating thumbnails
+  const paperRef = useRef(null);
+
+  useEffect(() => {
+    if (paperRef.current && document.activeElement !== paperRef.current) {
+      paperRef.current.focus();
+    }
+  }, [canvasIndex]);
+
   /** */
   const handleKeyDown = (e) => {
     let nextKey = 'ArrowRight';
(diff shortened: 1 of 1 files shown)


seanaery:
@marlo-longley Yes--that definitely fixes it! 🎉  Just checked out your branch and it is behaving how I'd expect it to. Thank you so much for the quick work on this one!

marlo-longley:
@seanaery give this a try?

lutzhelm (approved):

PR #4215, opened 2025-12-04
Safely access hitAnnotation within SearchHit component

### LINKED ISSUE #4214: SearchHit.js error
Image

After searching in the search, select a certain value based on the results, then switch to information, and then switch to search again. This problem will occur!

marlo-longley:
Fixes #4214

Small fix for a safe accessor. Now the component can render when the search state isn't there yet. Once hitAnnotation is defined the UI updates correctly with the hits. Please test to confirm this fixes your issue @wdongji and thank you for reporting.

### DIFF
--- a/src/containers/SearchHit.js
+++ b/src/containers/SearchHit.js
@@ -42,7 +42,7 @@ const mapStateToProps = (state, {
   const allAnnoIds = [annotationId, ...hit.annotations];
 
   return {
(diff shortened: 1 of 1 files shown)


geourjoa:
> Thanks @geourjoa! Do you have permission to approve and merge the PR?

You re welcome, no I have no right on Mirador repo

geourjoa (approved):

PR #4234, opened 2026-03-13
Upgrade react-window to v2 API

jcoyne:
Without doing this I get dependency errors when running `npm install`


jcoyne:
Thank you @geourjoa. I've updated that selector.

PR #4267, opened 2026-04-02
Replace react-image with our own ImageWithFallback

### LINKED ISSUE #4266: Implement placeholder without react-image dependency
It looks like react-image hasn't had a release in 3 years.  It's just getting dependency updates from a bot. https://github.com/mbrevda/react-image

Added in https://github.com/ProjectMirador/mirador/commit/35c78f0d78ab07d6c325fc2813d72a2351bcbff1

jcoyne:
react-image is no longer maintained and it's easy enough to write our own error handler

Fixes #4266

### DIFF
--- a/src/components/ImageWithFallback.jsx
+++ b/src/components/ImageWithFallback.jsx
@@ -1,0 +1,24 @@
+import { useState } from "react";
+import PropTypes from "prop-types";
+
+/**
+ * A plain img element that renders a fallback node when the image fails to load.
+ */
+export function ImageWithFallback({ alt, fallback, src, ...props }) {
+  const [hasError, setHasError] = useState(false);
+
+  if (hasError) return fallback;
+
+  return (
+    <img alt={alt} src={src} onError={() => setHasError(true)} {...props} />
+  );
+}
+
+ImageWithFallback.propTypes = {
+  /** Alt text for the underlying img element. */
+  alt: PropTypes.string.isRequired,
+  /** Node rendered in place of the image when it fails to load. */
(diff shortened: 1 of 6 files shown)


marlo-longley:
Looks good. My only question is about the prettier reformatting -- is this on purpose to match the settings that will go in https://github.com/ProjectMirador/mirador/pull/4263?

jcoyne:
@marlo-longley my editor has prettier installed.  It uses the defaults unless the application has a prettier config, then it will use that.  So if we can get #4263 merged, then it'll just be a reformat call to resolve.

src/components/ImageWithFallback.jsx
```diff
@@ -0,0 +1,22 @@
+import { useState } from 'react';
+import PropTypes from 'prop-types';
+
+/**
+ * A plain img element that renders a fallback node when the image fails to load.
+ */
+export function ImageWithFallback({ alt, fallback, src, ...props }) {
+  const [hasError, setHasError] = useState(false);
+
+  if (hasError) return fallback;
```
marlo-longley, line 10: I think you need to add:
``` 
useEffect(() => {
    setHasError(false);
  }, [src]);
```

hasError never resets if `src` changes, I think a later successful load would still be hidden behind the error

marlo-longley (changes requested):
Looking at one minor change (see comment)
Also, do we need to regenerate package-lock with the removal of the dependency?

marlo-longley:
Superseded by #4465

PR #4278, opened 2026-04-18
Commit package-lock.json

### LINKED ISSUE #4252: Add package-lock.json to Git?
Currently, the `package-lock.json` is not part of the Git repository. This makes the CI pipeline as well as anyone anyone who works with that Git repository on their local machine susceptible to certain attacks and vulnerabilities. If a new malicious version of a previously harmless package is released, anyone who runs an `npm install`on a freshly checked-out Mirador repo might download the new, malicious version. This does not happen if the `package-lock.json` is already present.

The cause for opening this issue is a current example: `axios` [has been compromised](https://www.stepsecurity.io/blog/axios-compromised-on-npm-malicious-versions-drop-remote-access-trojan). `axios^0.30.0` is a transitive dependency via `bundlewatch`. Anyone who ran `npm install` on the Mirador repo for the first time between 1:00 UTC and ca. 3:30 UTC on March 31st could have downloaded the malicious version and run the malicious post-install hook.

I'm not sure what the rationale is or was for keeping `package-lock.json` out of the Git repository - I'd be happy to hear about any reasons. But we should definitely consider adding it to the repo.

marlo-longley:
Discuss!

Closes #4252

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 dist/
 coverage/
 node_modules/
-package-lock.json
 *.log
 *.tgz
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,8891 @@
+{
+  "name": "mirador",
+  "version": "4.0.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "mirador",
+      "version": "4.0.0",
+      "license": "Apache-2.0",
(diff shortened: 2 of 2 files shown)


jcoyne (approved):
Yes! This allows reproducible builds.

lutzhelm:
I am also very much in favour. Should there be really significant arguments for keeping the file out of the repo, we can still move back.

PR #4280, opened 2026-04-23
Add missing export

morpheus-87:


lutzhelm:
@morpheus-87 Is this something that you need for plugin development? If so, we might want to get it into a release as soon as possible.

morpheus-87:
Correct, we would like to use the function `ns` in our plugins. Would be great to get it in a new release.

morpheus-87:
How about this change, @lutzhelm ?

lutzhelm (approved):

PR #4288, opened 2026-05-01
Rename test files that contain jsx syntax to have a jsx file suffix

jcoyne:
This supports a Vite 8 upgrade.


marlo-longley:
@jcoyne not sure if you are aware but I made a number of commits to your other vite 8 PR, including this change

https://github.com/ProjectMirador/mirador/pull/4260

jcoyne:
@marlo-longley yes, but I think it makes sense to make this change in it's own PR as it doesn't depend on any specific vite version.  It's good practice to have the file extension match the file type.

marlo-longley (approved):

PR #4289, opened 2026-05-01
Upgrade to eslint 10 and fix no-useless-assignment

jcoyne:


marlo-longley:
I was a little concerned about impacts from `.npmrc` on other packages but the risk seems low and we can deal with it if that comes up. I have a feeling this workaround will be in place for a while! Thanks Justin.

marlo-longley (approved):

PR #4318, opened 2026-06-18
Upgrade to react-i18next 17

jcoyne:


marlo-longley:
@jcoyne I rebased this, prettified it, and also ran `npm upgrade` so that the changes were recorded in package-lock. Since we both worked on it...I feel like if you approve I'd be fine with that. Thanks.

PR #4319, opened 2026-06-19
Support percent-based media fragment selectors for annotations

nakamura196:
## Summary

[W3C Media Fragments](https://www.w3.org/TR/media-frags/#naming-space) allow the spatial dimension to be expressed as a percentage — `xywh=percent:x,y,w,h` — in addition to the bare `xywh=x,y,w,h` (pixel) form and the explicit `xywh=pixel:...` form. Mirador's annotation target parsers only handled the bare form: they captured everything after `xywh=` and ran `parseInt` over it.

As a result, a target like `#xywh=percent:25,25,50,50` produced `parseInt('percent:25', 10)` → `NaN` for the first coordinate, and the remaining values were treated as raw pixels with no scaling. Percent-based annotations therefore rendered in the wrong place, or not at all.

## Changes

- Add a shared `parseFragmentSelector` helper that strips the optional `pixel:` / `percent:` prefix and reports the `unit` alongside the parsed dimensions.
- Use it in the three parsing sites that previously duplicated the regex/`parseInt` logic: `AnnotationItem`, `AnnotationResource`, and `MiradorCanvas.onFragment`.
- Scale `percent:` values against the canvas dimensions where those are known:
  - `MiradorCanvas.onFragment` resolves against its own canvas size.
  - The annotation overlay drawing (`CanvasAnnotationDisplay`) and hit-testing (`AnnotationsOverlay`) resolve against the canvas world coordinates.

Bare and explicit `pixel:` fragments are unchanged in behaviour.

## Tests

- New `parseFragmentSelector` unit tests.
- Added percent cases to `AnnotationItem`, `AnnotationResource`, `MiradorCanvas`, and `CanvasAnnotationDisplay` tests, plus `fragmentUnit` coverage.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/__tests__/src/lib/AnnotationItem.test.js
+++ b/__tests__/src/lib/AnnotationItem.test.js
@@ -138,6 +138,23 @@ describe('AnnotationItem', () => {
       expect(new AnnotationItem({ target: 'www.example.com' })
         .fragmentSelector).toEqual(null);
     });
+    it('percent-based fragment', () => {
+      expect(new AnnotationItem({ target: 'www.example.com/#xywh=percent:25,25,50,50' })
+        .fragmentSelector).toEqual([25, 25, 50, 50]);
+    });
+  });
+  describe('fragmentUnit', () => {
+    it('defaults to pixel', () => {
+      expect(new AnnotationItem({ target: 'www.example.com/#xywh=10,10,100,200' })
+        .fragmentUnit).toEqual('pixel');
+    });
+    it('detects percent', () => {
+      expect(new AnnotationItem({ target: 'www.example.com/#xywh=percent:25,25,50,50' })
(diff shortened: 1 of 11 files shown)


marlo-longley:
Hi @nakamura196, thank you for this contribution, it looks good to me.

1. Are you able to rebase? Sorry, but since you opened this PR we make some formatting changes but adding prettier. `npm run format` should fix formatting issues.
 
2. Do you have an example manifest that uses percent fragments? I love the thorough testing but I'd like to click around in the viewer if possible.

marlo-longley:
Closed in favor of https://github.com/ProjectMirador/mirador/pull/4383

PR #4324, opened 2026-06-19
Upgrade i18next to v26

jcoyne:


marlo-longley:
I think this would make #4323 irrelevant because looks like they removed support notices in 26
https://www.locize.com/docs/general-questions/why-was-there-a-support-notice-for-i18next/

jcoyne:
@marlo-longley yes, that sounds right.

jcoyne:
@marlo-longley I'm still seeing that message in this build. It looks like both PRs will be necessary: https://github.com/ProjectMirador/mirador/actions/runs/27845531158/job/82414369758?pr=4324#step:5:158

marlo-longley (approved):

PR #4333, opened 2026-07-01
Avoid barrel imports of MUI components

jcoyne:
See https://mui.com/material-ui/guides/minimizing-bundle-size/\#avoid-barrel-imports


eslint.config.js
```diff
@@ -95,6 +95,12 @@ export default [
         { name: 'isFinite', message: 'Use Number.isFinite instead' },
         { name: 'isNaN', message: 'Use Number.isNaN instead' },
       ],
+      'no-restricted-imports': [
+        'error',
+        {
+          patterns: [{ regex: '^@mui/[^/]+$' }],
```
marlo-longley, line 101:
```suggestion
          patterns: [
          {
            // Block barrel imports from Material UI and Icons 
            regex: '^@mui/(material|icons-material)$', 
            message: 'Please use direct sub-module component imports (e.g., @mui/material/Button)
          },
    ],
```

marlo-longley: I am wondering if we should scope this to material and icons-material. Then you don't have to do the disable rules on `import { visuallyHidden } from '@mui/utils'; // eslint-disable-line`, or on any future imports from `@mui/utils`, which doesn't provide non-barrel imports.

marlo-longley (changes requested):

marlo-longley:
Also, needs a rebase on `PluginHook`

Thanks @jcoyne

PR #4335, opened 2026-07-02
fix: Encode default and suggested search queries.

jbaiter:
Suggested searches in the search panel and the default search query for the companion window were passed unescaped to the content search service, which caused issues when the query contained special characters (like `^` for boosting).


marlo-longley (approved):

marlo-longley:
Thank you @jbaiter !

Issue #3590, opened 2023-01-11
Cannot copy annotation text


Ludovic99:
Annotations are highlighted on hovering, in a synchronized way, both on canvas and on annotation panel. It is the same on click, with bolder and more persistent highlighting. Fine. But the annotation text cant' be selected to be copied, which is expecially frustrating for full-text or transcriptions annotations. It would be fine to enable that.

ggeisler:
I don't know how technically challenging this might be, but one idea is:

- When an annotation is selected, either via the canvas or the sidebar, display a copy icon at the top-right portion of the selected annotation:

<img width="866" alt="Screen Shot 2023-01-26 at 5 08 17 PM" src="https://user-images.githubusercontent.com/101482/214979066-c25a0b2d-7bbc-4977-bbe6-cf011a2a85cc.png">

---

If the user hovers over the icon, follow the same pattern we use for other icons to display a tooltip:

<img width="866" alt="Screen Shot 2023-01-26 at 5 09 08 PM" src="https://user-images.githubusercontent.com/101482/214979276-4090cfe6-35d6-41bc-bb50-3481c7e8e2b8.png">

---

If the user selects the icon, copy the annotation text to the clipboard and display a "Copied to clipboard" (or maybe, `Annotation copied to clipboard`) confirmation message ([MUI/React snackbar](https://mui.com/material-ui/)):

<img width="866" alt="Screen Shot 2023-01-26 at 5 09 50 PM" src="https://user-images.githubusercontent.com/101482/214979476-71ede27a-4147-4547-a5e8-2704f705b1a0.png">

---

I'd suggest autohiding the snackbar after ~5 seconds so they don't have to close it manually.

### Possible challenges
I'm guessing one challenge might be grabbing the annotation text to copy? I don't know enough about the technical details to know.

The other challenge might be that we'd need to adjust our current underlying structure of the annotations. Currently it appears that annotations in the sidebar are rendered in a `ul`, with each annotation an `li` element. When a user wants to select an annotation, the entire `li` is part of the selection, which means displaying the copy icon within the annotation container and also enabling the user to select the icon separately isn't possible.

But perhaps we can adjust the structure of the annotation `li` slightly to deal with this. This is an idea that might not be feasible (it's hard for me to test changes like this to Mirador (because of React, I guess) using the Web Inspector), but a general approach might be:

- Make the selected portion of the annotation the main `div` _within_ the `li`, rather than the `li` itself.
[...]

PR #4337, opened 2026-07-06
Supporting copying of selected annotation text

abcdavid:
A copy button appears for selected annotation text in the sidebar.  When clicked it copies the annotation text to the clipboard and shows a tick for two seconds.  Addressing the issue raised here:

https://github.com/ProjectMirador/mirador/issues/3590

### DIFF
--- a/src/components/CanvasAnnotations.jsx
+++ b/src/components/CanvasAnnotations.jsx
@@ -1,5 +1,6 @@
-import { useCallback } from 'react';
+import { useCallback, useState } from 'react';
 import PropTypes from 'prop-types';
+import Box from '@mui/material/Box';
 import Chip from '@mui/material/Chip';
 import MenuList from '@mui/material/MenuList';
 import MenuItem from '@mui/material/MenuItem';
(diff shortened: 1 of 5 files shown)


marlo-longley:
@abcdavid thank you for this contribution. We are glad you made this PR.

It seems very strange (well not that strange because we had some redundant issue tracking) that two PRs would come in the same week to address this issue, but this is what happened! 
See https://github.com/ProjectMirador/mirador/pull/4353

I need to review both PRs to compare -- I should be able to that this week.

src/components/CanvasAnnotations.jsx
```diff
@@ -50,6 +76,16 @@ export function CanvasAnnotations({
     hoverAnnotation(windowId, []);
   }, [hoverAnnotation, windowId]);
 
+  const handleCopyClick = useCallback(
+    (annotation) => (e) => {
+      e.stopPropagation();
+      copy(annotation.content);
```
marlo-longley, line 82: Does this copy raw HTML if it's embedded, rather than the displayed text?

src/components/CanvasAnnotations.jsx
```diff
@@ -50,6 +76,16 @@ export function CanvasAnnotations({
...
+      setCopiedAnnotationId(annotation.id);
+      setTimeout(() => setCopiedAnnotationId(null), annotationCopyConfirmDuration);
```
marlo-longley, line 84: I think every click schedules a new timer, and nothing ever cancels an old one?

I am thinking some check like this earlier in the method?
```
  if (copyTimeoutRef.current) {
      clearTimeout(copyTimeoutRef.current);
    }
```

marlo-longley (changes requested):

marlo-longley:
Thank you for your contribution @abcdavid -- this PR is being closed in favor of https://github.com/ProjectMirador/mirador/pull/4353, which does similar work.

Issue #3971, opened 2024-10-31
Inflight dependency has security error


jcoyne:
https://security.snyk.io/vuln/SNYK-JS-INFLIGHT-6095116

Inflight is no longer maintained.
it is pulled in by glob 7.2.3
which is used by jest-config, jest-runtime, rimraf (via canvas), test-exclude, @babel/cli, @jest/reporters, babel-plugin-lodash, bundlewatch

marlo-longley:
Closed by https://github.com/ProjectMirador/mirador/pull/4341

PR #4341, opened 2026-07-09
Replace bundlewatch with size-limit

jcoyne:
Bundlewatch was not update and causes CVE-2025-62718 due to a dependency on an old version of axios


marlo-longley (approved):

marlo-longley:
Closes https://github.com/ProjectMirador/mirador/issues/3971 by removing its parent bundlewatch

PR #4343, opened 2026-07-09
Generate language labels for each locale code

### LINKED ISSUE #2475: Generate language labels for each locale code [i18n
Split out from work in #2250 / #1877
This ticket is primarily concerned with the following scenario: A manifest provides metadata in a language that is NOT provided by their local Mirador 3 implementer in their [i18n keys](https://github.com/ProjectMirador/mirador/blob/73c7984c5ff4f6e984692d880ad25e059b5d8a03/src/config/settings.js#L125-L128).

IIIF manifests with multilingual properties `MUST` provide keys that use [RFC 5646 language codes](https://tools.ietf.org/html/rfc5646). It would be nice to provide the full language name in our interface (e.g. `de` -> `Deutsch`).

**Bonus**: check whether they key is provided in settings.js before relying on some external service

## Example issue from a proposed interface
i18n-picker displays 2 letter language codes

marlo-longley:
Closes #2475
This browser API of `Intl` wasn't available before. This covers the case where the language isn't already in our settings config, such as Finnish.

### DIFF
--- a/src/components/LocalePicker.jsx
+++ b/src/components/LocalePicker.jsx
@@ -31,6 +31,9 @@ export function LocalePicker({ availableLocales = [], locale = '', setLocale = u
     }
   };
 
+  // Resolve the label shown for a locale: manifest-provided label, then Intl.DisplayNames, then raw code
+  const getLocaleLabel = (code) => labels[code]?.label || getDisplayName(code) || code;
+
   return (
     <FormControl variant="standard">
       <InputLabel>{t('language')}</InputLabel>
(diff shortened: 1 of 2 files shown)


src/components/LocalePicker.jsx
```diff
@@ -41,7 +51,7 @@ export function LocalePicker({ availableLocales = [], locale = '', setLocale = u
       >
         {availableLocales.map((l) => (
           <MenuItem key={l} value={l}>
-            <Typography variant="body2">{labels[l]?.label || l}</Typography>
+            <Typography variant="body2">{labels[l]?.label || getDisplayName(l) || l}</Typography>
```
jcoyne: Can we move this all into a function call (and out of the template)?

marlo-longley:
@jcoyne extracted `getLocaleLabel`

dnoneill:
This seems stale? I am not quite sure why the drop down disappeared from below "About this item" but its no longer there and I think it should go back but I am not quite sure how to review the PR which doesn't seem to have something for me to review visually. Because it is using availableLanguages and it is a Object I am not quite sure how this is ever tripped.

dnoneill:
In order to review you need to go into settings and set showLocalePicker to true.

dnoneill (approved):

Issue #4342, opened 2026-07-09
Evaluate using react-compiler


marlo-longley:
React Compiler (stable as of React 19) automatically memoizes components and values, removing the need for manual useMemo, useCallback, and memo() throughout the code.

This idea was inspired by auditing this old draft PR: https://github.com/ProjectMirador/mirador/pull/2365 -- Mirador has widespread use of inline arrow function props. The React Compiler would be a modern solution to this same concern. 

Apparently you can run:
`npx react-compiler-healthcheck`

https://react.dev/learn/react-compiler

Payoff would potentially be better performance, but definitely cleaner code if we can remove those manual hook calls.

PR #4344, opened 2026-07-09
Add react-compiler and configure it to run in annotation mode

jcoyne:
Ref #4342


marlo-longley:
Nice. My idea on a quick triage reading of https://github.com/ProjectMirador/mirador/issues/4342 was that we could potentially remove a lot of code `useMemo` from the repo. Thoughts on that? I haven't researched it.

jcoyne:
It's my understanding that React Compiler can make many performance-only useMemo/useCallback calls redundant, but it won't work for "semantic identity/lifecycle memo".  Furthermore this will require any downstream consumers to also adopt React Compiler.  As a complication, we have external libraries like OpenSeadragon, and if we need any references we pass to those libraries, we have to do it manually.

Issue #3930, opened 2024-08-27
Undefined companionWindowIds property of window on updateWindow


barmintor:
See #3437

I'm working on a plugin that updates the window from an interaction in the sidebar. When the update is processed, a selector (which is somewhat obscured by the intermediating software) triggers a similar error - namely, that `companionWindowIds` is undefined on the window (which does exist), and thus the call to `companionWindowIds.map(...)` fails.

This is happening on a fork of M3, but as far as I can tell the relevant code is unchanged in M4. I'm suspicious that there's a race condition here. It might also be something in my plugin, but I don't want to lose this thread.

marlo-longley:
I am triaging this ticket
https://github.com/ProjectMirador/mirador/pull/4368 is an attempt to prove the failure mode

Developer todo:
Pick up #4368 and fix it

PR #4352, opened 2026-07-14
Don't re-create removed windows on UPDATE_WINDOW

alb3rtino:
## Purpose

When an `UPDATE_WINDOW` action arrives for a window that has already been removed from the state (e.g. a deferred `updateWindow` dispatch from a plugin resolving a fetch after the user closed the window), the reducer re-creates the window as a stub containing only the action payload. Such a ghost window has no `companionWindowIds`, which makes `getCompanionWindowsByWindowAndPosition` throw for every window in the workspace and breaks rendering:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'map')
    at src/state/selectors/companionWindows.js:79:40
    at Array.reduce (<anonymous>)
    at src/state/selectors/companionWindows.js:75:32
```

## Approach

In the `UPDATE_WINDOW` case of the windows reducer, return the state unchanged if `state[action.id]` does not exist, mirroring the existing guard on `SET_CANVAS`. A reducer test covers this case.

### DIFF
--- a/__tests__/src/reducers/windows.test.js
+++ b/__tests__/src/reducers/windows.test.js
@@ -165,6 +165,22 @@ describe('windows reducer', () => {
       };
       expect(windowsReducer(beforeState, action)).toEqual(expectedState);
     });
+
+    it('does not re-create a window that no longer exists', () => {
+      const action = {
+        id: 'gone',
+        payload: {
+          foo: 11,
+        },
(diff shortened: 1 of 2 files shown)


marlo-longley (approved):

marlo-longley:
Ref https://github.com/ProjectMirador/mirador/issues/3930

PR #4379, opened 2026-07-20
Remove redundant window title announcement

### LINKED ISSUE #3904: Redundant title read by screen reader
### Description
In the window body, voiceover reads the title aloud twice in very close proximity to one another. 

The areas include:
- A `<section>` has aria-label that reads “Window: TITLE”, so the title is describing the window region of the page (file: `src/components/Window.js`)
- The navigation nested directly below this section, `<nav arial-label=“Window navigation”>`, includes the title as an `<h2>` and gets read (file: `src/components/WindowTopBarTitle.js`)

Note: These come very close to one another, the only things that get read in between are “Window navigation" and "toggle sidebar button".

### Work-shopping solutions
While technically not an accessibility issue, it’s certainly a usability concern to those using screen readers since it is likely cumbersome to constantly hear the title (especially long ones). Things to think about
* Perhaps the section could be called something more generic, like “Item window”, though I can see how it's helpful to know the title right away when you have multiple images/workspaces open
* Consider hiding one from being read, but make sure that the heading is still present in the rotor, as this is a common way people navigate through pages 
   * It's actually quite confusing to navigate multiple workspaces using the rotor, there's not always a clear indication of what workspace you're interacting with
   * Would love to know what native users think, assuming it's along the lines of what @brittnylapierre has reported and that there's "not enough screen reader feedback"

AysajanE:
## Summary

- remove the window section's redundant `aria-label`
- keep the manifest title available as an `h2` for heading and rotor navigation
- update affected tests to use the retained heading or window structure instead of the removed named region

This prevents screen readers from announcing the window title twice in close succession.

Fixes #3904.

## Testing

- focused component suite — 23 passed
- affected integration tests — 3 passed, 1 pre-existing test skipped
- ESLint and Prettier checks for all changed files — passed
- repository-wide lint reports the same pre-existing formatting error in unchanged `setupTest.js:36` on both the exact base and patched trees

## Limitation

No assistive-technology session was run; the regression asserts the accessible DOM contract while retaining the heading.

---
AI assistance was used. This change was reviewed by Northset, and I accept responsibility for this submission.

### Verification

[Northset proof-of-pass receipt M-1051](https://northset-oss.github.io/verification-pilot/receipts/M-1051/)  
Contributor self-run; not maintainer verification.

### DIFF
--- a/__tests__/integration/tests/minimalist.test.js
+++ b/__tests__/integration/tests/minimalist.test.js
@@ -8,8 +8,9 @@ describe('Minimalist configuration to Mirador', () => {
 
   it('Loads a manifest and displays it without some of the default controls', async () => {
     expect(
-      await screen.findByRole('region', {
-        name: /Window: Cambridge, Corpus Christi College, MS 640: Antiphoner Leaf/i,
+      await screen.findByRole('heading', {
+        level: 2,
+        name: /Cambridge, Corpus Christi College, MS 640: Antiphoner Leaf/i,
       }),
     ).toBeInTheDocument();
 
--- a/__tests__/integration/tests/thumbnail-navigation.test.js
+++ b/__tests__/integration/tests/thumbnail-navigation.test.js
@@ -10,11 +10,14 @@ describe('Canvas navigation by clicking thumbnails', () => {
 
   it.skip('navigates a manifest using thumbnail navigation', async (context) => {
     // Make sure we have the manifest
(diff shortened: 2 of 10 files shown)


marlo-longley:
Closing due to the PR being abandoned, and also:
Deleting the `<section>`'s aria-label outright is not the right approach I don't think. No name means no landmark, so the fix removes the window's landmark entirely, leaving heading-based navigation as the only way to find it.

PR #4383, opened 2026-07-21
Rebased and reformatted: support percent-based fragment annotations

marlo-longley:
- **Support percent-based media fragment selectors for annotations**
- **Reformat files related to percent based fragment annotations**

@nakamura196 please take a look. I rebased and reformatted your PR, but I don't want to have messed up any of the logic. Thank you for work and well-scoped PR. We are auditing all our open PRs and trying to move forward. Thanks!

Supersedes https://github.com/ProjectMirador/mirador/pull/4319 if accepted

### DIFF
--- a/src/components/AnnotationsOverlay.jsx
+++ b/src/components/AnnotationsOverlay.jsx
@@ -11,8 +11,7 @@ import CanvasAnnotationDisplay from '../lib/CanvasAnnotationDisplay';
 
 /** @private */
 function isAnnotationAtPoint(canvasWorld, osdCanvasOverlay, resource, canvas, point) {
-  const [canvasX, canvasY, canvasWidth, canvasHeight] = canvasWorld
-    .canvasToWorldCoordinates(canvas.id);
+  const [canvasX, canvasY, canvasWidth, canvasHeight] = canvasWorld.canvasToWorldCoordinates(canvas.id);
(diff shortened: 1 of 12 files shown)


src/components/AnnotationsOverlay.jsx
```diff
@@ -22,7 +22,13 @@ function isAnnotationAtPoint(canvasWorld, osdCanvasOverlay, resource, canvas, po
   }
 
   if (resource.fragmentSelector) {
-    const [x, y, w, h] = resource.fragmentSelector;
+    let [x, y, w, h] = resource.fragmentSelector;
+    if (resource.fragmentUnit === 'percent') {
```
dnoneill, line 26: I am not quite sure this is the correct unit. The spec seems to indicate it should be pct:
https://iiif.io/api/image/3.0/#41-region, https://iiif.io/api/registry/selectors/

dnoneill: Unit is correct.

dnoneill:
I have created an alternative. I am not sure this is actually correct. My understanding is we should be using the canvas dimensions for the percentage calculation. https://github.com/ProjectMirador/mirador/pull/4499 does that.

marlo-longley:
Superseded by #4499

PR #4392, opened 2026-07-23
Give CompanionWindow's aside role="region" instead of the implicit complementary landmark

### LINKED ISSUE #4225: Aside is contained within main landmark
The sliding panels for "about this item", "rights", and "index" use the HTML landmark `<aside>`. 

However, `aside` needs to be at the top level, rather than inside `main`. 

Impact: **moderate**

https://www.w3.org/WAI/ARIA/apg/practices/landmark-regions/

https://www.accessibilitychecker.org/wcag-guides/ensure-the-complementary-landmark-or-aside-is-at-the-top-level/

Narasimha2211:
## Summary
Closes #4225.

The sliding panels for "about this item", "rights", and "index" use `<aside>`, which needs to be at the top level per WAI-ARIA landmark rules, but is nested several component layers inside `<main>` (`CompanionWindow.jsx` -> ... -> `WorkspaceArea.jsx`'s `<main>`).

## Approach
The issue thread had two suggestions from maintainers after it was opened:
- @jcoyne: assign `role="region"` to the `<aside>` tag
- @lhenze: or use a different HTML tag (`section`, `div`)
- @marlo-longley: "Developer action: chose one of the ideas above and implement."

I went with `role="region"` rather than physically hoisting the `<aside>` out of `<main>` in the DOM. Structurally, `CompanionWindow`'s `<aside>` is wired up several layers deep via absolute/flex positioning, `react-rnd` drag-resize, and animation (`WorkspaceArea.jsx` -> `ViewerArea` styled as `main` -> `Workspace` -> `WorkspaceMosaic`/`Window` -> `CompanionWindow`) -- physically moving it out via a portal would have been a much higher-risk change to that resize/position/animation behavior for a landmark-only fix.

## Change
`src/components/CompanionWindow.jsx`: added `role="region"` to the root `<Paper component="aside" ...>` element (1 line). Keeps the `<aside>` tag, all CSS classes, positioning, drag/resize, and z-index untouched -- only the exposed ARIA role changes, from the implicit "complementary" (which requires top-level placement per the WAI-ARIA APG, the actual bug) to "region" (no top-level constraint, valid nested). The element already has an accessible name via `aria-label={ariaLabel || title}`, satisfying `region`'s naming requirement.

Updated 5 test files that asserted the old `complementary` role to assert `region` instead:
- `__tests__/src/components/CompanionWindow.test.jsx`
- `__tests__/src/components/CompanionArea.test.jsx` (also had to scope one assertion to `aside[role="region"]` specifically -- MUI's `Accordion` independently renders its own nested `role="region"` divs for expandable content inside the panels, so `getAllByRole('region')` was over-matching)
- `__tests__/src/components/SearchPanel.test.jsx`
- `__tests__/integration/tests/companion-windows.test.js`
- `__tests__/integration/tests/annotations.test.js`

## How I tested this
- `npm install` (this repo uses npm/package-lock.json)
- `npx vitest run __tests__/src` -- full unit suite: 176 files, 1177 tests passed, 8 skipped, 0 failed
- `npx eslint` on all changed files -- 0 errors
- `npm run lint` (full project) -- exits 0; only pre-existing i18n key-completeness warnings on unrelated locale files
- Integration tests `companion-windows.test.js`/`annotations.test.js` have 2 network-dependent failures (fetch aborts to external endpoints) in a sandboxed environment -- verified via `git stash` that these same 2 failures occur identically on unmodified `main`, i.e. pre-existing sandbox/network limitations, not caused by this change

No DOM structure, CSS, or tag-name changes -- only an ARIA attribute was added, so there's no layout or animation regression surface.

## One nuance worth flagging for review
Because MUI's `Accordion` (used inside some panel contents) already uses `role="region"` for its own expandable sections, screen-reader landmark navigation will now show nested "region" landmarks (the CompanionWindow itself, plus any accordion sections inside it). This is spec-valid (only `complementary`/`banner`/`contentinfo` require top-level placement; `region` permits nesting) but is a minor semantic wrinkle worth being aware of, since @jcoyne's suggestion vs. @lhenze's `section`/`div` alternative would behave slightly differently here.

### DIFF
--- a/__tests__/integration/tests/annotations.test.js
+++ b/__tests__/integration/tests/annotations.test.js
@@ -44,7 +44,7 @@ describe('Annotations in Mirador', () => {
 
     // use safeFindByRole here. This specific test is consistently hard to interpret on failure,
     // since another async operation has happened after setupIntegrationTestViewer
-    const annotationPanel = await safeFindByRole('complementary', { name: /annotations/i });
+    const annotationPanel = await safeFindByRole('region', { name: /annotations/i });
     expect(annotationPanel).toBeInTheDocument();
 
     const listItems = await within(annotationPanel).findAllByRole('menuitem');
--- a/__tests__/integration/tests/companion-windows.test.js
+++ b/__tests__/integration/tests/companion-windows.test.js
@@ -11,19 +11,19 @@ describe('Basic end to end Mirador', () => {
     fireEvent.click(toggleButtons[0]);
 
     // Companion window is on the left
-    expect(await screen.findByRole('complementary', { name: /About this item/i })).toHaveClass('mirador-companion-window-left');
+    expect(await screen.findByRole('region', { name: /About this item/i })).toHaveClass('mirador-companion-window-left');
 
     const openButton = screen.getByRole('button', { name: /Open in separate panel/i });
     fireEvent.click(openButton);
 
     // Companion window is on the right
-    expect(await screen.findByRole('complementary', { name: /About this item/i })).toHaveClass('mirador-companion-window-right');
+    expect(await screen.findByRole('region', { name: /About this item/i })).toHaveClass('mirador-companion-window-right');
 
     // Close the panel
     const closeButton = screen.getByRole('button', { name: /Close panel/i });
     fireEvent.click(closeButton);
 
     // The companion window is removed
-    expect(screen.queryByRole('complementary', { name: /About this item/i })).not.toBeInTheDocument();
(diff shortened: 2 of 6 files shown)


marlo-longley:
This looks good to me. @lhenze are you able to review? I tested with my Axe tools before and after and didn't find any violations, so I want to make sure your tool (maybe it was just knowledge/manual review) is satisfied.

Thank you @Narasimha2211.

jcoyne (approved):

PR #4396, opened 2026-07-23
Ignore non-quick touch/pen gestures over annotations in onCanvasClick

### LINKED ISSUE #4184: Touchscreen gestures over annotations are always interpreted as clicks
I'm using Mirador v4.0.0-rc4 and OpenSeaDragon 5.0.1, testing primarily in Firefox on a mobile touchscreen device.

I have noticed that various touchscreen gestures, such as pinch zooms, swipes, or holding your finger to pan the canvas, if performed over an annotation, are always being interpreted as a click on that annotation as soon as you lift your finger, which causes the annotation to become selected.

## Steps to Reproduce:
1. Load any manifest with an annotation that has a visible rectangular region on the canvas.
2. On a touchscreen device, perform a pinch zoom or pan over the annotation.
3. Release your fingers while still over the annotation.

## Expected Behaviour:
OpenSeadragon emits ``canvas-click`` but with ``event.quick`` as ``false`` and ``isTouchEvent`` as ``true``, so Mirador should treat this as a non-click and leave the annotation selection unchanged.

## Actual Behaviour:
The ``onCanvasClick`` function in ``AnnotationsOverlay.jsx`` toggles annotation selection regardless of ``event.quick`` or touch gesture type. This makes normal pinch/pan gestures behave like taps.

## Possible Fix?
I have not tested this, but I'm wondering if the fix could be as simple as an early exit in ``onCanvasClick`` by checking certain attributes of the ``event`` that it receives. For example, maybe something like:
```javascript
  const onCanvasClick = useCallback((event) => {
    const pointerType = event.originalEvent?.pointerType;
    if ((event.isTouchEvent || pointerType === 'touch' || pointerType === 'pen') && event.quick === false) {
      return;
    }
    ...
  }
```

These attributes were derived from OpenSeaDragon ``canvas-click`` event, documented here: https://openseadragon.github.io/docs/OpenSeadragon.Viewer.html#.event:canvas-click

Narasimha2211:
## Summary
Closes #4184.

OpenSeadragon fires `canvas-click` at the end of touch/pen gestures too, not just quick taps -- a pinch-zoom or pan over an annotation was being interpreted as a click the moment the finger lifted, toggling the annotation's selection.

## Fix
`canvas-click` carries `event.quick` (`false` for a "slow" gesture) and `event.isTouchEvent` / `event.originalEvent.pointerType`, which distinguish a real tap from a gesture ([docs](https://openseadragon.github.io/docs/OpenSeadragon.Viewer.html#.event:canvas-click)). Added the maintainer-approved early exit from the issue thread: skip toggling selection when `(isTouchEvent || pointerType is touch/pen) && quick === false`.

## Tests
Added 3 tests to `AnnotationsOverlay.test.jsx`:
- a non-quick touch gesture over an annotation doesn't select it
- a non-quick pen gesture over an annotation doesn't select it
- a genuine quick touch tap still selects, same as before

Confirmed these are real regression tests, not just passing coincidentally: reverting the fix makes the two "ignores" tests fail (`selectAnnotation` gets called when it shouldn't).

## How I tested this
- `npx vitest run __tests__/src` -- 176 files, 1181 tests passed, 8 skipped (pre-existing, network-dependent, unrelated to this change)
- `npx eslint` on both changed files -- clean

Developer checklist from the issue:
- [x] Add the suggested early-exit condition
- [x] Write a test for this
- [x] Test manually -- via the added Testing Library tests exercising the real `onCanvasClick` handler through `viewer.raiseEvent('canvas-click', ...)`, same pattern as the existing tests in this file

### DIFF
--- a/__tests__/src/components/AnnotationsOverlay.test.jsx
+++ b/__tests__/src/components/AnnotationsOverlay.test.jsx
@@ -240,6 +240,96 @@ describe('AnnotationsOverlay', () => {
 
       expect(selectAnnotation).toHaveBeenCalledWith('base', 'http://example.org/identifier/annotation/anno-line');
     });
+
+    it('ignores a non-quick touch gesture (pinch/pan) over an annotation', () => {
+      const selectAnnotation = vi.fn();
+
+      const { viewer } = createWrapper({
+        annotations: [
+          new AnnotationList({
+            '@id': 'foo',
+            resources: [
+              {
+                '@id': 'http://example.org/identifier/annotation/anno-line',
+                '@type': 'oa:Annotation',
+                motivation: 'sc:painting',
+                on: 'http://iiif.io/api/presentation/2.0/example/fixtures/canvas/24/c1.json#xywh=100,100,250,20',
+              },
+            ],
+          }),
+        ],
+        selectAnnotation,
+      });
+
+      viewer.raiseEvent('canvas-click', {
+        eventSource: { viewport: viewer.viewport },
+        isTouchEvent: true,
+        position: new OpenSeadragon.Point(101, 101),
+        quick: false,
+      });
+
+      expect(selectAnnotation).not.toHaveBeenCalled();
+    });
+
+    it('ignores a non-quick pen gesture over an annotation', () => {
+      const selectAnnotation = vi.fn();
+
+      const { viewer } = createWrapper({
+        annotations: [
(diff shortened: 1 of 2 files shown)


marlo-longley:
Closing in favor of https://github.com/ProjectMirador/mirador/pull/4469

PR #4397, opened 2026-07-23
Reorder canvas nav controls: prev/next before zoom controls

### LINKED ISSUE #3190: UI RFE: order of canvas nav controls
I would like to suggest a different order of the zoom and pagination controls.

Instead of 
M3-zoomcontrols

[ Zoom in | Zoom out | Reset zoom][ Previous item | Next item ]

I would like to propose
[ Previous item | Next item ][ Zoom out | Reset zoom | Zoom in ]

Rational:
a) right is bigger is more zoom power. Hence the mapping feels a little bit more intuitive (for LTR locales)
b) Zoom controls are optional. Hence they should extend the standard nav controls to the right. (Though, I am not so sure to swap the groups.)

Narasimha2211:
## Summary
Closes #3190.

Swaps the two control groups in `WindowCanvasNavigationControls`' `Stack` so navigation (Previous/Next item) renders before the zoom controls, matching the order the maintainers settled on in the issue thread ([comment](https://github.com/ProjectMirador/mirador/issues/3190#issuecomment-653164913), re-affirmed [here](https://github.com/ProjectMirador/mirador/issues/3190#issuecomment-4937660711)):

```
[ Previous item | Next item ][ Zoom in | Zoom out | Reset zoom ]
```

(previously zoom controls came first). The internal order within each group is unchanged -- `ZoomControls` already renders zoom in, zoom out, reset zoom in that order, matching what was requested.

## Change
One-line swap in `src/components/WindowCanvasNavigationControls.jsx` -- `ViewerNavigation` now renders before the conditional `ZoomControls`.

## Tests
Added a test asserting the full button order via `aria-label`. Confirmed it's a genuine regression test: reverting the swap makes it fail (previous/next end up after the zoom buttons instead of before).

## How I tested this
- `npx vitest run __tests__/src` -- 176 files, 1179 tests passed, 8 skipped (pre-existing, unrelated to this change)
- `npx eslint` on both changed files -- clean

### DIFF
--- a/__tests__/src/components/WindowCanvasNavigationControls.test.jsx
+++ b/__tests__/src/components/WindowCanvasNavigationControls.test.jsx
@@ -40,4 +40,10 @@ describe('WindowCanvasNavigationControls', () => {
     expect(screen.getByRole('button', { name: 'Zoom out' })).toBeInTheDocument();
     expect(screen.getByRole('button', { name: 'Reset zoom' })).toBeInTheDocument();
   });
+
+  it('orders the nav controls before the zoom controls (issue #3190)', () => {
+    render(<Subject showZoomControls />);
+    const labels = screen.getAllByRole('button').map((button) => button.getAttribute('aria-label'));
+    expect(labels).toEqual(['Previous item', 'Next item', 'Zoom in', 'Zoom out', 'Reset zoom']);
+  });
 });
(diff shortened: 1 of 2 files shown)


gdicristofaro:
This looks good to me too.  I ran through the tests and linting.  Other than preexisting skipped tests and linting errors, tests passed including the added tests.  A manual test of the change revealed desired results.

**On Main:**
Screenshot From 2026-07-26 22-48-21 (old)


**This PR:**
Screenshot From 2026-07-26 22-47-37 (new)

marlo-longley (approved):

PR #4401, opened 2026-07-26
Give CompanionWindow's <aside> an explicit region role

### LINKED ISSUE #4225: Aside is contained within main landmark

Narasimha2211:
Closes #4225

## Problem

The sliding panels ("about this item", "rights", "index", search, annotations, etc.) render as `<aside>` (`src/components/CompanionWindow.jsx`), which has an implicit ARIA `complementary` landmark role. Per the WAI-ARIA APG, complementary landmarks are only valid as a **top-level** landmark (a sibling of `main`), not nested inside it — and `CompanionWindow` is always rendered inside `WorkspaceArea`'s `<main>` (`ViewerArea`), so every companion window panel was a landmark-nesting violation.

## Fix

Per the issue's own suggested options: kept the `<aside>` element, added `role="region"` to override its implicit landmark role. `region` is an ARIA-in-HTML-permitted role override for `<aside>`, is valid when nested inside `main` (unlike `complementary`), and `CompanionWindow` already provides an accessible name via `aria-label={ariaLabel || title}`, which is exactly what a region landmark needs to be meaningful to assistive tech.

## Test fallout

Updated 5 test files that queried for the old `'complementary'` role (11 assertions total, across `CompanionWindow.test.jsx`, `CompanionArea.test.jsx`, `SearchPanel.test.jsx`, `companion-windows.test.js`, `annotations.test.js`) to query `'region'` instead.

One of those (`CompanionArea.test.jsx`) needed more than a search-and-replace: MUI's `Accordion` component also renders its own unrelated `role="region"` divs internally (for its expandable content area), so an unscoped `getAllByRole('region')` now overcounts (found 4 instead of 2 when I ran it). Scoped that one assertion to `<aside>` elements specifically, since `CompanionWindow`'s own landmark is always an `<aside>` and MUI's Accordion regions aren't.

## Testing

Ran the full suite before and after (`git stash`): identical results except for this change — the same 7 pre-existing failures exist on unmodified `main` (unrelated integration tests, nothing to do with landmark roles or `CompanionWindow`). The files this PR actually touches are fully green: 34 tests, 0 failures. `eslint` clean.

### DIFF
--- a/__tests__/integration/tests/annotations.test.js
+++ b/__tests__/integration/tests/annotations.test.js
@@ -44,7 +44,7 @@ describe('Annotations in Mirador', () => {
 
     // use safeFindByRole here. This specific test is consistently hard to interpret on failure,
     // since another async operation has happened after setupIntegrationTestViewer
-    const annotationPanel = await safeFindByRole('complementary', { name: /annotations/i });
+    const annotationPanel = await safeFindByRole('region', { name: /annotations/i });
     expect(annotationPanel).toBeInTheDocument();
 
     const listItems = await within(annotationPanel).findAllByRole('menuitem');
--- a/__tests__/integration/tests/companion-windows.test.js
+++ b/__tests__/integration/tests/companion-windows.test.js
@@ -11,19 +11,19 @@ describe('Basic end to end Mirador', () => {
     fireEvent.click(toggleButtons[0]);
 
     // Companion window is on the left
-    expect(await screen.findByRole('complementary', { name: /About this item/i })).toHaveClass('mirador-companion-window-left');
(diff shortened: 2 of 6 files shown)


marlo-longley:
@Narasimha2211 is there a reason you opened this PR in addition to #4392 ?

marlo-longley:
Superseded by https://github.com/ProjectMirador/mirador/pull/4392

PR #4404, opened 2026-07-29
Remove the lodash dependency

jcoyne:
This drops the build size to 685.5kb from 716.8kb


__tests__/src/extend/pluginMapping.test.js
```diff
@@ -48,6 +48,17 @@ describe('createTargetToPluginMapping', () => {
       },
     });
   });
+
+  it('preserves targeted plugins when a registry-only plugin is also present', () => {
```
marlo-longley, line 52: @jcoyne I am unclear why this file is in this PR

src/extend/pluginValidation.js
```diff
@@ -1,58 +1,56 @@
...
-import values from 'lodash/values';
-
 /** */
 export const validatePlugin = (plugin) =>
   [checkPlugin, checkName, checkTarget, checkMode, checkMapStateToProps, checkMapDispatchToProps, checkReducers].every((check) =>
     check(plugin),
   );
 
+/** */
+const isObject = (value) => value !== null && (typeof value === 'object' || typeof value === 'function');
```
marlo-longley, line 8: Should this be in the utils file? I see other files with the logic inline, where the `isObject` import from lodash was removed.

jcoyne: I only see it in SearchPananelControls.jsx and that uses a different implementation.

PR #4407, opened 2026-07-29
Avoid unnecessary network request that fails due to CORS error

### LINKED ISSUE #4406: plugin-companion-window integration test depends on an external manifest and can fail due to CORS
To reproduce run:
```
 npx vitest run __tests__/integration/tests/plugin-companion-window.test.js
```

You will see this warning:
```
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at "https://purl.stanford.edu/hg676jb4964/iiif/manifest".
```

We can instead create an empty window with `windows: [{}]`, since the test does not use manifest data

jcoyne:
Fixes #4406


marlo-longley:
@jcoyne if you delete this purl, then the demo page loads nothing:
http://localhost:4444/plugin-companion-window

Issue #3294, opened 2020-10-19
Unusably small thumbnails on right hand side for wide scroll


triplingual:
With a horizontal scroll, setting thumbnails to be on the right side of the object's window results in a teeny set of thumbs with a blue border. (Pretty sure the border is not some browser setting of mine, but it's not impossible.)

Example:
[IIIF Recipe](https://preview.iiif.io/cookbook/0011-book-3-behavior/recipe/0011-book-3-behavior/) in progress, and the [recipe's manifest in Mirador](https://projectmirador.org/embed/?iiif-content=https://preview.iiif.io/cookbook/0011-book-3-behavior/recipe/0011-book-3-behavior/manifest-continuous.json)

Screenshot
![Screen Shot 2020-10-19 at 12 28 06](https://user-images.githubusercontent.com/1789625/96479208-9ee76280-1206-11eb-9c66-d4f91a0e73ed.png)

markmatney:
Related: #2628

marlo-longley:
Same root cause as #2628, but the opposite: calculateScaledSize() in ThumbnailNavigation.jsx fixes width and derives height from aspect ratio with no floor for the far-right layout

marlo-longley:
Closed by https://github.com/ProjectMirador/mirador/pull/4428

PR #4428, opened 2026-08-10
guard against small thumbnails in right sidebar

dnoneill:
Part of #3294. Design needed for when thumbnail is the super small

Before:
Screenshot 2026-08-10 at 2 10 51 PM

After:
Screenshot 2026-08-10 at 2 11 02 PM

This does cause an issue where the thumbnail will scroll because in order to not have a horribly small height we do have to have a rather large width

### DIFF
--- a/src/components/ThumbnailNavigation.jsx
+++ b/src/components/ThumbnailNavigation.jsx
@@ -119,7 +119,8 @@ export function ThumbnailNavigation({
     const calc = Math.ceil((availableWidth * canvases.length * bounds[3]) / bounds[2]);
 
     if (!Number.isInteger(calc)) return thumbnailNavigation.height + spacing;
-    return calc + spacing;
+    // Guard against incredibly small thumbnails
+    return Math.max(calc, Math.round(thumbnailNavigation.height / 3)) + spacing;
   };
 
   /** */


marlo-longley (approved):

marlo-longley:
Look good re: code but let's get a design review.

src/components/ThumbnailNavigation.jsx
```diff
@@ -101,7 +101,8 @@ export function ThumbnailNavigation({
       case 'far-right': {
         const calc = Math.floor((calculatingWidth(canvases.length) * bounds[3]) / bounds[2]);
         if (!Number.isInteger(calc)) return thumbnailNavigation.width + spacing;
-        return calc + spacing;
+        // Guard against really short thumbnails
+        return Math.max(calc, thumbnailNavigation.height / 3) + spacing;
```
marlo-longley:
```suggestion
        return Math.max(calc, Math.floor(thumbnailNavigation.height / 3)) + spacing;
```
There is a lot of `isInteger` checking around in this file, specifically lower down where we check `Number.isInteger(style.height)`. I am worried that this division / 3 could produce a decimal and then break stuff. Math.floor just rounds it down to an Integer.

marlo-longley (changes requested):

src/components/ThumbnailNavigation.jsx
dnoneill: I went with round because floor converts 6.95 to 6

marlo-longley (approved):

PR #4430, opened 2026-08-10
add all svg shapes to display on annotations

### LINKED ISSUE #3875: Displaying Annotations with SVG selectors
Currently, can only get rectangles and SVG path displayed correctly. However, other SVG shapes are not showing outlines in Mirador.

Example selector values:

**Circles**

```json
"selector": {
      "type": "SvgSelector",
      "value": "<svg><circle cx=\"711.9518432617188\" cy=\"1822.427490234375\" r=\"248.37626687954497\"></circle></svg>"
}
```

**Ellipses**

```json
"selector": {
    "type": "SvgSelector",
    "value": "<svg><ellipse cx=\"3476.093994140625\" cy=\"1242.9108276367188\" rx=\"400.323974609375\" ry=\"175.38006591796875\"></ellipse></svg>"
}
```

**Polygons**

```json
"selector": {
    "type": "SvgSelector",
    "value": "<svg><polygon points=\"2611.90234375,2079.505615234375 3570.349853515625,2211.8876953125 3501.51123046875,2370.74658203125 3294.994873046875,2524.31005859375 2770.760986328125,2450.17578125 2596.016357421875,2254.25\"></polygon></svg>"
}
```

Are these shapes supported in Mirador 3?

dnoneill:
Closes #3875 

I used this manifest to test other shapes: https://dnoneill.github.io/annonatate/manifests/iiif.lib.harvard.edu/manifest.json .

In regards to https://github.com/ProjectMirador/mirador/issues/3875#issuecomment-2767578620. There is a bigger problem where it is being set properly but it is never not hovered on or selected. I am not sure how to address.

### DIFF
--- a/__tests__/fixtures/version-3/0001-mvm-image.json
+++ b/__tests__/fixtures/version-3/0001-mvm-image.json
@@ -1,0 +1,43 @@
+{
+  "@context": "http://iiif.io/api/presentation/3/context.json",
+  "id": "https://iiif.io/api/cookbook/recipe/0001-mvm-image/manifest.json",
+  "type": "Manifest",
+  "label": {
+    "en": [
+      "Single Image Example"
+    ]
+  },
+  "items": [
+    {
+      "id": "https://iiif.io/api/cookbook/recipe/0001-mvm-image/canvas/p1",
+      "type": "Canvas",
+      "height": 1800,
+      "width": 1200,
+      "items": [
+        {
+          "id": "https://iiif.io/api/cookbook/recipe/0001-mvm-image/page/p1/1",
+          "type": "AnnotationPage",
+          "items": [
+            {
+              "id": "https://iiif.io/api/cookbook/recipe/0001-mvm-image/annotation/p0001-image",
+              "type": "Annotation",
+              "motivation": "painting",
+              "body": {
+                "id": "https://iiif.io/api/presentation/2.1/example/fixtures/resources/page1-full.png",
+                "type": "Image",
+                "format": "image/png",
+                "height": 1800,
+                "width": 1200
+              },
+              "target": "https://iiif.io/api/cookbook/recipe/0001-mvm-image/canvas/p1"
+            }
+          ]
+        }
+      ],
+      "annotations": [{
+        "type": "AnnotationPage",
+        "id": "__tests__/fixtures/version-3/svg-annotations.json"
+      }]
+    }
+  ]
+}
--- a/__tests__/fixtures/version-3/svg-annotations.json
+++ b/__tests__/fixtures/version-3/svg-annotations.json
@@ -1,0 +1,136 @@
+{
+  "id": "https://iiif.io/api/cookbook/recipe/0001-mvm-image/canvas/p1/page/p1",
+  "type": "AnnotationPage",
+  "items": [
+    {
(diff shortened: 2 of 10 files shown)


marlo-longley:
@dnoneill thanks for working on this.
The approach suggested in the ticket still seems right to me -- using path2D, which should already support these shapes. I don't think we need this complex math.

I generated a method for CanvasAnnotationDisplay that looks like this:

```js
export function buildPath2D(element) {
  const tag = element.tagName.toLowerCase();
  const num = (attr) => parseFloat(element.getAttribute(attr)) || 0;

  if (tag === 'path') return new Path2D(element.getAttribute('d'));

  const path = new Path2D();
  switch (tag) {
    case 'circle':
      path.arc(num('cx'), num('cy'), num('r'), 0, 2 * Math.PI);
      return path;
    case 'ellipse':
      path.ellipse(num('cx'), num('cy'), num('rx'), num('ry'), 0, 0, 2 * Math.PI);
      return path;
    case 'rect':
      path.rect(num('x'), num('y'), num('width'), num('height'));
      return path;
    case 'polygon':
    case 'polyline':
      parsePoints(element.getAttribute('points')).forEach(([x, y], i) => {
        if (i === 0) path.moveTo(x, y);
        else path.lineTo(x, y);
      });
      if (tag === 'polygon') path.closePath();
      return path;
    default:
      return null;
  }
}
```
What do you think about working with something like this instead? 

I also generated a fixture (attached) from Rainer's comment in the issue that showcases all the forms, you could use that in the /svg-annotations demo page. Happy to talk more about it.
[0001-mvm-image-with-svg-annotations.json](https://github.com/user-attachments/files/30915608/0001-mvm-image-with-svg-annotations.json)

dnoneill:
Okay. It is going to require me to call this in multiple places and hopefully I didn't miss where it would break anything else but that should work. I need to fix the polygon function but I can switch to this approach.

marlo-longley:
@dnoneill this is looking really good!

1. I am wondering if you can adjust some of these shapes in the new example manifest. I think they were generated and a few are confusing
- Test (polygon) -- this one seems messed up. Delete? It's like a weird line sticking off a rectangle.
- Test (path, mislabeled "ellipse" in original bug report)
   - this one seems to just be a polygon. its coordinates also go off the canvas
   - change the label, it is misleading, should say polygon
- Test (SVG rect element, not a xywh fragment)
    - Edit label just to say rect. xywh is irrelevant/extra info
 
Screenshot 2026-08-12 at 11 03 11 AM
Screenshot 2026-08-12 at 11 02 50 AM

2. Re-reading the original issue, I am concerned about [the comment on scaling](https://github.com/ProjectMirador/mirador/issues/3875#issuecomment-2638219251). I am hoping you can think through this. Pasting in [his example manifest](https://gist.githubusercontent.com/mk-pmb/bb9783a4327963ed34a0bc6890ea7d2e/raw/16888807646d8b674be08e533432c77aa7beb693/svgselector-scaling.json), it is true that the scaling is wrong. This should probably be a new issue not in this PR. But since I haven't looked into it, maybe it is related/easy to include.

dnoneill:
@marlo-longley In response to your concerns:
For test polygon I put a line in it because I wanted to make sure I tested a <g> wrapper with multiple elements. That line is on purpose. But I did not update the text. I will make sure to update it to the correct element
If I understand the issue correctly I think I have run into this. I think we need to use a combination of xywh coordinates to draw the svg but I am not quite sure I understand

marlo-longley (dismissed):

marlo-longley (approved):

PR #4438, opened 2026-08-11
Migrate to RTK's configureStore

### LINKED ISSUE #4420: Migrate store setup to RTK configureStore

marlo-longley:
- **Migrate from legacy Redux createStore to RTK's configureStore**
- **Don't store un-serializable Error object in state**
- **Suppress warning about non-serializable imageResource**

Closes #4420 

- rename file createStore.js > store.js
- rename file createPluggableStore.js > pluggableStore.js
- keep export names the same:
```
export { default as createStore } from './store';
export { default as createPluggableStore } from './pluggableStore';
```
- suppress or fix new warnings about non-serializable objects in state (RTK's new warnings)


jcoyne:
This needs a rebase. When you do that, do you want to squash these commits? I'm not sure they stand on their own.

jcoyne:
Should `redux-thunk` be removed from package.json too?

marlo-longley:
@jcoyne Thanks, OK I squashed most of the commits but left one for removing redux-thunk as you suggested.

PR #4441, opened 2026-08-12
use OpenSeadragon built in function to calculate zoom levels

### LINKED ISSUE #4246: Annotation bounding box too thick with large osdConfig.maxZoomPixelRatio
This is the same issue as https://github.com/ProjectMirador/mirador/issues/3924, but narrowed down to the config setting at fault.

Setting osdConfig.maxZoomPixelRatio to higher values causes the annotation box to be drawn with a very thick stroke.

Image

The above is with maxZoomPixelRatio set to 100, using this IIIF cookbook sample: https://iiif.io/api/cookbook/recipe/0021-tagging/manifest.json . Of course it's quite extreme at a value of 100, but less severe versions of this effect are visible with smaller maxZoomPixelRatio settings (it's noticeable at 2). The effect is worse the further zoomed out the user is; zooming in makes the stroke get smaller.

It would seem that the issue stems from https://github.com/ProjectMirador/mirador/blob/v4.0.0/src/components/AnnotationsOverlay.jsx#L63 : setting maxZoomPixelRatio alters the OSD maxZoom return value, and the ratio of the current zoom to max zoom gets calculated and passed to the CanvasAnnotationDisplay, which itself sets the stroke width to 1/zoomRatio. So: the more zoomed-out vs. the max the viewport is, the smaller the current ratio is, and the thicker the stroke gets.

The default max zoom ratio is 1.1 so there's not usually as much range to play with here so it's not so noticeable, but bigger maxZoomPixelRatio makes the scaling get out of whack. I guess this is intended to keep the bounding box clearly visible as the user zooms far out, and it's just that big maxZoomPixelRatio settings make that stroke size that's getting maintained big also.

OSD also supports Infinity as a setting here to allow unlimited zooming in: using that breaks the stroke scaling entirely. The calculated zoomRatio is always 0, leading to a calcuated lineWidth of Infinity... which is presumably just getting ignored and some fixed value (1?) is getting used instead.

I don't know if it's as simple as this, but multiplying the calculated zoomRatio in AnnotationsOverlay by the viewport's maxZoomPixelRatio seems to fix it.

dnoneill:
Closes #4246
[viewportToImageZoom](https://openseadragon.github.io/docs/OpenSeadragon.Viewport.html#viewportToImageZoom): `Convert a viewport zoom to an image zoom. Image zoom: ratio of the original image size to displayed image size. 1 means original image size, 0.5 half size... Viewport zoom: ratio of the displayed image's width to viewport's width. 1 means identical width, 2 means image's width is twice the viewport's width... Note: not accurate with multi-image`. Since we are using the data from the manifest in CanvasAnnotationDisplay is specific to the image.

I was worried this would cause an issue because of the note that it doesn't work with multi-image and it was causing a console error. The way around this is to find the correct canvas the image is on in the OSD viewer and do the calculation that way. This ended up working.

Also for historical reference

|maxZoomPixelRatio |Old calc | New calc |
|---|---|---|
| No | 0.07294561583071928 | 0.08156810121729129 |
| = 100 | 0.0008024017741379122 | 0.08024017741379122 |

### DIFF
--- a/__tests__/src/components/AnnotationsOverlay.test.jsx
+++ b/__tests__/src/components/AnnotationsOverlay.test.jsx
@@ -114,7 +114,9 @@ describe('AnnotationsOverlay', () => {
         viewer: null,
       });
 
-      vi.spyOn(viewer.viewport, 'getMaxZoom').mockImplementation(() => 1);
+      const getItemAt = vi
+        .spyOn(viewer.world, 'getItemAt')
+        .mockImplementation((index) => (index === 0 ? { viewportToImageZoom: vi.fn(() => 0.05) } : undefined));
       vi.spyOn(viewer.viewport, 'getZoom').mockImplementation(() => 0.05);
 
       rerender(
--- a/src/components/AnnotationsOverlay.jsx
+++ b/src/components/AnnotationsOverlay.jsx
@@ -70,11 +70,14 @@ export function AnnotationsOverlay({
   const annotationsToContext = useCallback(
     (renderedAnnotations, currentPalette) => {
       const context = osdCanvasOverlay.context2d;
-      const zoomRatio = viewer.viewport.getZoom(true) / viewer.viewport.getMaxZoom();
       renderedAnnotations.forEach((annotation) => {
         annotation.resources.forEach((resource) => {
-          if (!canvasWorld.canvasIds.includes(resource.targetId)) return;
+          const osdCanvasIndex = canvasWorld.canvases.findIndex((canvas) => canvas.id === resource.targetId);
+          if (osdCanvasIndex === -1) return;
+          const viewportCanvas = viewer.world.getItemAt(osdCanvasIndex);
+          if (!viewportCanvas) return;
           const offset = canvasWorld.offsetByCanvas(resource.targetId);
+          const zoomRatio = viewportCanvas.viewportToImageZoom(viewer.viewport.getZoom(true));
           const canvasAnnotationDisplay = new CanvasAnnotationDisplay({
             hovered: hoveredAnnotationIds.includes(resource.id),
             offset,


dnoneill:
Okay there is a console error. Not 100% sure what to do since it works. I will poke some more

dnoneill:
Fix explained in comment above

__tests__/src/components/AnnotationsOverlay.test.jsx
```diff
@@ -114,7 +114,9 @@ describe('AnnotationsOverlay', () => {
         viewer: null,
       });
 
-      vi.spyOn(viewer.viewport, 'getMaxZoom').mockImplementation(() => 1);
+      vi.spyOn(viewer.world, 'getItemAt').mockImplementation(() => ({
+        viewportToImageZoom: vi.fn(() => 0.05),
+      }));
```
marlo-longley:
```suggestion
     const getItemAt = vi
        .spyOn(viewer.world, 'getItemAt')
        .mockImplementation((index) => (index === 0 ? { viewportToImageZoom: vi.fn(() => 0.05) } : undefined));
```

src/components/AnnotationsOverlay.jsx
```diff
@@ -69,11 +69,13 @@ export function AnnotationsOverlay({
...
       const context = osdCanvasOverlay.context2d;
-      const zoomRatio = viewer.viewport.getZoom(true) / viewer.viewport.getMaxZoom();
       renderedAnnotations.forEach((annotation) => {
         annotation.resources.forEach((resource) => {
-          if (!canvasWorld.canvasIds.includes(resource.targetId)) return;
+          const osdCanvas = canvasWorld.canvases.find((canvas) => canvas.id === resource.targetId);
+          if (!osdCanvas) return;
           const offset = canvasWorld.offsetByCanvas(resource.targetId);
+          const osdCanvasIndex = osdCanvas.canvas.index - 1;
+          const zoomRatio = viewer.world.getItemAt(osdCanvasIndex).viewportToImageZoom(viewer.viewport.getZoom(true));
```
marlo-longley:
```suggestion
          // viewer.world only holds items for canvasWorld.canvases (this window's currently
          // loaded canvases), not the manifest as a whole, so index into that list rather
          // than the canvas's manifest-wide index.
          const osdCanvasIndex = canvasWorld.canvases.findIndex((canvas) => canvas.id === resource.targetId);
          if (osdCanvasIndex === -1) return;
          const offset = canvasWorld.offsetByCanvas(resource.targetId);
          const osdCanvasItem = viewer.world.getItemAt(osdCanvasIndex);
          if (!osdCanvasItem) return;
          const zoomRatio = osdCanvasItem.viewportToImageZoom(viewer.viewport.getZoom(true));
```

marlo-longley (changes requested):
I ran into an error looking at /svg-annotations on this branch -- trying to look at the Bodleian circular fixture.

```
Uncaught TypeError: Cannot read properties of undefined (reading 'viewportToImageZoom')
    at AnnotationsOverlay.jsx:78:67
```
`osdCanvas.canvas.index` is the canvas's position in the whole manifest, but `viewer.world` only holds items for the canvases actually loaded into this window. So `canvas.index - 1` and "position in viewer.world" are two different numbers that only coincidentally match when a manifest's first canvas happens to be viewed alone.

This fixed it for me locally.

src/components/AnnotationsOverlay.jsx
dnoneill: Yeah I misunderstood how the index worked. ~~This still gives a broken view. I think because of the LTR vs RTL but I need to check.~~ Nevermind I think my local instance was being wonky. This was my first thought but I thought RTL/LTR might cause a problem but it looks like it is ordered properly.

marlo-longley: Ok cool. I was personally confused about the index at first!

dnoneill: Actually this is causing a race condition. I need to explore further. I am getting an error: ```can't access property "viewportToImageZoom", viewer.world.getItemAt(...) is undefined``` If I move to fast. Yikes.

dnoneill: Okay I added a guard. From what I can tell we are adding/updating the annotations everytime the viewport changes which seems bananas to me but nothing I can do about it right now.

marlo-longley:
Re-review: use a manifest with big size differences.
Look closely at multi-image -- Niqui limited the call to only work on single canvas

dnoneill:
https://dnoneill.github.io/annonatate/manifests/sg359db9318/iiif/manifest.json has annotations on book pages that are different sizes

marlo-longley (approved):

dnoneill:
@marlo-longley is this ready to merge?

Issue #4354, opened 2026-07-14
Accessibility: visible focus indicator on OSD canvas + keyboard landmark navigation within window


marlo-longley:
Broken out from https://github.com/ProjectMirador/mirador/issues/1660 

Two related keyboard/focus accessibility gaps:

1. Visible focus indicator on OSD canvas
The OpenSeadragon canvas (`<section>` wrapping OSD) receives focus when tabbed to but has no visible focus ring. OSD renders into a <canvas> element which is not natively focusable — the focus state needs to be applied to the wrapping element with a visible indicator. Visual design TBD.

2. Keyboard landmark/region navigation within a window
There is no easy way for keyboard users to jump between major regions inside a window (OSD canvas, sidebar, companion windows, top bar). Standard approach would be skip links or ARIA landmark roles (main, navigation, complementary) on window regions, allowing screen reader and keyboard users to navigate directly to a region without tabbing through all controls.

Designer action: describe the proper flow of interactions

PR #4442, opened 2026-08-13
Define the thumbnail menu as a nav landmark

### LINKED ISSUE #4185: WCAG: Thumbnail menu should be defined as <nav> element
This is from an external accessibility audit of our Mirador 3 install (WCAG 2.2 AA).

Feedback: Marking the thumbnail menu as a nav element would make navigation clearer for screen readers.
WCAG 1.3.1: https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html

yarrrly:
## Summary
Closes #4185.

The thumbnail navigation strip renders its root `Paper` as a plain `div` with `aria-label="Thumbnails"` and `tabindex="0"`, but no landmark role. Screen readers expose it as a generic container: it has a name, but it never shows up in landmark navigation. This is the finding from the WCAG 2.2 AA audit behind #4185 (W3C technique H97 per the comment above).

## Change
`src/components/ThumbnailNavigation.jsx`: added `component="nav"` to the root `Paper` (1 line), following the in-codebase precedent @marlo-longley pointed to (`WorkspaceControlPanel.jsx`: `component="nav"` plus `aria-label` on its AppBar root). The existing `aria-label={t('thumbnailNavigation')}` ("Thumbnails") becomes the landmark's accessible name, so no new translation keys are needed. Keyboard handling (`tabindex`, arrow-key canvas switching), focus behavior, react-window layout, and all CSS classes are untouched; only the rendered tag changes from `div` to `nav`.

Added one test asserting the `navigation` role with its accessible name, in the style of #4392's role assertions.

## Landmark exposure before/after
Measured with axe-core 4.12.1 injected into the local dev demo (`/demo/index.html`), on current main (`a922434`):

- Before: the two-window demo exposes 14 landmarks; the thumbnail menu is not among them (the accessibility tree shows `generic "Thumbnails"`).
- After: `navigation "Thumbnails"` appears (nav landmarks go from 3 to 4).
- axe violations: 2 before, 2 after, identical nodes. Both are pre-existing and unrelated to this change (the duplicated "Window navigation" labels across the two windows, and `scrollable-region-focusable` on the react-window grid).

## One nuance worth flagging for review
With several windows open that all have thumbnails enabled, each window contributes a `nav` labeled "Thumbnails", so axe's `landmark-unique` reports the duplicate pair. Measured A/B on the demo: baseline has 1 `landmark-unique` node (the existing per-window "Window navigation" pair); with this change and two thumbnail strips visible it is 2 nodes. This is the same per-window static-label pattern the codebase already ships for "Window navigation"; disambiguating per-window landmark labels looks like #4354 territory rather than part of this one-line fix.

## How I tested this
- `npm install`, `npm test` (build, lint including i18n-lint and container-lint, size-limit, vitest): exit 0; 191 test files passed, 2 skipped; 1204 tests passed, 12 skipped
- `npx vitest run __tests__/src/components/ThumbnailNavigation.test.jsx`: 16/16
- Manual on the dev demo: keyboard focus lands on the `nav`, ArrowRight still advances the canvas ("1 of 84" to "2 of 84"); on the RTL demo (`viewingDirection: right-to-left` manifest) the grid keeps `direction: rtl` in both `far-bottom` and `far-right` positions

### DIFF
--- a/__tests__/src/components/ThumbnailNavigation.test.jsx
+++ b/__tests__/src/components/ThumbnailNavigation.test.jsx
@@ -44,6 +44,12 @@ describe('ThumbnailNavigation', () => {
 
     focusSpy.mockRestore();
   });
+
+  it('renders as a nav landmark with an accessible name', () => {
+    render(<Subject />);
+
+    expect(screen.getByRole('navigation')).toHaveAccessibleName('Thumbnails');
+  });
   it('renders containers based off of number of canvases', () => {
     render(<Subject />);
 
--- a/src/components/ThumbnailNavigation.jsx
+++ b/src/components/ThumbnailNavigation.jsx
@@ -176,6 +176,7 @@ export function ThumbnailNavigation({
         scrollbarGutter: 'stable',
       }}
       aria-label={t('thumbnailNavigation')}
+      component="nav"
       square
       elevation={0}
       style={style()}


yarrrly:
Bumping this once, and it is about CI rather than review time.

The 2026-08-23 rebase onto 7128a023 (head `d524d85c`) cleared the conflict with #4487,
but it also reset the Actions suite to `action_required`, so the build needs a maintainer
click before it can be green again. The Netlify checks pass. The diff is still +7/-0 in
two files, and on the rebased tree the full suite was 1231 passed / 10 skipped locally,
with eslint and prettier clean.

If this is waiting on an accessibility reviewer, that is fine and I am not asking anyone
to jump the queue. I only wanted to flag that it is currently blocked on a button rather
than on anything I can fix. It claims #4185.

marlo-longley:
@yarrrly we are discussing whether `nav` is the appropriate landmark container. See discussion in ticket.

Issue #4393, opened 2026-07-23
Bug when switching from single to book view


morpheus-87:
I just noticed that there seems to be a bug when switching from single to book view.

How to reproduce:
1. open the Mirador demo: https://projectmirador.org/demo/
2. load the manifest "Wunder der Vererbung / von Fritz Bolle." from the catalogue
3. switch to single view
4. scroll to an even page
5. switch to book view

Results in a missing second page 🤔 after scrolling back and forth, the second page is visible.

PR #4449, opened 2026-08-13
make sure fetch visible canvases when they are updated

### LINKED ISSUE #3276: Chosing scroll view does not update visible canvases
Hello! 
I've noticed some strange behavior while trying to add views in Mirador - chosing scroll view doesn't trigger downloading canvases - only canvases visible are the ones previously "visited" (behavior presented in gif below). I've tried with fields `viewingHint: continuous, viewingDirection: left-to-right` in sequence and `{key: "scroll", behaviors: ["continuous"] }` in config, but in this case initially displayed view became scroll, despite the `defaultView: single`  (but scroll view was loaded correctly). 
  Is it expected behavior, bug or maybe I'm missing something? 

Used manifest  [on pastebin](https://pastebin.pl/view/1956ada2) (maintained by me). It's currently not on production, so urls are from localhost.

Part of the  Mirador's config: 
```    
window: {
      defaultView: "single",
      views: [
        {
          key: "single"
        },
        {
          key: "book"
        },
        {
          key: "scroll"
        },
        {
          key: "gallery"
        }
      ],
    },
    windows: [
     {
         canvasIndex: 0,
         loadedManifest: manifestUrl
     }
],
...,
```
mirador-2020-10-05_14 31 38

dnoneill:
closes #3276 

Honestly I thought the solution to this was fixing the reducers so that SET_WINDOW_VIEW_TYPE included a sequenceId which was causing this method to not run. https://github.com/ProjectMirador/mirador/blob/main/src/state/sagas/windows.js#L60. I was having a heck of a time with updating the payloads because they make no sense to me and weren't working. I was looking at updating it in line 209 but Claude said to add the code below. I have confirmed it looks right. I really don't understand the pattern at all in Mirador...

Also I added a note but this wasn't just broken in continuous. It was broken in book as well.

### DIFF
--- a/__tests__/src/sagas/windows.test.js
+++ b/__tests__/src/sagas/windows.test.js
@@ -403,12 +403,14 @@ describe('window-level sagas', () => {
         .provide([
           [select(getWindow, { windowId }), { canvasId: 'y' }],
           [select(getCanvasGrouping, { canvasId: 'y', windowId }), [{ id: 'y' }, { id: 'z' }]],
+          [call(fetchInfoResponses, { visibleCanvases: ['y', 'z'], windowId })],
         ])
         .put({
           id: windowId,
           payload: { visibleCanvases: ['y', 'z'] },
           type: ActionTypes.UPDATE_WINDOW,
         })
+        .call(fetchInfoResponses, { visibleCanvases: ['y', 'z'], windowId })
         .run();
     });
   });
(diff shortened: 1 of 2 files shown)


marlo-longley:
@dnoneill let's look at this as a team tomorrow. Thanks for digging in.

marlo-longley:
This PR may also close https://github.com/ProjectMirador/mirador/issues/4393

PR #4460, opened 2026-08-18
Use react-hook/resize-observer instead of react-resize-observer

cbeer:
(maybe react-resize-observer is not compatible with vite? something about esmodules 🤷 )


__tests__/src/components/WorkspaceElastic.test.jsx
```diff
@@ -91,25 +91,5 @@ describe('WorkspaceElastic', () => {
         y: -1 * (300 - 20),
       });
     });
-
-    it('when workspace itself is resized', () => {
```
cbeer, line 95: There isn't a great way to test the resize observer without mocking the whole world.

marlo-longley (approved):

Issue #4130, opened 2025-03-06
Mirador seems to draw the fill color in front of the border.


mk-pmb:
Manifest URLs: [JSON](https://gist.githubusercontent.com/mk-pmb/d193cdd398749bde1c02e01e561b9874/raw/0f772787fd9a405ed58d5c4e7684d751b880f116/iiif-line-style.json) | [View in Mirador](https://projectmirador.org/embed/?iiif-content=https://gist.githubusercontent.com/mk-pmb/d193cdd398749bde1c02e01e561b9874/raw/0f772787fd9a405ed58d5c4e7684d751b880f116/iiif-line-style.json)

When I click the annotation in the list to have it highlighted permanently, the green fill area is drawn above the yellow border. I think the border should sit on top of the fill area, hiding its inner part.

PR #4463, opened 2026-08-18
fix stroke/fill opacity to actually work with svg values

dnoneill:
Part of what I found with #4130 is that the alpha color was being set in an rgba value and not actually set via the code. This should fix the problem that even though the red on the manifest had a fill-opacity set ` d='M100,100 L400,700 L700,100 Z' stroke-width='10' stroke='red' fill='firebrick' fill-opacity='30%` it didn't actually work and it was a 100% fill opacity. 
Screenshot 2026-08-18 at 2 57 22 PM

Also now that I understand what globalAlpha is doing the code to set it for stroke needs to be moved down.


marlo-longley:
Nice find and nice fix Niqui!

This suggestion isn't critical or blocking, but I noticed the test file getting a bit long. I think the context is set up 6 times. Consider pulling out:
```
function createMockContext(onFill) {
  return {
    fill: vi.fn(onFill),
    restore: vi.fn(),
    save: vi.fn(),
    setLineDash: vi.fn(),
    stroke: vi.fn(),
    translate: vi.fn(),
  };
}                                                                                                                                              
```

Then each test can use ` const context = createMockContext(() => { alphaAtFill = context.globalAlpha; })` or just `createMockContext()` if globalAlpha is empty.

marlo-longley:
Another question -- can this code handle plain opacity?
https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Attribute/opacity 
So like `<circle cx="150" cy="50" r="40" fill="black" opacity="0.3" />`

If not we can address elsewhere not in the PR

dnoneill:
@marlo-longley No. It has never been supported in the code, so yes I think that is a separate ticket to add the functionality.

marlo-longley (approved):
Nice find Niqui.

PR #4464, opened 2026-08-18
Remove duplicate announcement of window titles by screen readers

### LINKED ISSUE #3904: Redundant title read by screen reader

marlo-longley:
This PR uses a generic `Item Window <windowNum>` for the aria-label of each window
The visible title in the UI still displays the manifest title
Closes #3904

@mferrarini -- are you able to review this? I followed your suggestion about using `Item window`. Thank you.

### DIFF
--- a/__tests__/integration/tests/minimalist.test.js
+++ b/__tests__/integration/tests/minimalist.test.js
@@ -9,7 +9,14 @@ describe('Minimalist configuration to Mirador', () => {
   it('Loads a manifest and displays it without some of the default controls', async () => {
     expect(
       await screen.findByRole('region', {
-        name: /Window: Cambridge, Corpus Christi College, MS 640: Antiphoner Leaf/i,
+        name: /Item window 1/i,
+      }),
+    ).toBeInTheDocument();
+
+    expect(
+      await screen.findByRole('heading', {
+        level: 2,
+        name: /Cambridge, Corpus Christi College, MS 640: Antiphoner Leaf/i,
       }),
     ).toBeInTheDocument();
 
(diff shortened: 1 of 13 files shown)


cbeer:
Are window numbers at all meaningful for the user? I think a little meaningful repetition is nicer than something that doesn't help the user identify what they're doing. At least the screen-reader users I know are happy to skip over text when they get the context.

marlo-longley:
@cbeer thanks, are you saying just have "Item window" with no number, or not do this PR at all?

cbeer:
I guess I'm saying, I don't see the problem with `Item window: {{TITLE}}`. It's clear what it is, and I know what I'm going to find when I descend in there.

PR #4465, opened 2026-08-18
Replace react-image with our own ImageWithFallback

### LINKED ISSUE #4266: Implement placeholder without react-image dependency

### LINKED ISSUE #4378: Unify the two separate "image failed to load" fallback mechanisms
Written with Claude's help sorry!

### Problem

There are two independent, non-interacting mechanisms in the codebase for handling image load failures, doing conceptually the same thing:

1. FailedImageContext/FailedImageProvider (1src/contexts/FailedImageContext.js`, `FailedImageProvider.jsx`) — a centralized context tracking a Set of failed image URLs, with a `notifyFailure()` callback. Used by:
- `IIIFThumbnail.jsx:38,135-137`
- `OpenSeadragonTileSource.jsx:16,72`

Failures reported through this path are visible app-wide via `ImageFailureMessage.jsx`, which reads `failedImages` and shows an in-viewer status message.

2. A local, per-component unloader/hasError pattern — currently react-image's unloader prop, being replaced in #4267 by ImageWithFallback's own local useState. Used by:
- `AttributionPanel.jsx:57`
- `ManifestListItem.jsx:134,167`

Failures here are invisible to the rest of the app — no shared state, no ImageFailureMessage visibility, nothing a plugin could observe

### Why this matters

- Inconsistent behavior for the same conceptual event. An image failing to load in a thumbnail surfaces a shared, observable failure; the exact same kind of failure on a manifest logo or attribution image does not. There's no principled reason these should be handled differently — it's an accident of which component was written first / against which library.
- Duplicated logic. Both paths solve "show a fallback node when  fires onError" — one well-abstracted and shared, one reimplemented per call site.
- Compounds a known gap. Per the plugin API audit done earlier (ask @marlo-longley), `FailedImageContext`'s data isn't even exposed to plugins yet — but at least it's one place that could be fixed. Two parallel mechanisms means fixing that gap only helps half of the app's image failures.

### Proposed direction

Migrate `AttributionPanel`/`ManifestListItem` (and `ImageWithFallback` generally, once #4267 lands) onto `FailedImageContext` instead of local component state — `ImageWithFallback` should call `notifyFailure(src)` from `FailedImageContext` on error, the same way `IIIFThumbnail`/`OpenSeadragonTileSource` already do, rather than tracking `hasError` itself.

marlo-longley:
- **Replace react-image with our own ImageWithFallback**
- **Report to failedImageContext and use Img.decode**

Replaces / rebases / adds to @jcoyne 's https://github.com/ProjectMirador/mirador/pull/4267

Closes https://github.com/ProjectMirador/mirador/issues/4378 
Closes https://github.com/ProjectMirador/mirador/issues/4266

Looks like
Screenshot 2026-08-18 at 3 35 02 PM

### DIFF
--- a/src/components/ImageWithFallback.jsx
+++ b/src/components/ImageWithFallback.jsx
@@ -1,0 +1,22 @@
+import { useState } from 'react';
+import PropTypes from 'prop-types';
+
+/**
+ * A plain img element that renders a fallback node when the image fails to load.
+ */
+export function ImageWithFallback({ alt, fallback, src, ...props }) {
+  const [hasError, setHasError] = useState(false);
+
+  if (hasError) return fallback;
+
+  return <img alt={alt} src={src} onError={() => setHasError(true)} {...props} />;
+}
+
+ImageWithFallback.propTypes = {
+  /** Alt text for the underlying img element. */
+  alt: PropTypes.string.isRequired,
+  /** Node rendered in place of the image when it fails to load. */
+  fallback: PropTypes.node.isRequired,
+  /** URL of the image to display. */
+  src: PropTypes.string.isRequired,
+};
--- a/src/components/ImageWithFallback.jsx
+++ b/src/components/ImageWithFallback.jsx
@@ -1,15 +1,43 @@
-import { useState } from 'react';
+import { useState, useEffect, useContext } from 'react';
 import PropTypes from 'prop-types';
+import FailedImageContext from '../contexts/FailedImageContext';
 
 /**
  * A plain img element that renders a fallback node when the image fails to load.
+ *
+ * Reports failures through FailedImageContext like IIIFThumbnail/OpenSeadragonTileSource
  */
 export function ImageWithFallback({ alt, fallback, src, ...props }) {
   const [hasError, setHasError] = useState(false);
+  const { notifyFailure } = useContext(FailedImageContext);
+
+  // A later src may succeed even if an earlier one failed.
+  useEffect(() => {
+    setHasError(false);
+  }, [src]);
+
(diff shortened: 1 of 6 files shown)


src/components/ImageWithFallback.jsx
```diff
@@ -0,0 +1,50 @@
...
+
+  // A later src may succeed even if an earlier one failed.
+  useEffect(() => {
+    setHasError(false);
+  }, [src]);
+
+  useEffect(() => {
+    const img = new Image();
+    img.src = src;
+    if (typeof img.decode === 'function') {
```
cbeer, line 22: I think there's some magic here I'm not understanding.   To me, it looks like we're just decoding the image for fun and never do anything with it?

marlo-longley:
@cbeer I removed the Img.decode stuff. Here is my understanding of the before/after state as far as decoding:

- Before, react-image never never mounted an  until an offscreen probe image had fully loaded and decoded
- Before, we never showed a loading state with react-image. We either showed the loaded image or failure.
- After, we mount the image immediately, browser handles decode/paint
- This explains why I could un-skip the test in this PR, because before the internal decoding step of react-image was messing with our testing setup
- I guess there is some minor risk of ... ? An un-decode-able image? But these are only for thumbnails. It does not touch our main viewer images. So the risk area is limited and not a core function.

PR #4469, opened 2026-08-18
Ignore non-quick touch/pen gestures over annotations in onCanvasClick

### LINKED ISSUE #4184: Touchscreen gestures over annotations are always interpreted as clicks

marlo-longley:
This rebases and takes over https://github.com/ProjectMirador/mirador/pull/4396/changes which was abandoned 
Closes #4184

### DIFF
--- a/src/components/AnnotationsOverlay.jsx
+++ b/src/components/AnnotationsOverlay.jsx
@@ -29,6 +29,19 @@ function isAnnotationAtPoint(canvasWorld, osdCanvasOverlay, resource, canvas, po
 }
 
 /**
+ * See: https://openseadragon.github.io/docs/OpenSeadragon.Viewer.html#.event:canvas-click
+ * Gestures returned from this method are likely viewer interactions such as pan and zoom,
+ * not taps and clicks targeted to an annotation.
+ * In deciding whether to change the display of the annotation overlay,
+ * we want to ignore these irrelevant gestures.
+ * @private
+ */
+function isIgnoredGesture(event) {
+  const pointerType = event.originalEvent?.pointerType;
+  return (pointerType === 'touch' || pointerType === 'pen') && event.quick === false;
+}
+
+/**
  * Represents a OpenSeadragonViewer in the mirador workspace. Responsible for mounting
  * and rendering OSD.
  */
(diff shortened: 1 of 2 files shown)


src/components/AnnotationsOverlay.jsx
```diff
@@ -28,6 +28,15 @@ function isAnnotationAtPoint(canvasWorld, osdCanvasOverlay, resource, canvas, po
   return false;
 }
 
+/**
+ * See: https://openseadragon.github.io/docs/OpenSeadragon.Viewer.html#.event:canvas-click
+ * @private
+ */
+function isNonTapGesture(event) {
```
cbeer: I think the `non` in here is bothering me. Is there a positive framing that'd clarify what we're trying to do (`isCanvasInteractionGesture`? `isIgnoredGesture`?)

marlo-longley: Thanks. I went with `isIgnoredGesture` and added a comment.

PR #4478, opened 2026-08-19
Guard against missing companion windows ids.

cbeer:
I don't think this is a real case (outside of tests) but seems like a harmless enough fallback to introduce 🤷


marlo-longley:
@cbeer I think this is covered here, but you may want the 1-file change alone first
https://github.com/ProjectMirador/mirador/pull/4368

PR #4479, opened 2026-08-19
Migrate layers reducer and action to RTK layersSlice

marlo-longley:
### Notes:

- RTK encourages past tense reducer names such as `layersUpdated` -- but I kept `updateLayers` because it's already the name of action creator that other code and potentially third-party plugins already call: `updateLayers(windowId, canvasId, payload)`
- To avoid our old `payload` being nested with RTK's `payload`, I renamed that key `layerData`
- I manually put in an error message to this slice, but if we continue the migration we can extract that for use across slices

### Breaking changes:

1. A plugin that skips the creator and hand-constructs the action itself — `dispatch({ type: ActionTypes.UPDATE_LAYERS, windowId, canvasId, payload })` rather than `updateLayers(windowId, canvasId, payload)` will break but will throw a useable error

2. A saga that watches and destructures the action — matches Mirador's own use in sagas, and could probably be present in plugins:
https://github.com/ProjectMirador/mirador/blob/2e779c5851fb7f537902c708bce1a3038e3d819f/src/state/sagas/annotations.js#L9
This pattern doesn't apply to layers, so doesn't impact this PR. But it will be problematic if/as the migration continues

3. A plugin that hardcodes the literal type string instead of importing the constant — `takeEvery('mirador/UPDATE_LAYERS', ...)` The dispatched type changed to `layers/updateLayers`, so this watcher would just stop working, with no error. But anyone who imports `ActionTypes.UPDATE_LAYERS` rather than hardcoding is fine.


marlo-longley:
@cbeer I folded in your work from #4480 into the latest commit on this branch: https://github.com/ProjectMirador/mirador/pull/4479/changes/323e7892c662f95c015267e306959db0983c2ba3

src/state/reducers/layersSlice.js
```diff
@@ -0,0 +1,46 @@
...
+import ActionTypes from '../actions/action-types';
+
+const layersSlice = createSlice({
+  name: 'layers',
+  initialState: {},
+  reducers: {
+    updateLayers: {
+      reducer(state, action) {
+        const { windowId, canvasId, ...payload } = action.payload || {};
+        if (windowId === undefined || canvasId === undefined) {
```
cbeer, line 12: This and the prepare call are a little awkward.. in #4491, I was trying to figure out how we handle this when we have lots of actions to migrate. I wonder if (as in 4491), it's cleaner if we keep the legacy path where it always was and let the slice just handle the new approach?

PR #4480, opened 2026-08-19
Keep layers payload at the top-level.

cbeer:


marlo-longley:
This is cleaner! I guess there is an unlikely case where there is the same key within payload and it overwrites what's already at the top-level through the `...` merge.

__tests__/src/reducers/layersSlice.test.js
```diff
@@ -53,10 +53,8 @@ describe('layers reducer', () => {
         layersReducer(
           {},
           {
-            canvasId,
```
marlo-longley, line 58: I don't understand why you removed these lines from the test

cbeer: In order to trigger the custom error message? I think your test was relying on `layerData` not being set

marlo-longley: I think we want to test what happens when you pass the old shape:
```
{
  canvasId,
  windowId,
  payload: { some: 'data'},
  type: ActionTypes.UPDATE_LAYERS,
}
```

I don't think this test passes as written here (apparently there are no CI run on PRs to not-main):
```
AssertionError: expected [Function] to throw error matching /updateLayers expects action.payload/ but got 'Cannot read properties of undefined (…'

- Expected:
/updateLayers expects action.payload/

+ Received:
"Cannot read properties of undefined (reading 'windowId')"
```

Maybe that means my error message guard has some problems. But as is something is not quite right.

PR #4481, opened 2026-08-19
Use setCanvas instead of updateWindow to change visible canvases.

### LINKED ISSUE #3276: Chosing scroll view does not update visible canvases

### LINKED ISSUE #4393: Bug when switching from single to book view

cbeer:
An alternative to #4449, with the added benefit that we'll fetch annotations (or anything else that is normally triggered by SET_CANVAS)

closes https://github.com/ProjectMirador/mirador/issues/3276
closes #4393

### DIFF
--- a/__tests__/src/sagas/windows.test.js
+++ b/__tests__/src/sagas/windows.test.js
@@ -403,12 +403,9 @@ describe('window-level sagas', () => {
         .provide([
           [select(getWindow, { windowId }), { canvasId: 'y' }],
           [select(getCanvasGrouping, { canvasId: 'y', windowId }), [{ id: 'y' }, { id: 'z' }]],
+          [call(setCanvas, windowId, 'y', ['y', 'z']), { type: 'expectedThunk' }],
         ])
-        .put({
-          id: windowId,
-          payload: { visibleCanvases: ['y', 'z'] },
-          type: ActionTypes.UPDATE_WINDOW,
-        })
+        .put({ type: 'expectedThunk' })
         .run();
     });
   });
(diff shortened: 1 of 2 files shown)


dnoneill:
I personally prefer this. I kept trying to get AI to figure this out because I still don't quite get sagas and even when I asked 4 times are you sure this makes sense/matches the way Mirador is structured it kept giving me the PR I put up. I still don't understand sagas and I am not quite sure I understand this PR but it seems more in line with the way I have seen sagas fixed in Mirador

dnoneill (commented):
This is almost there. It doesn't resize correctly from scroll to book view and I am not 100% sure that this is really an issue that needs to be solved in this PR or another ticket. @marlo-longley do you have an idea. See attached video for what I am talking about

https://github.com/user-attachments/assets/886f7ce0-c4ab-4ba7-b461-18078b90287e

dnoneill (approved):

PR #4483, opened 2026-08-20
make sure to set initialBounds every time an image loads

### LINKED ISSUE #4104: Paged manifests are aligning the 2up display with the cover image
Viewing a paged manifest displays the cover image correctly, but the subsequent 2-up display's left edge is aligned with the left edge of the cover image, which pushes the images out of the viewport.

Image

Image

dnoneill:
closes #4104 

For book view we shouldn't be setting the initial viewer config on the first image load. This makes sure to reset zoom bounds when a second image is loaded.

### DIFF
--- a/__tests__/src/components/OpenSeadragonComponent.test.jsx
+++ b/__tests__/src/components/OpenSeadragonComponent.test.jsx
@@ -5,18 +5,17 @@ import OpenSeadragonComponent from '../../../src/components/OpenSeadragonCompone
 vi.mock('openseadragon');
 
 describe('OpenSeadragonComponent', () => {
-  let addOnceHandler;
+  let addHandler;
   let fitBoundsWithConstraints;
 
   beforeEach(() => {
-    addOnceHandler = vi.fn();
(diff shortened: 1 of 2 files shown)


marlo-longley:
@dnoneill for these mysterious bugs, could you post a testing flow or manifest that I could see before/after? 

And also do you have any theories for why you were consistently able to reproduce the problem on page 8 but not other pages for the Durham manifest, is it related to this fix? I can confirm page 8 loads centered/correctly on this branch. Is it a super large image or something?

marlo-longley (changes requested):
This is not the right solution, it is a serious regression.
Try zooming in or panning on any image on this branch.
With this change, since tile-loaded fires continuously during normal use — panning to a new area, zooming into a region, anything that streams in new tiles — this means the viewport snaps back to the initial fit-to-bounds view on essentially every interaction.

dnoneill:
@marlo-longley is correct this won't work. Pretty sure found much cleaner way around.

PR #4490, opened 2026-08-21
Stop using deprecated getDescription from Manifesto.js

marlo-longley:
Manifesto's getSummary handles the Presentation 3 to 2 fallback
We were printing both getSummary and the old getDescription, which led to the bug reported by a user on the Mirador Slack channel

### DIFF
--- a/src/state/selectors/manifests.js
+++ b/src/state/selectors/manifests.js
@@ -291,10 +291,17 @@ export const getManifestTitle = createSelector(
  * @param {string} props.windowId
  * @return {string|null}
  */
-export const getManifestSummary = createSelector(
+const getManifestSummary = createSelector(
   [getLocale, getManifestoInstance],
   (locale, manifest) => manifest && manifest.getSummary().getValue(locale),
 );
+
+export { getManifestSummary };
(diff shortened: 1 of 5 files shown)


cbeer:
It looks like `getSummary` was nice in manifesto v4.3.0

src/state/selectors/manifests.js
```diff
@@ -284,29 +284,16 @@ export const getManifestTitle = createSelector(
...
 
 /**
- * Return manifest description (IIIF v2) -- distinct from any description field nested under metadata.
- * @param {object} state
- * @param {object} props
- * @param {string} props.manifestId
- * @param {string} props.windowId
- * @returns {string|null}
- */
-export const getManifestDescription = createSelector(
```
cbeer, line 294: Is it worth re-exporting `getManifestSummary` as `getManifestDescription` to keep backwards-compatibility?

marlo-longley: @cbeer thanks, I added the export back and marked it as deprecated

dnoneill (approved):

PR #4493, opened 2026-08-24
fragment overlay annotation fix

### LINKED ISSUE #3469: Strange offsets in search result highlighting
Hi,

we've seen some strange highlighting of search results in the "book" view of latest Mirador version.

Steps to reproduce:

* Go to demo site: `https://mirador-dev.netlify.app/__tests__/integration/mirador/`
* Import our IIIF manifest: `https://api.digitale-sammlungen.de/iiif/presentation/v2/bsb00092424/manifest`
* Search for the term `Zimmer`
* Go to search result 50, which is on page 1147

Then you'll see the following highlighting:

image

You can see that `Zimmer` is definitely not matched and there's an offset.

Then switch to "single" view. The search term is perfectly highlighted then:

image

/cc @cbeer 

Any help would highly be appreciated :hugs:

dnoneill:
closes #3469

### DIFF
--- a/__tests__/src/components/AnnotationsOverlay.test.jsx
+++ b/__tests__/src/components/AnnotationsOverlay.test.jsx
@@ -114,9 +114,16 @@ describe('AnnotationsOverlay', () => {
         viewer: null,
       });
 
-      const getItemAt = vi
-        .spyOn(viewer.world, 'getItemAt')
-        .mockImplementation((index) => (index === 0 ? { viewportToImageZoom: vi.fn(() => 0.05) } : undefined));
+      const getItemCount = vi.spyOn(viewer.world, 'getItemCount').mockReturnValue(1);
+      const getItemAt = vi.spyOn(viewer.world, 'getItemAt').mockImplementation((index) =>
+        index === 0
+          ? {
+              canvasId: 'http://iiif.io/api/presentation/2.0/example/fixtures/canvas/24/c1.json',
+              imageToViewportRectangle: vi.fn().mockReturnValue({ x: 10, y: 10, width: 100, height: 200 }),
+              viewportToImageZoom: vi.fn(() => 0.05),
+            }
+          : undefined,
+      );
       vi.spyOn(viewer.viewport, 'getZoom').mockImplementation(() => 0.05);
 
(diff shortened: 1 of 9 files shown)


geourjoa:
Hi 

This seem LGTM. 

Have you tested edge case like : 
- really tall or really wide image 
- really small image or really big image with stacked zoom (zoom in Mirador and Zoom in browser)
- Canvas with annotation targeting different canvas (we wnt to see only the annonation targeting the current one)
- Resize the window viewer 

Also inside `src/components/AnnotationsOverlay.jsx` the variable `viewportCanvas` is shared across multiple iteration.
It can cause issue if annotations target unloaded canvas.

Suppose the viewer has only canvas A loaded:

- Annotation 1 targets canvas A → found
- Annotation 2 targets canvas B → not found

If viewportCanvas is not reset before searching for Annotation 2, it may still equal canvas A.

Then Annotation 2 could be rendered using canvas A’s coordinate conversion, which is wrong.

My fix can be : 

```js
renderedAnnotations.forEach((annotation) => {
  annotation.resources.forEach((resource) => {
    let viewportCanvas; // reset per resource

    const count = viewer.world.getItemCount();
    for (let i = 0; i < count; i++) {
      const item = viewer.world.getItemAt(i);
      if (item.canvasId === resource.targetId) {
        viewportCanvas = item;
        break;
      }
    }

    if (!viewportCanvas || !canvasWorld.canvasIds.includes(resource.targetId)) return;
    ...
  });
});
```

Note : I'm the maintainer of the annotation plugin (MAE). Annotation overlay is a painful subject ...

dnoneill:
@geourjoa I have not because I am not aware of any search annotations on the images you have described. If you know of any please let me know. And this fix is really only affecting search annotations.

I think what you have outlined is an extreme edge cause considering renderAnnotations but I also don't think there is any harm guarding against that so I have updated the code.

Also since you are the maintainer of the annotation plugin I would love to talk to you about this spike I am working on to make annotation overlay way less painful https://github.com/ProjectMirador/mirador/pull/4494. I think it would actually make the annotation plugin easier but I would love your perspective.

PR #4499, opened 2026-08-28
add percentage fragment selector support

dnoneill:


marlo-longley (commented):
This is looking good!
I noticed how many function signatures change in this PR and was curious if there's another way...I have concerns about the multi layered threading of the new `canvas` param through many different files.

I was thinking about another approach where we only need to know about `canvas` at  `components/AnnotationsOverlay.jsx` and  `lib/CanvasAnnotationDisplay.js`. The selector layer, the container, and AnnotationList (the v2 wrapper, which has nothing to do with rendering) would not need to be passing `canvas` around. We'd add something like `fragmentSelectorIsPercent` as a boolean check. So the percent to pixel conversion would happen at the two places that actually draw/hit-test a fragment (both already have canvas data).

The con of that approach is that it's less foolproof than yours -- in your version fragmentSelector always returns pixels.

Given there are only two consumers of fragmentSelector right now, I am sort of thinking we should limit the passing of parameters around. Just thinking about the overall readability/keeping the codebase less tangled.

If you have time check it out and lmk what you think: https://github.com/ProjectMirador/mirador/compare/review/pct-implementation-alt-on-main?expand=1 -- mostly it is your code but rearranged. Your solution works though and I am not blocking on this!

PR #4502, opened 2026-09-01
Add classnames to make styling with custom CSS possible

morpheus-87:


jcoyne:
Can you say more about why you are making this change?

morpheus-87:
I need to select this region to style it with custom CSS.

Our usecase is that we use Tailwind CSS in the application hosting Mirador. They have some very global rules like this one:

`
img {
  max-width: 100%
}
`

In Mirador 3 that wasn't a big deal because there was a custom `max-width` set on every image in the gallery view. Mirador 4 doesn't have this behaviour anymore, so we have to fix it with custom CSS.

marlo-longley (approved):

PR #4518, opened 2026-09-03
add listener for removed canvases

### LINKED ISSUE #4516: Zoom issue when switching from scroll to single
This is almost there. It doesn't resize correctly from scroll to book view and I am not 100% sure that this is really an issue that needs to be solved in this PR or another ticket. @marlo-longley do you have an idea. See attached video for what I am talking about

https://github.com/user-attachments/assets/886f7ce0-c4ab-4ba7-b461-18078b90287e

_Originally posted by @dnoneill in https://github.com/ProjectMirador/mirador/pull/4481#pullrequestreview-5021807789_

dnoneill:
closes #4516 

https://github.com/ProjectMirador/mirador/blob/3e6fc4cc6dcefaaed1b6f182a49c42933b67da77/src/components/OpenSeadragonComponent.jsx#L112-L126 is working properly but the listener for `tile-loaded` is never going to trip because the tile is already loaded and the thing that is happening is the other images are getting removed. This just does the work. I could use some more information on how the viewportChange listener is supposed to work. It seems like we could rip out a ton of unncessary code with the bounds changed but I am not sure when else that might trip.

### DIFF
--- a/src/components/OpenSeadragonComponent.jsx
+++ b/src/components/OpenSeadragonComponent.jsx
@@ -191,6 +191,11 @@ function OpenSeadragonComponent({
       setInitialBounds(viewer);
     });
 
+    viewer.world.addHandler('remove-item', () => {
+      initialViewportSet.current = false;
+      setInitialBounds(viewer);
+    });
+
     forceUpdate();
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [ref]);


marlo-longley:
Steps to test:
1. Open a book view manifest such as https://iiif.wellcomecollection.org/presentation/v2/b18031511_0004
2. Page to a facing page canvas
3. Switch from Book to Single view: notice the Single page is centered
4. Switch back to Book view: also centered

On this Branch, all works well.
On `main` (and this is the bug), on step 3, the Single view will not properly center.

PR #4523, opened 2026-09-04
add contentState canvas/manifest capability

dnoneill:
Time code doesn't work but everything else does. I think the time code should probably be a separate PR since we don't have https://iiif.io/api/cookbook/recipe/0015-start/ implemented

Disclaimer: This code was written with the help of AI because sagas still don't make sense to me. I have pretty carefully reviewed the code but it could use probably a little extra than normal review.

### DIFF
--- a/src/components/OpenSeadragonComponent.jsx
+++ b/src/components/OpenSeadragonComponent.jsx
@@ -192,7 +192,7 @@ function OpenSeadragonComponent({
     viewerRef.current = viewer;
     setViewer(viewer);
 
-    viewer.world.addHandler('add-item', () => {
+    viewer.addHandler('open', () => {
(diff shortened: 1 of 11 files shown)


src/components/OpenSeadragonComponent.jsx
```diff
@@ -192,7 +192,7 @@ function OpenSeadragonComponent({
     viewerRef.current = viewer;
     setViewer(viewer);
 
-    viewer.world.addHandler('add-item', () => {
+    viewer.addHandler('open', () => {
```
dnoneill, line 195: This was updated because we don't want to set the intialViewer position until the image is loaded. That is what this change does and it stops a race condition from being tripped.

Issue #4226, opened 2026-02-05
Initializing Mirador on a specific search result


fitnycdigitalinitiatives:
Hello,

I'm trying to get Mirador viewer (using UMD) to initialize on a search result on a specific canvas. I tried doing this by initializing with `defaultSearchQuery` and also setting the specific `canvasId`. This almost works but if you have `switchCanvasOnSearch` set to default true, the `canvasId` get overridden and it goes to the canvas of the very first search result, which is not necessarily the canvas we want. If I change `switchCanvasOnSearch` to false, it's all good...but if a user then wants to change their search, it just doesn't feel like a good user experience because it should then jump to that first search result.

Just wondering if there's something obvious I might be missing.

Thanks,

Joe

marlo-longley:
@fitnycdigitalinitiatives thank you for documenting this and engaging with Mirador

Audit 2026:
Same developer should pick up #4231 

Actions:
- Reproduce the behavior and provide screenshots
- Find the technical root cause and make a PR if fixable
- Mark as design or info needed if there is a product decision to be made about settings priority/clashes

PR #4528, opened 2026-09-09
Add setting preserveInitialCanvasOnSearch to resolve conflict between switchCanvasOnSearch and initial canvasId

marlo-longley:
… switchCanvasOnSearch and initial canvasId

Idea for resolving https://github.com/ProjectMirador/mirador/issues/4226

Sort of a long/confusing setting name -- but we could document its existence.
@fitnycdigitalinitiatives would this help you?

### DIFF
--- a/__tests__/src/reducers/windows.test.js
+++ b/__tests__/src/reducers/windows.test.js
@@ -13,6 +13,24 @@ describe('windows reducer', () => {
       ),
     ).toEqual({
       abc123: {
+        explicitInitialCanvasId: false,
+        id: 'abc123',
+      },
+    });
+  });
+  it('should handle ADD_WINDOW with an explicit canvasId', () => {
+    expect(
+      windowsReducer(
+        {},
+        {
+          type: ActionTypes.ADD_WINDOW,
+          window: { canvasId: 'canvas-1', id: 'abc123' },
+        },
+      ),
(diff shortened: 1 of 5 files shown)


fitnycdigitalinitiatives:
That would seem to do the trick, thanks!

marlo-longley:
@fitnycdigitalinitiatives OK thanks! I'll show this to my colleagues, one already said it seemed potentially too specific/obscure use case for a whole new setting but I will keep you updated. If I understand correctly, you want to use the `defaultSearchQuery` functionality but have it open a page other than the first page of search results.

fitnycdigitalinitiatives:
Yes, that's correct. I would like default functionality but basically the option to open the viewer on the page of specific search results.

Our use case is we have a full-text search across multiple items and the results are shown to users with snippets of text. When the user selects one of those results, it should open the viewer of that item with on the same page and search of the snippet. Here's an [example](https://sparcdigital.fitnyc.edu/ocrsearch?q=yellow+dress) from our site, but it requires `switchCanvasOnSearch ` being set to false. If you open one of the results, but then enter a new search, it doesn't jump to the first result.

marlo-longley:
I see! Nice to see your site as well, it looks really good! Thanks for the use case. I will keep working on this.

PR #4529, opened 2026-09-10
Defer initial viewport fit until the viewer element is visible

### LINKED ISSUE #3540: Mirador not working in Bootstrap5 Tabs
I would like to organise the content page through the Bootstrap5 Tabs and insert the Mirador viewer in the third tab. 
In this way the viewer does not work and creates _TypeError: Cannot read properties of null (reading 'resetAfterIndex')_. 
But if the viewer is in the first tab, the one that is open and displayed when the page is loaded, then it works perfectly. 
Is there a solution to make the viewer work even if it is not displayed immediately?

Problem:
mirador-tab-problem

Error:
typeError

marlo-longley:
add-item/remove-item can fire while the viewer's element has no
rendered box at all (e.g. an inactive Bootstrap/tab panel), so
OpenSeadragon can't compute a fit. Extracted the wait-until-visible
logic into a reusable useDeferUntilVisible hook.

Fixes #3540
Supersedes #4506 -- needed a rebase / rewriting after #4526; extracted logic into a hook

### DIFF
--- a/__tests__/src/components/OpenSeadragonComponent.test.jsx
+++ b/__tests__/src/components/OpenSeadragonComponent.test.jsx
@@ -9,12 +9,16 @@ describe('OpenSeadragonComponent', () => {
   let addHandler;
   let fitBoundsWithConstraints;
   let goHome;
+  let checkVisibility;
+  let element;
 
   beforeEach(() => {
     addOnceHandler = vi.fn();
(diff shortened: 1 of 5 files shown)


marlo-longley:
Currently superseded by #4539 (in draft)

PR #4530, opened 2026-09-10
Rename function setInitialBounds to applyViewport for clarity

marlo-longley:
I believe the name of this method caused confusion and contributed us (me at least) to not seeing the bug in #4526 
I thought it was initial, only set once on component load -- but in fact it runs per add or remove item.


marlo-longley:
HOLD: we might think of a different name
`applyViewport` may also be inaccurate

PR #4535, opened 2026-09-10
wait until image loads to set bounds

dnoneill:
This is needed for content state but I also noticed a race condition in https://iiif.archive.org/iiif/3/CAT31342177/manifest.json that this fixes as well.


marlo-longley:
@dnoneill can you share the steps to reproduce the error you were getting?

marlo-longley (changes requested):
If there are race conditions, this is likely due to the different systems between OSD's event-emitter style and React's lifecycle system. All our bugs are really about this mismatch. I do not think we want to be in the business of adding `raiseEvent` through OSD to communicate within Mirador, this will be even messier. We can talk about other approaches.

dnoneill:
Okay. The race condition I was seeing was just loading https://iiif.archive.org/iiif/3/CAT31342177/manifest.json into mirador then going to page two. I think it is just happening locally

PR #4537, opened 2026-09-14
use canvas dimensions and not image dimensions for annotation overlay

### LINKED ISSUE #3469: Strange offsets in search result highlighting

### LINKED ISSUE #4533: Annotation #xywh scaled by image pixel size instead of canvas coordinate space
## Summary

`#xywh` fragments on a canvas target are scaled by the **pixel dimensions of the image actually loaded**, not by the canvas coordinate space. Where a canvas is painted with a derivative smaller than the canvas dimensions, every annotation box is drawn too large by exactly `canvas.width / image.width`.

Per the Presentation 3 spec, a fragment on a canvas target is expressed in canvas coordinate space, and the canvas is an abstract coordinate space that need not match the pixel size of the image painting it.

## Reproduce

Canvas 3600 × 4831, painted by an image whose real size is 930 × 1248, with two annotations:

```json
"target": "https://example.org/canvas/p3#xywh=0,0,100,100"
"target": "https://example.org/canvas/p3#xywh=0,100,3600,100"
```

Rendered at 576 screen px for the full page width:

| scaled by | predicted box | observed |
|---|---|---|
| canvas width 3600 | 100 × 576/3600 = **16 px** | no |
| real image width 930 | 100 × 576/930 = **62 px** | **yes**, ~60–63 px |

Boxes come out 3.87× too big — exactly 3600 / 930. The second annotation, nominally the full canvas width, is drawn 3.87 canvas-widths wide and runs off the page.

## The declared body dimensions are not what's used

Worth stressing, because it's the part that surprised me: the manifest in that test **declares** `body.width` as 3600 (matching the canvas), and the boxes are *still* scaled by 930. So this isn't a case of trusting a bad declaration — the declared image dimensions are ignored and the decoded image's true size is used.

Consequently, correcting or fudging `body.width` has no effect either way. I verified both directions:

- `body` declared 930 × 1248 (true size), canvas 3600 → boxes 3.87× too big
- `body` declared 3600 × 4831 (false), canvas 3600 → boxes 3.87× too big

The only manifests that render correctly are those where the canvas dimensions happen to equal the served image's pixel dimensions.

## Why this shape is common

Internet Archive and Biodiversity Heritage Library material is the motivating case. IA's `_djvu.xml` OCR coordinates are expressed in the scan's coordinate space and match the `<cropBox>` dimensions in `_scandata.xml` — for one item I checked, 3600 × 4831. The web derivative served from S3 is 930 px wide. Using the scan dimensions as the canvas size means OCR-derived annotation coordinates need no transformation, which is the natural way to build these manifests. Every such manifest renders its annotations at the wrong size.

## Environment

Mirador 4.2.4 (`https://unpkg.com/mirador@latest/dist/mirador.min.js`), Chrome.

## Related

Possibly connected to the concern in #587 ("Verify if canvas/image scaling works when image is larger than canvas"), which covers the opposite ratio.

To see the boxes at all, every annotation on the canvas must carry a fragment — see the companion issue about coordless annotations suppressing all overlays.

dnoneill:
closes #4533 
closes https://github.com/ProjectMirador/mirador/issues/3469


src/lib/CanvasWorld.js
```diff
@@ -103,6 +103,17 @@ export default class CanvasWorld {
     return [canvasDimensions.x, canvasDimensions.y, canvasDimensions.width, canvasDimensions.height];
   }
 
+  /** */
+  canvasScale(canvasId) {
```
dnoneill, line 107: From what I can tell we need this for two page view. https://api.digitale-sammlungen.de/iiif/presentation/v2/bsb00092424/manifest is a good example. When you search Zimmer the first result is on page 0008. The canvas dimensions for 0008 and 0009 are different so we need the canvasDimensions.width which will be a different size than the actual canvasWidth.

marlo-longley (approved):
Looks good to me and this is in line with other things I'm thinking about this week. Category issue being the reliance on OSD for data, and this getting tangled with data we own in React/from IIIF.

Basically stop asking OSD/the loaded image for scale info at all — here you remove `viewer.world.getItemAt()`, `viewportToImageZoom`, `viewer.viewport.getZoom()` from the annotation path entirely, and add `CanvasWorld.canvasScale(canvasId)` which comes from the IIIF manifest data. Seems like the right approach.

marlo-longley (changes requested):
We noticed problems with hovering on annotations
Also unsure about anno line width

dnoneill:
@marlo-longley should be fixed.

marlo-longley (approved):

PR #4539, opened 2026-09-14
Stop relying on OSD events to fit viewport -- use IIIF manifest data instead

marlo-longley:
In the current architecture, Mirador's own add-item/remove-item handler refit the viewport on every tile event, which happened to overwrite OSD's goHome() a moment after it fired -- masking the bug rather than preventing it.

Fixes this by:
- Hardcoding `preserveViewport: true` on the OSD `Viewer` constructor, unconditionally, so this internal auto-recenter can never fire.
- Replacing add-item/remove-item-triggered camera positioning with a Redux-driven `useApplyViewport` hook, reacting to a `canvasKey`-tagged `desiredViewport` (read from the manifest, not waited on via `add-item`/`tile-loaded`) instead of OSD's own content-loading events.
- A settle-confirmation gate (`isApplying`/`expected` in `useApplyViewport`) so `animation-finish` events from other canvases settling their own placement mid-transition aren't misreported as user gestures.

Supersedes #4529 (`useDeferUntilVisible`), already open separately.

Issues that motivated this work:

- The underlying issue (#4516, "Zoom issue when switching from scroll to single"): a view-type switch landing on a canvas that's already visible only fires OSD's remove-item for the discarded companion canvases — never add-item for the survivor, since it was never removed and re-added. The original code only had a handler wired to add-item, so the survivor's refit never ran.

- #4527 ("Add handler for remove-item with current refs," merged) fixed that specific symptom by adding a second handler wired to remove-item, alongside the existing add-item one — a targeted, symmetric patch.

- #4526 ("Fix stale viewConfig from closure around handler," merged) fixed a separate symptom in the same code: the add-item handler was captured once at mount time via closure, so it kept reading the first render's viewerConfig forever instead of the current one. Fixed via a ref indirection.

Both #4526 and #4527 are separate one-off fixes to the same underlying design flaw: wiring persistent camera-positioning logic directly to raw OSD add-item/remove-item events, whose firing guarantees don't naturally line up with React's render lifecycle. 

`useApplyViewport` doesn't listen to add-item/remove-item for camera positioning at all anymore — it reacts to Redux's desiredViewport (canvasKey + bounds read from the manifest) changing


src/components/OpenSeadragonComponent.jsx
```diff
@@ -36,137 +36,69 @@ function OpenSeadragonComponent({
...
+      // OSD's own preserveViewport option (unrelated to Mirador's own
+      // preserveMiradorViewport feature -- see settings.js) governs an
+      // internal, undocumented auto-goHome() whenever world.getItemCount()
+      // === 1, which can fire on whichever single canvas happens to be
+      // momentarily alone in the world mid-transition (each canvas's own
+      // <TileSource> mounts independently), racing against our own
+      // useApplyViewport fit. Since this architecture fully owns camera
+      // positioning, that internal auto-recenter must always be disabled --
+      // never a pass-through of any Mirador config value.
+      preserveViewport: true,
```
marlo-longley, line 101: this is where I hardcode OSD preserve viewport which is def up for discussion

PR #4545, opened 2026-09-17
Call OSD's applyConstraints on button zoom; Extract zoom method

marlo-longley:
- **Apply OSD's applyConstraints on custom button zooms to enforce min/max**
- **Extract method for applying viewport**

We noticed this while reviewing work in another branch.
Reproduce:
1. https://mirador-dev.netlify.app/
2. Using mouse or trackpad, zoom as far in as possible. Notice at a certain point you cannot zoom further.
3. Using Mirador's + zoom button, zoom in as far as possible. Notice you can zoom all the way to a single pixel -- which looks like a gray screen.

Reason:
OSD's own buttons use `viewport.applyConstraints` natively
We have custom buttons so this was not being applied
It was being applying on trackpad, because those events are captured and handled in native OSD.
I guess this was always true in Mirador? Can anyone tell me this is a regression?

Further work:
I noticed a lot of duplicate code so I extracted a shared method.
Overall goal is simplifying this difficult file piece by piece.

Update:
3rd commit to expand test coverage due to Codecov failure. These tests were mostly written by Claude.


src/components/OpenSeadragonComponent.jsx
```diff
@@ -5,6 +5,38 @@ import { useDebouncedCallback } from 'use-debounce';
...
+  ) {
+    viewport.panTo(point, immediately);
+  }
+
+  // Independent of x/y -- zoom can be applied on its own.
+  if (zoom != null && (immediately || zoom !== viewport.zoomSpring.target.value)) {
+    viewport.zoomTo(zoom, point, immediately);
+    // OSD's own zoom buttons always pair a zoom call with applyConstraints.
+    // Because we have custom buttons, we must apply it here to enforce min/max zoom.
+    viewport.applyConstraints();
```
dnoneill, line 33: I am not sure this is really the correct solution. This means if I keep hitting the zoom in button zoom is still set to the really large number, i.e. if .5 is max zoom, this code will still zoom into .1 zoom and when I try to zoom out it will need to zoom out again. Please see attached video for demonstration. Also I feel like the zoom button should be disabled if you can't zoom anymore. It is hard to tell based on the screen recording but whenever my mouse is on a zoom button I am clicking on it repeatedly.

https://github.com/user-attachments/assets/9b4f652a-cfff-488e-9f87-34542357d1b3

